PR #293, opened 2021-06-15
Filter Extension Re-organisation

### LINKED ISSUE #280: Refactor Filter Module Structure
Building off #265, currently the filters module structure is very dense, for example  the `filter_registry` module contains the `Filter` and `FilterFactory` traits in addition to `FilterRegistry` . I would like to propose that we break it up a bit more internally to make it easier to find things.

I would like to propose that we remove the `extensions` part of the tree, because the filter traits, chain, and registry feel like a core part of Quilkin, and the filters themselves are included by default. At the very least I think it should be `filters::extensions` rather than the other way around..

XAMPPRocky:
This Pr builds on-top of #286 and refactors the API structure of the filter changes to be easier to use, and cleans up the public API surface.

Changes:

- All filter extension modules are now public.
- All filters and filter factories are removed from public API. Previously we exposed and asked the user to create the factory type themselves. However we don't actually need to expose the concrete types since we are only ever using them as trait objects. Hiding these implementation details will give us more freedom in the future to change and optimise the layouts of filter and filter factory structs without worrying about those details being relied on by users.
- To replace the public factory types, every filter extension module as a convention now has a top level free function called `factory` that constructs a new instance of `DynFilterFactory`. This makes paths to construct filters much less verbose while conveying the same information, and hiding the underlying implementation. Considering the following example.
    ```rust
    // Old
    let factory = quilkin::filters::extensions::DebugFactory::new(&logger);
    // New
    let factory = quilkin::filters::debug::factory(&logger);
    ```
- The filter extension documentation that was external has been moved into respective module's API documentation. This felt like it made more sense, as this would allow this documentation to use intra-doc links, and also has the side benefit of removing the nightly dependency for testing. Using rustdoc here reduces the amount of boilerplate we need to document and removes a lot of potential for docs to get out of sync, as we can document things like configuration through the public Rust types, even if you aren't using Rust, you'll be able to see what you can set.
- The `extensions` module is removed, and some filter extensions were further broken up into submodules if they had a significantly enough complex configuration type and traits.

closes #280 

### Filter Extension Docs Example
Screenshot 2021-06-15 at 10 43 24


docs/extensions/filters/writing_custom_filters.md
```diff
@@ -207,7 +208,8 @@ The [Serde] crate is used to describe static YAML configuration in code while [P
 
    # use quilkin::filters::{Filter, ReadContext, ReadResponse, WriteContext, WriteResponse};
 
-   #[quilkin::filter("greet.v1")]
+   const NAME: &str = "greet.v1";
+
```
iffyio, line 212:
```suggestion
```

src/filters/factory.rs
```diff
@@ -16,12 +16,22 @@
...
 pub type DynFilterFactory = Box<dyn FilterFactory>;
 
 /// Provides the name and creation function for a given [`Filter`].
+///
+/// An implementation of [`FilterFactory`] provides a `name` and
+/// `create_filter` method.
+///
+/// - The `create_filter` method takes in [configuration][filter configuration]
+///   for the filter to create and returns a new instance of its filter type.
+/// - The `name` method returns the unique identifier of the filter.
```
iffyio, line 34:
```suggestion
```
This seems redundant to list and describe the trait methods here and repeat them alongside the functions themselves

iffyio (approved):

docs/extensions/filters/writing_custom_filters.md
```diff
@@ -67,9 +67,10 @@ To extend Quilkin's code with our own custom filter, we need to do the following
    // src/main.rs
    use quilkin::filters::{Filter, ReadContext, ReadResponse, WriteContext, WriteResponse};
 
+   const NAME: &str = "greet.v1";
```
markmandel, line 70: I was hoping that #319 's snapshotting would solve this issue, but looks like that's blocked for the moment.

Since this is a change in API, and this will break the docs for 0.1.0 -- I'm wondering how to handle this in the short term. 

Maybe we add a note here: 
image

That says something like:

> This documentation is for the development version of Quilkin. To view the documentation for a specific release, switch to the release tag in the above dropdown.

(I'm happy to do this as a seperate PR if we are all happy with it).

Which can then we can remove it once #319 is merged and good to go.

Other than that, this is 👍🏻  as far as I am concerned - and I want to get it in, as I know a lot of your follow up work is based on this.

Sound good?

XAMPPRocky: SGTM, I can make a separate PR.

markmandel (approved):

PR #301, opened 2021-06-24
Add wrapper function filter factories

XAMPPRocky:
This is an iteration on the original idea I had for `#[quilkin::filter]`, but I realised we could create an abstraction within the typesystem itself. This abstraction allows you to write simple single function filters without needing the boilerplate of defining your own types.

```rust
let factory = quilkin::filters::read("print-contents.v1", |ctx| {
     println!("packet: {:?}", ctx.contents);
     Some(ctx.into())
});
```

### DIFF
--- a/src/filters.rs
+++ b/src/filters.rs
@@ -22,6 +22,7 @@ mod factory;
 mod read;
 mod registry;
 mod set;
+mod wrap;
 mod write;
 
 pub(crate) mod chain;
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Just noting the current status of this, I want to waitt until #346 is available, as I want to use that in combination with this to allow you to write the following for making a filter.

```rust
#[derive(quilkin::Config, Default)]
struct Config {
    hex_output: bool,
}

let factory = quilkin::filters::read::<Config>("print-contents.v1", |ctx, config| {
     if config.hex_output {
        println!("packet: {:?}", ctx.contents);
     } else {
        println!("packet: {:x?}", ctx.contents);
     }
     
     Some(ctx.into())
});
```

PR #308, opened 2021-06-30
Cleanup other parts of the public API.

XAMPPRocky:
This now makes the `proxy` and `runner` private modules as they don't really provide much for users of the library as the `proxy` and the `runner` could both be considered Quilkin. So their types are now exported at the top-level. So instead of `quilkin::runner::run(…)` it's `quilkin::run(…)` for "running Quilkin", and it's `quilkin::Builder` instead of `quilkin::proxy::Builder`, as you're building a Quilkin server. This is how the top-level docs looks like with these changes.

Screenshot 2021-06-30 at 09 37 01

### DIFF
--- a/benches/throughput.rs
+++ b/benches/throughput.rs
@@ -31,7 +31,7 @@ static SERVER_INIT: Lazy<()> = Lazy::new(|| {
                 )],
             )
             .build();
-        let server = quilkin::proxy::Builder::from(std::sync::Arc::new(config))
+        let server = quilkin::Builder::from(std::sync::Arc::new(config))
             .validate()
             .unwrap()
             .build();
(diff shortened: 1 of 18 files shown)


iffyio:
Is this PR good to review? I'm a bit unsure what's what because from a quick look I could have sworn I had reviewed the changes in another PR :smile:

XAMPPRocky:
> Is this PR good to review? I'm a bit unsure what's what because from a quick look I could have sworn I had reviewed the changes in another PR 😄

That's because it includes #293, for the specific changes in this PR, it's in this commit.

https://github.com/googleforgames/quilkin/pull/308/commits/0d07c3fd284a8156aca944935a0df02ec09646ba

.github/workflows/docs.yml
```diff
@@ -0,0 +1,21 @@
+name: Deploy docs
```
markmandel, line 1: Guessing this is blocked on #319 right now?

markmandel (commented):
No major objections on my end, looks like a nice cleanup to pull the public API up a level.

Just looks like it's blocked on a few other PRs that are also merged into this one.

@iffyio any issues on your end?

.github/workflows/docs.yml
XAMPPRocky: Nope unintentional, don't know how this commit made it in here, will remove.

iffyio (approved):

Issue #309, opened 2021-06-30
How do Filters send packets? (or should they?)


markmandel:
There are lots of use cases for being able to send packets from a Filter, or something similar, e.g.

- DTLS/Encryption that requires a handshake
- Sending regular packets to track ping time and/or jitter

Right now, Filters can only intercept and manipulate packets.

This could be:
- An extension or modification to how Filters work
- A whole new different type of manipulation module (are Envoy [Listeners](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/listeners/listeners#arch-overview-listeners) a good model?)
- Something else?

Issue #13, opened 2020-01-28
Filter idea: Scriptable filter


markmandel:
(Placeholder, more to work out)

The idea being, we would integrate some kind of scripting langauge into Quilkin, such that if you needed custom filters, you could script them, rather than having to use Quilkin as a library and build your own binaries.

This might tie well into #12 - since scripted filters are expected to be slower than native rust (although worth experimentation).

Useful resources:
- https://arewegameyet.com/categories/scripting/
- https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/lua_filter

iffyio:
Tested out adding both a [lua](https://github.com/amethyst/rlua) and [rhai](https://github.com/rhaiscript/rhai) script filter https://github.com/iffyio/quilkin/commit/35d2cef4682e3b310f8e7ecbf6f50c191b55a8e1
both turned out to be painfully slow unfortunately, tested with a single filter without any logic, the rhai filter was 4x slower than a native rust noop filter while lua was even slower than rhai (though I suspect part of Lua's slowness was due to having to use a [Mutex](https://github.com/iffyio/quilkin/commit/35d2cef4682e3b310f8e7ecbf6f50c191b55a8e1#diff-f0a31551f27229288c4476603798ac4a4378ba0e2d1cec9d735e976ff1a3ba7bR59) to get stuff to compile, essentially serializing packet forwarding)

XAMPPRocky:
> both turned out to be painfully slow unfortunately, tested with a single filter without any logic, the rhai filter was 4x slower than a native rust noop filter while lua was even slower than rhai 

Yeah, that's to be expected, I'm interested to see how wasmtime compares, but there's always going to be some slowdown in the process of copying the packet memory into the language VM, maybe with something like [`witx-bindgen`](https://github.com/bytecodealliance/witx-bindgen) we could generate more efficient bindings.

markmandel:
Two thoughts:

1) Do we want to drop scripting all together? Maybe it just doesn't fit here? Our API surface does seem to be aimed at "if you want a custom filter, use quilkin as a library and write some Rust -- and then let's look at (#346) how we can make it the best experience there we can.

2) We discussed in #318 - maybe scripting is _only_ for testing, and not for general usage?

If we did give it up -- does that make some parts of Quilkin simpler for us to design, implement and manage long term?

XAMPPRocky:
[...]

PR #315, opened 2021-07-05
Make `Filter` an `async trait`.

### LINKED ISSUE #12: Filter idea: Asynchronous Filtering
Be able to have a filter that is processed Asynchronously.

This wouldn't have the ability to block, or route requests, but would allow for processing and parsing of packets not on the critical path.

We could also provide options to have a "sample-rate" in which you could retrieve a percentage of packets, either randomly, or deliberately sequentially, for systems that don't need to query every packet to process.

The main use case I can think of is doing something like deep introspection of packets for malicious elements - taking a random sample of packets for processing, could be fast enough to find a bad actor without them being able to cause large scale harm, without slowing down processing speed drastically.

XAMPPRocky:
This PR makes the `Filter` trait asynchronous, allowing implementors to insert yield points. It's worth noting that you could always spawn background tasks even before this change it just wouldn't have been useful. This PR doesn't change any behaviour other adding async, but this will also allow us to try things like cancellable, concurrent, and parallel filters.

This is currently built off #293 to reduce merge conflicts, the important work is in  https://github.com/googleforgames/quilkin/commit/7c0d5fd42afae63018debed77f58f1acec4cd204

closes #12

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,7 @@ tokio-stream = "0.1.2"
 tonic = "0.5.0"
 uuid = {version = "0.8.1", default-features = false, features = ["v4"]}
 thiserror = "1.0.25"
+async-trait = "0.1.50"
 
 [dev-dependencies]
 reqwest = "0.11.0"
--- a/docs/extensions/filters/writing_custom_filters.md
+++ b/docs/extensions/filters/writing_custom_filters.md
@@ -73,12 +73,13 @@ To extend Quilkin's code with our own custom filter, we need to do the following
    // to `"greet.v1"`.
    struct Greet;
 
+   #[async_trait::async_trait]
    impl Filter for Greet {
-       fn read(&self, mut ctx: ReadContext) -> Option<ReadResponse> {
+       async fn read(&self, mut ctx: ReadContext) -> Option<ReadResponse> {
            ctx.contents.splice(0..0, String::from("Hello ").into_bytes());
            Some(ctx.into())
        }
-       fn write(&self, mut ctx: WriteContext) -> Option<WriteResponse> {
+       async fn write(&self, mut ctx: WriteContext<'async_trait>) -> Option<WriteResponse> {
            ctx.contents.splice(0..0, String::from("Goodbye ").into_bytes());
            Some(ctx.into())
        }
@@ -210,13 +211,14 @@ The [Serde] crate is used to describe static YAML configuration in code while [P
 
    struct Greet(String);
 
+   #[async_trait::async_trait]
    impl Filter for Greet {
-       fn read(&self, mut ctx: ReadContext) -> Option<ReadResponse> {
+       async fn read(&self, mut ctx: ReadContext) -> Option<ReadResponse> {
            ctx.contents
                .splice(0..0, format!("{} ",self.0).into_bytes());
            Some(ctx.into())
        }
-       fn write(&self, mut ctx: WriteContext) -> Option<WriteResponse> {
+       async fn write(&self, mut ctx: WriteContext<'async_trait>) -> Option<WriteResponse> {
            ctx.contents
                .splice(0..0, format!("{} ",self.0).into_bytes());
            Some(ctx.into())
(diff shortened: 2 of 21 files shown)


XAMPPRocky:
I ran a comparison benchmark, and there is no noticeable impact between this and the `main` branch. So with that in mind, I would prefer that we make these `async`. Also with the [push for GAT stabilisation](https://blog.rust-lang.org/2021/08/03/GATs-stabilization-push.html) `async trait` is getting closer to stabilisation, so I'd rather have the capability of running filters as tasks and accept the `Box` for now.

comparison

iffyio:
So I ran a quick load test with that C tool I linked earlier, I added a noop filter (simple filter that checks if the packet length is even and drops the packet if its odd, only to simulate _some_ logic though I only sent even length packets) then ran quilkin with a config that has 10 (not sure if this is too large or too small for a typical use case) noops in the filter chain.
On my machine without async the proxy could process in the range 300-380k pps, with async it could do 260-350k pps. Its not a drastic difference but it is significant (and its larger with more filters). I mentioned earlier that the criterion benchmark we have currently is proabably too noisy to be reliable, that could be why its not detecting any difference between the two branches.

My point is not only whether or not there's a perf difference but also that its weird to add such a change at the same time that we're trying to optimize the hot path by reducing clones/allocations etc I think if there comes a need to add async filters then we should accept the `async-trait` Boxing (hopefully by then there's rust support for async traits :crossed_fingers: or we're in a better shape per-wise) but there isn't any such a need that we know of today? so I don't see any upside

markmandel:
> My point is not only whether or not there's a perf difference but also that its weird to add such a change at the same time that we're trying to optimize the hot path by reducing clones/allocations etc I think if there comes a need to add async filters then we should accept the async-trait Boxing (hopefully by then there's rust support for async traits crossed_fingers or we're in a better shape per-wise) but there isn't any such a need that we know of today? so I don't see any upside

I agree with this sentiment - unless we have a feature/use case that requires this change, it makes sense to leave these kinds of changes up until we need it. No need to add extra complexity if it's not required.

XAMPPRocky:
> I agree with this sentiment - unless we have a feature/use case that requires this change, it makes sense to leave these kinds of changes up until we need it. No need to add extra complexity if it's not required.

In #309 you say *"There are lots of use cases for being able to send packets from a Filter,"*, that requires I/O, and if we're going to have filters performing I/O they need to use async so that we more control on how to react to filesystem failures, slow network requests, etc, and we also have #13 which is another filter which could take an indeterminate length. So I think we've already found plenty of reasons to have this feature, and have found very little cost to it.

XAMPPRocky:
>  calling e.g into an ffi isn't inherently async, the same rules for our rust filters goes for the scriptable filter I imagine.

It isn't inherently asynchronous but's a lot harder to write code that handles indeterminism when it is a function rather than a task, like I don't know how you'd write a cancellable script filter without either using async or spawning a new thread. If we chose wasmtime we could have a fuel limit in the interpreter but that would require us to always have a fuel limit on any script which I don't think would be intuitive as a default.

markmandel:
With #309, I don't feel like we have reached consensus there - it seems to me like there are a lot of questions unanswered. I'm not even sure that allowing a Filter to send a packet is a good idea at all -- maybe an entirely different construct is required?. With #13 - is this even a good idea? (Maybe it isn't) - what scripts would we want to support, if we did support them, what performance would we want - I have a lot of questions on both.

I'd love to see either one or the other ticket have a few "yep, that looks good to me, let's implement that" type comments before we start heading down towards implementation.

gilesheron:
only just spotted this PR.   it certainly would have made writing my API lookup flavour of load-balancer easier (currently I'm spinning up a new OS thread for the API client and then using this trick: https://stackoverflow.com/a/66280983 to run an async fn from within the load-balancer that communicates with the API thread).

Feels to me like there are a bunch of things that could be pulled together here (e.g. #374 and #375).  From looking at the code we already have a session construct for returned traffic.  If we extended the session construct to outbound traffic then we could have a concept of a "routing" chain in addition to the filter chain, and have that chain run async and do things like load-balancing, stats export etc. that only need to happen once per session (or periodically for long-lived sessions...).

is it worth discussing this in the next community call?

markmandel:
Would love to hear if more people have use cases for adding the async trait to the Filter, just because of the performance hit.

So if you find one, please share!

PR #319, opened 2021-07-07
Add mdbook and GitHub pages deployment

XAMPPRocky:
This PR adds mdbook for documentation and adds a GitHub Action to deploy both it and the API documentation to GitHub pages, so we'll always have the latest documentation available if needed.

@markmandel You'll need to enable GitHub Pages in settings for this to work.


markmandel:
Merging with main, and running this on my fork I ran into the following error ([build](https://github.com/markmandel/quilkin/runs/3106538523?check_suite_focus=true)):

```
error: failed to run custom build command for `quilkin v0.2.0-dev (/Users/runner/work/quilkin/quilkin)`

Caused by:
  process didn't exit successfully: `/Users/runner/work/quilkin/quilkin/target/debug/build/quilkin-08b7231bd3875076/build-script-build` (exit code: 1)
  --- stderr
  Error: Custom { kind: Other, error: "protoc failed: Could not make proto path relative: /Users/runner/work/quilkin/quilkin/proto/data-plane-api/envoy/config/accesslog/v3/accesslog.proto: No such file or directory\n" }
error: build failed
Error: Process completed with exit code 101.
```
I also tried it with just the `deploy-docs` branch, and had the same issue ([build](https://github.com/markmandel/quilkin/runs/3106602381?check_suite_focus=true)).

Now that we're able to make forks, this should hopefully be easier to test as well.

Another thought: with the 0.1.0 snapshot: Should we be doing that manually? (I think that's fine). If so, did you want to just do that now based on the 0.1.0 tag?

XAMPPRocky:
@markmandel Ah, just an issue with submodules (of course). I've updated it now that it should work.

markmandel:
Ran it again, ran into a stack of issues unfortunately. Have fixed several things here:
https://github.com/markmandel/quilkin/blob/pr/deploy-docs/.github/workflows/docs.yml

(Ignore where I have `pr/**` as a push branch, that's just for testing).

* Setting of env variables ([docs](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable)) didn't pass between steps, so switched that out.
* Fixed issue with the `snapshot` variable so it can handle branch names with "/" in them
* Fix issues with `mkdir` for the same thing.

Successful run:
https://github.com/markmandel/quilkin/runs/3139428581?check_suite_focus=true

I learnt some things about github actions though!

Preview:
https://markmandel.github.io/quilkin/pr/deploy-docs/book/
https://markmandel.github.io/quilkin/pr/deploy-docs/api/quilkin/

I haven't had a chance to review the docs yet, but at least we can now see them 🙌🏻 

I also haven't tested it against creating a tag, so one of us should also take that for a spin so we can make sure it works as expected. That will likely be my next step unless you get to it first.

markmandel:
Tagging seems to be working as expected (as long as the action are in the tagged content).

I cleaned up the GH pages on my fork, and [manually made a snapshot for 0.1.0](https://github.com/markmandel/quilkin/tree/gh-pages) - once we resolve the below blocking issues, and if we're happy, I'll push it to main Quilkin repo, and make a PR to point the README.md to it, and then we can get this merged with my fixes. (TIL you can't make a PR to a branch that doesn't exist!)

0.1.0 documentation ([guide](https://markmandel.github.io/quilkin/main/book/), [api](https://markmandel.github.io/quilkin/v0.1.0/api/quilkin/))

The blocking issues I found were:

* The documentation for Filters is missing from the TOC. 
* Let's add an examples page that points to https://github.com/googleforgames/quilkin/blob/main/examples (I'd love to have it point to the same release branch as the snapshot with something like mdbook-variables, but can experiment with that later, I'm sure an explanatory note that they might need to switch to the correct release tag will in the repo will suffice).

I am keen to get this merged before we change the API surface, as otherwise it becomes quite difficult to get a mdbook snapshot of 0.1.0 (which we are still inline with at this point).

The other option is: We skip a snapshot of 0.1.0, and instead point the documentaton for 0.1.0 straight at https://github.com/googleforgames/quilkin/blob/release-0.1.0/README.md, and then 0.2.0 will be the first documentation snapshot we take with mdbook.

Thoughts?

XAMPPRocky:
> Let's add an examples page that points to https://github.com/googleforgames/quilkin/blob/main/examples (I'd love to have it point to the same release branch as the snapshot with something like mdbook-variables, but can experiment with that later, I'm sure an explanatory note that they might need to switch to the correct release tag will in the repo will suffice).

Well you can't link to it directly in the TOC, and we already link to that page in the introduction, so if we want a more detailed page, I think that should come as a separate PR.

I've implemented the rest of the feedback.

markmandel (commented):
> > Let's add an examples page that points to https://github.com/googleforgames/quilkin/blob/main/examples (I'd love to have it point to the same release branch as the snapshot with something like mdbook-variables, but can experiment with that later, I'm sure an explanatory note that they might need to switch to the correct release tag will in the repo will suffice).
> 
> Well you can't link to it directly in the TOC, and we already link to that page in the introduction, so if we want a more detailed page, I think that should come as a separate PR.

Sounds good, I'll jump on it once this has merged 👍🏻 

(Reminder to myself: Update the release guide once this has merged too, to link to the tagged snapshot from README.md)

markmandel (approved):
YAY!

PR #321, opened 2021-07-07
Add naive benchmark

XAMPPRocky:
This adds a naive criterion benchmark suite that runs two separate benchmarks to create a comparison. In both cases we're simply sending packets back and forth, but the first case is the UDP socket talking directly to another socket, and the second case is with Quilkin as a middle-man between the two. It should be strongly noted that this benchmark heavily favours the direct case, as there is zero latency or drop, it mostly is blocked by syscalls, so this is not a fair comparison of Quilkin's performance in the real world, however what it does is provide a baseline for the overhead of processing a packet in Quilkin.

I've provided what it looks like on 2019 MBP and currently Quilkin is about 2/3's slower than reading it directly. (Unit is in μ/secs)

Screenshot 2021-07-07 at 18 02 55


build/README.md
```diff
@@ -50,6 +50,10 @@ To run our external documentation tests:
 
 `cargo +nightly test --doc`
 
+To run our benchmarks:
+
+`cargo bench`
```
markmandel, line 55:
```suggestion
`cargo bench`

We use [criterion](https://github.com/bheisler/criterion.rs) for benchmarking. You can find visual reports under `./target/criterion` for easy comparison.
```

^ small nit, just to make things easier for those not familiar with Rust bechmarking tools.

markmandel (approved):
LGTM! Really like the html reports for comparison, just had a small addition I'd like to see for those not as familiar with Rust tooling.

Side thought: Do we want to set this up to run nightly and dump the reports into a cloud storage bucket somewhere?

Approving so you can merge when ready 👍🏻

XAMPPRocky:
> Side thought: Do we want to set this up to run nightly and dump the reports into a cloud storage bucket somewhere?

Well, if we do we shouldn't try to use it for measuring performance regressions. Criterion specifically [warns against](https://bheisler.github.io/criterion.rs/book/faq.html#how-should-i-run-criterionrs-benchmarks-in-a-ci-pipeline) running on CI environments, because very sensitive to noisy neighbouring. I don't know how much of an issue that would be with GCP.

We could still publish them just to look at, we could add it as part of the deploy-docs job.

markmandel:
> Well, if we do we shouldn't try to use it for measuring performance regressions. Criterion specifically warns against running on CI environments, because very sensitive to noisy neighbouring. I don't know how much of an issue that would be with GCP.

Ah interesting. We're currently running on E2 series VMs, which are not VMs that are specifically built for isolation. I'd be surprised if it was as noisy as a CI platform, but it's not totally isolated. 

Something to think about at least, or maybe experiment with, see what results we get.

build/README.md
markmandel: Just gently bumping this comment, just in case you didn't see it.  Also updated the text, didn't like the "easy comparison" bit.

Issue #330, opened 2021-07-12
Get packet buffer size from operating system


XAMPPRocky:
Currently Quilkin always allocates a `u16::MAX` length buffer for every message, this is incorrect for a number of reasons. Firstly, the max length is slightly less than that as it's really `65535 - (IP + UDP headers)`, further is far too large compared to realistic workloads which are typically in the 200–500 bytes range (to prevent IPv4 fragmentation), and is much bigger than what most OS' set as the limit for UDP payloads. For example; macOS by default has a limit of 9216.

Just testing #321 with different workloads and buffer sizes, having a lower buffer has a significant impact on the performance. For example having a buffer of 1500 bytes had a 15% increase in throughput for smaller workloads. So instead of always using the maximum, we should try to be a bit smarter about how much we allocate. A good default would be to query the OS for the maximum UDP buffer, I also think it would be worth adding a server level option for setting a buffer limit manually allowing you to lower it for when you're using it with a protocol with a smaller MTU.

https://github.com/googleforgames/quilkin/blob/35005f0f0899394a6892062d4f8da3d70fa2a126/src/proxy/server.rs#L240-L242

PR #334, opened 2021-07-15
Move packet buffer to heap

iffyio:
Allocating the 64K buffer on the stack makes the future expensive to move.
This allocates the buffer on the heap instead.

Noticed some significant perf improvements in load tests
via https://github.com/majek/dump/tree/master/how-to-receive-a-million-packets

Tried using the naive benchmark from #321 but that doesn't seem
consisten atm - constantly got perf regressions/improvements on reruns even with
no code change (I think either running both the proxy and the mock server within the same process/scheduler
adds too much noise or we don't have a large enough unit of work)

Relates to #330


XAMPPRocky (approved):
LGTM! Probably worth also setting this one to use `with_capacity` too rather than creating a filled buffer.

https://github.com/googleforgames/quilkin/blob/fa983acdb6f7c27d537101b8efa67c195af041a6/src/proxy/sessions/session.rs#L138

XAMPPRocky:
Oh sorry, it's the other way around, this needs to be `vec![0; 65535]`, because `with_capacity` creates a zero length array so it doesn't fill up any of the buffer.

iffyio:
> Oh sorry, it's the other way around, this needs to be `vec![0; 65535]`, because `with_capacity` creates a zero length vector so it doesn't fill up any of the buffer.

right that explained the perf improvment I saw too heh since we were passing zero sized packets as a result. With the proper fix now perf is pretty much identical in this PR

> 
> How difficult do you think it would be to port this to Rust & criterion? It would be nice to have something more reliable.

Not too sure since I haven't looked much into criterion but I imagine it should be possible to at least, yeah would be nice to have something equivalent that we can use

PR #338, opened 2021-07-15
Make licence explicit in Cargo.toml

markmandel:
Seems like crates.io thinks our licence is "non-standard", so switching `license-file` for `license` in the Cargo.toml.

image


XAMPPRocky:
LGTM, though I'd also file an issue on the crates.io repository if there isn't one already, the licence-file option was added later so crates.io might not be accounting for it.

markmandel:
Excellent suggestion. Doing that now.
https://github.com/rust-lang/crates.io/issues/3784

XAMPPRocky (approved):

Issue #158, opened 2020-12-09
Blocklist Filter


markmandel:
Thinking about Server side DDOS mitigation, or just general abuse mitigation - being able to block specific addresses from sending traffic through to any gameserver should be useful.

Something like:

```yaml
version: v1alpha1
static:
  filters:
    - name: quilkin.extensions.filters.block.v1alpha1.Block
      config:
        addresses:
          - "194.56.36.79"
          - "3FFE:0000:0000:0001:0200:F8FF:FE75:50DF"
  endpoints:
    - name: server-1
      address: 127.0.0.1:7001
```

Question: Not sure if we need to block IP and port, or just IP?

markmandel:
> Since blocklists are such an essential part of proxies, I think it would be worth thinking of the blocklist not as a filter, but as a feature as part of a filter chain, so you'd set the list on that level, and then with #63 filters would be able to add and remove from the blocklist as part of their response.

With #63 - I would suggest, let's take a user journey (this one seems like a good example), and flesh it out to see what we think we would like it to look like. I have some thoughts that think we can keep this quite simple, at least for first pass, but I won't dilute this ticket with those ideas for now.

I don't think we need to have a new special construct for blocklist management, it can work as a Filter, it just depends where it should be in the Filterchain ordering, but also keeps the cognitive overhead for users down -- but doing a user journey discussion on #63 will likely also find where the edge cases are.

markmandel:
I was thinking about this the other day, and wanted to get people's feedback:

1. I think Quilkin should have it's own Allow/Deny features based on addresses. Yes, the end user may also block at a higher level than Quilkin, but if you are doing address based access control, it seems useful to be able to do redundant levels in case one fails. 
2. I'm not 100% convinced it's Quilkin's job to be managing anything outside of itself. But long term, it can push data out to other systems that may do so through rules and actions (when we work out what that looks like).
2. I think it makes sense to combine block and allowlist functionality into a single Filter. 

I've called this `Firewall` since I figure that's basically what it is.

```yaml
version: v1alpha1
static:
  filters:
    - name: quilkin.extensions.filters.debug.v1alpha1.Firewall
      config:
        - direction: Read # ("Read": respond to on_read or "Write" for on_write filter events)
          action: Allow # (action on address/port match. "Allow" would allow traffic through, "Deny" would drop the packets)
          source: "192.168.51.0/24" # (ip4 or ipv6 Cidr)
          ports: # (which ports to match on)
[...]

PR #339, opened 2021-07-19
Refactor Quilkin's data pipeline into distinct types.

XAMPPRocky:
This PR refactors Quilkin's internal proxy structure to more clearly separate responsibilities for handling the "data pipeline" that moves through Quilkin. It achieves this by first pulling that out into its own `Pipeline` type, which is responsible for creating the `DownstreamDistributor` and `UpstreamDistributor` types, which both manage the `Downstream` and `Upstream` types respectively. Most of these types are what was already there, just renamed and reorganised. (`RunRecvArgs` is `DownstreamDistributor`, `Session` is now `Upstream`, etc.).

- The other main motivator was due to how we were creating `Session`, currently a `Session` is essentially created at the end of `process_downstream_received_packet `, which spawns and returns itself to be managed inside `SessionManager`. This is a problem because it essentially requires that all shared upstream state has to also be first passed through the downstream, leading to more boilerplate. Where as now, `UpstreamDistributor` is responsible for creating sessions, and the `Downstream` communicates with the `UpstreamDistributor` to send packets to upstream.
- This PR refactors these components into using what you could consider a rudimentary actor model. Where we construct types that have an inherent `spawn` method, call that method to move their task onto the runtime, and instead of returning itself, it returns a channel to communicate with it (and/or handle to block on completion). I think this pattern lets you more clearly see what's happening, and it had an unintended side-effect of removing about 12 `.clone`s  as there were few cases were we overly cloned an item by one when spawning other tasks.
- `SessionManager` has been removed entirely, as part of this `Upstream` sessions are now be self managed, meaning they track their own expiration, and terminate when they expire. So instead of storing the session itself in a map, we store the channel we use to send messages instead (as mentioned above), and we know that if we can't send a message to that channel, it's expired and we need to insert a new entry.
  - Moving this into its own task with its own message queue removes the need for the `RwLock` on the map, as we messages will just fill the distributor queue while we prune old sessions.
  - Moved the `Server` type into `src/proxy.rs`, and renamed to be `Proxy`, because it is the main type responsible for the proxy, and it is more than one server technically as you can enable the administration interface and the prometheus' metrics server. This also removed some weird visibility cases, like the fields being `pub(super)` to be available to the `Builder`.

---

I haven't migrated the tests yet, because this requires rewriting a lot of the server test, since those tests were unfortunately using a lot of the same internals, but I wanted to get a review on the structure first, and then if it all looks good I'll migrate over the tests and add further documentation.


src/proxy/pipeline/upstream.rs
```diff
@@ -1,232 +1,295 @@
...
+            packet.to_vec(),
+        )) {
+            slog::debug!(
+                self.log,
+                "Sending packet back to origin";
+                "origin" => self.endpoint.address,
+                "contents" => debug::bytes_to_string(&response.contents),
+            );
+
+            if let Err(err) = self.downstream_socket.send(&response.contents).await {
```
iffyio, line 262: ooh I like this change of sending it directly on the socket rather than the previous queue, good catch!

src/proxy/pipeline/upstream.rs
```diff
@@ -1,232 +1,295 @@
...
+
+                sender.send(contents).await.map_err(Error::Sender)?;
+            }
+            Entry::Occupied(mut entry) => {
+                // If we get an error, then the session has expired, so
+                // insert a new upstream.
+                if let Err(_) = entry.get().send(contents).await {
+                    entry.insert((new_upstream)().await?.spawn(shutdown_rx.clone()));
+                }
+            }
```
iffyio, line 124: this seems more like an unwanted edge case rather than intended behavior which would be that the item id removed from the map when the session is no longer in use. it doesn't seem that we delete items from the map at all?

src/proxy/pipeline/upstream.rs
```diff
@@ -1,232 +1,295 @@
...
-impl Packet {
-    pub fn new(dest: SocketAddr, contents: Vec<u8>) -> Packet {
-        Packet { dest, contents }
-    }
+impl UpstreamDistributor {
+    pub fn spawn(
+        self,
+        mut shutdown_rx: ShutdownRx,
+    ) -> (DistributorTx, JoinHandle<Result<()>>) {
+        let manager: dashmap::DashMap<(SocketAddr, SocketAddr), UpstreamTx> =
```
iffyio, line 62: the issue with dashmap is that it isn't async-aware so that any contention would block the waiting threads rather than the current impl that yields (though the current impl locks the whole map). I'm wondering if this is the lesser evil though but not sure what alternatives are available or if we can write something similar ourselves and tackle this in a separate PR? Ideally we want to use a sharded map that yields instead.

iffyio (commented):
I think the PR has two parts, the refactoring/separation-of-concerns and the switch to use channels. The first part sounds good to me but I'm not a fan of the channel changes atm - that adds a bit of indirection with the channel communication compared to regular function calls which I think makes it harder to follow (and adds more locking on the way too) and I'm not really seeing any benefits in doing so?

src/proxy/pipeline/upstream.rs
XAMPPRocky: Yeah, though I believe it should only block on the upstream distributor thread, though your comment made me realise that this whole problem is actually best solved by instead using an LRU cache here, and there are couple of async libraries for that.

src/proxy/pipeline/upstream.rs
XAMPPRocky: Good catch, I'll add that depending on whether or not a LRU abstraction works out.

XAMPPRocky:
I drew a small architecture diagram to hopefully help visualise the structure.

Pipeline 2

src/proxy/pipeline/upstream.rs
iffyio: I don't think a plain lru cache will do here (at least not without redefining what session expiration means), we'd need to limit the size of the map (which could be a nice thing to have configurable in any case) but don't necessarily want to remove older ones when it gets full and won't be able to delete things when they're no longer in use (i.e the cache will remain full even if all sessions are no longer active)

XAMPPRocky: > we'd need to limit the size of the map

Well what I was thinking was just setting the capacity of the to `usize` to `usize::MAX` which effectively would give the cache the same capacity as the map and make it so nothing would expired by size, and it would only really be removing based on expiry.

However I realised, that the locking here is actually needless, because we have unique access to the map in the distributor thread, we don't actually need to provide any kind of locking, so now it just uses `HashMap`.

iffyio (commented):
My take on this is similar to my [previous comment](https://github.com/googleforgames/quilkin/pull/339#pullrequestreview-709773682) - I'm against the channel changes because they're not neccessary afaict and mostly complicate things. With the current changes, looking up a session as well as sending a packet on a session's behalf are both queued oåerations whereas both were previously function calls - I think we'll be misusing channels in both cases since there isn't any concurrency going on that warrants it

@XAMPPRocky could you rather add new commits to ongoing PRs vs force pushing to them? It makes it a bit harder for me to review after the fact since I can e.g no longer see what's changed since previous reviews :pray:

XAMPPRocky:
> I'm against the channel changes because they're not neccessary afaict and mostly complicate things.

I would like to unpack why you think it complicates things, because the goal of this PR is to simplify, and I’d like to show how I think it achieves that. I don’t think it’s a misuse of channels as this is how is actor message passing works which is a well established pattern, and what it simplifies doesn’t have much to do with concurrency, removing the lock is a side benefit. The main advantage of this approach is that we have a clean separation of state between creating an upstream and downstream.

In the current codebase if you want to add new component to the upstream, you have to also add it to the downstream state, because the downstream has to spawn the upstream. To me, this is more complicated and makes it hard to add to, as it leads to a lot redundant passing state between structs. Look at the difference in the current downstream struct definitions for example. Why does downstream need to know `session_ttl`? It doesn’t, it only needs it to create a new session, but that information is not important to downstream, we’re not changing the title we send based on the downstream. So instead having to pass around all of the upstream state to each downstream, we can instead have a single sender which sends only what downstream needs to initialise the upstream.

```rust
// Current
struct ProcessDownstreamReceiveConfig {
    log: Logger,
    proxy_metrics: ProxyMetrics,
    session_metrics: SessionMetrics,
    cluster_manager: SharedClusterManager,
    filter_manager: SharedFilterManager,
    session_manager: SessionManager,
    session_ttl: Duration,
    send_packets: mpsc::Sender<Packet>,
}

// Proposed
struct Downstream {
     cluster_manager: SharedClusterManager,
     filter_manager: SharedFilterManager,
     log: Logger,
     proxy_metrics: ProxyMetrics,
     distributor_tx: DistributorTx,
 }
```

---

> could you rather add new commits to ongoing PRs vs force pushing to them? It makes it a bit harder for me to review after the fact since I can e.g no longer see what's changed since previous reviews 🙏

Absolutely 🙂, it makes no real difference to me, I know some people prefer force pushes because it works well with GitHub’s “viewed” feature as that will track what’s updated since your last review overall including force pushes.

iffyio:
As I see the current impl, the flow is largely sequential - a task receives a packet downstream, looks up the correct session in the map and calls `session.send(packet)`. This is straightforward to reason about and we can read the code linearly to see what goes on. Using a queue breaks this linear flow with indirect communication, it also adds backpressure concerns (now we have to reason about packets in any of the queues in addition to packets being processed, increased memory usage due to buffering, tasks potentially blocking/deadlocking due to queue operations/bugs), also a queue is more expensive than using a lock (and this introduces multiple queues). I get the idea behind communicating via message passing (its in theory possible to replace any function call with it), rather I don't think its worth it in this case.

> In the current codebase if you want to add new component to the upstream, you have to also add it to the downstream state, because the downstream has to spawn the upstream. To me, this is more complicated and makes it hard to add to, as it leads to a lot redundant passing state between structs

That isn't hard though, its a matter of adding a new field to the `ProcessDownstreamReceiveConfig` struct? And the setup sort of happens before the packet flow as well. I don't see this as a huge issue tbh (I'll rather have the setup boilerplate) especially because I expect it rarely changes but definitely if we feel it can be improved upon we should first look to solve such issues individually if we can (refactor/rename things so that they are more-isolated/clearer) e.g `session_ttl` technically doesn't need to be included there today - it could be hardcoded like this PR does and we can figure out what to do with it when we make it configurable in the future.

XAMPPRocky:
> That isn't hard though, its a matter of adding a new field to the ProcessDownstreamReceiveConfig struct? And the setup sort of happens before the packet flow as well. I don't see this as a huge issue tbh (I'll rather have the setup boilerplate) especially because I expect it rarely changes

In the main flow to add new state to upstream you need to add it three places, in `Upstream`, in `Downstream`, and in `DownstreamDistributor`, which are all at one in the code, which on its own wouldn't be too bad. However our unit tests also construct all of these types, and this is the another advantage to using message passing here, we can test `Downstream` without constructing the `Upstream`, we can pass a queue, and read the results of the operation directly off that, making our tests much less brittle.

I'll update the tests now to hopefully show this, since it seems like aside from this message queue, we're generally aligned on this refactor  (though @markmandel hasn't weighed in yet), and these will need updating before merging anyway.

markmandel:
> I'll update the tests now to hopefully show this, since it seems like aside from this message queue, we're generally aligned on this refactor (though @markmandel hasn't weighed in yet), and these will need updating before merging anyway.

Sorry, haven't had the big chunk of time to dig into this yet 😢 I definitely trust y'all on the refactoring parts - it might be easier to go through that on its own PR - since at a quick perusal, seems like some 👍🏻  wins there.

My top level question is: What functionality were you trying to build that you otherwise couldn't with the current architecture? That may help sway which way we should also go architecturally.

XAMPPRocky:
> My top level question is: What functionality were you trying to build that you otherwise couldn't with the current architecture? That may help sway which way we should also go architecturally.

I was implementing the blocklist with a simple event queue, and when I realised it would be more code to make the existing tests work (before adding new tests that actually test the functionality) that was disheartening, I think we should able to add and changes components without it having a large fallout in the rest of the codebase, there's a lot of code that's reimplemented in tests, and unit tests that use other component's private APIs, such that we need to change a lot of code any time components are added, removed, or change APIs. And that's what this PR tries to achieve, every component is decoupled at their logical point and only communicate what's needed to the other components, the prime example (and the part we're discussing) being `Downstream` no longer needs all of `Upstream`'s state to create it, and each `Downstream` is no longer responsible for managing the upstream session map, this logic is instead in the `UpstreamDistributor` which runs in parallel with `DownstreamDistributor`.

markmandel:
> I was implementing the blocklist with a simple event queue

Do you have this in a branch/design doc somewhere, I'd love to see where the complexity comes in -- I have an idea in my head for a blocklist filter, but it's a pretty simple matching system of checking incoming packets `from` field against something like a CIDR range, with some port information as well (as needed), so am wondering what else is needed here?

Or maybe you have a different design in mind for the blocklist? Maybe it would help to flesh that out in #158 some more?

XAMPPRocky:
> I'd love to see where the complexity comes in

Maybe I wasn't being clear, the complexity does not come the blocklist (I've included the implementation to show that), it comes from trying to add this component to Quilkin and getting the project back to a testable state. If you want an example of that you can look at #358 where the actual changes are small, but it's spread across 46 files, which makes adding these simple components cumbersome.

```rust
use std::net::{IpAddr, SocketAddr};

use chrono::Duration;

pub type SyncBlocklist = std::sync::Arc<tokio::sync::RwLock<Blocklist>>;

/// A blocked IP address and/or port.
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
pub struct BlockEntry {
    ip: Option<IpAddr>,
    port: Option<u16>,
    expiration: Option<chrono::DateTime<chrono::Utc>>,
}

impl BlockEntry {
    const BAN_ALL: Self = Self {
        ip: None,
        port: None,
        expiration: None,
    };

    pub fn from_ban_event(
        ip: Option<IpAddr>,
        port: Option<u16>,
        duration: Option<Duration>,
    ) -> Self {
        Self {
            ip,
            port,
            expiration: duration.map(|duration| chrono::Utc::now() + duration),
        }
    }

    fn matches(&self, ip: IpAddr, port: u16) -> bool {
        self.ip.map_or(true, |banned| ip == banned)
            || self.port.map_or(true, |banned| port == banned)
            || self
                .expiration
                .map_or(true, |expiration| expiration > chrono::Utc::now())
    }
}

impl std::fmt::Display for BlockEntry {
    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
        write!(
            f,
            "Block(ip: {}, port: {}, expiration: {})",
            self.ip
                .as_ref()
                .map(IpAddr::to_string)
                .as_deref()
                .unwrap_or("*"),
            self.port
                .as_ref()
                .map(u16::to_string)
                .as_deref()
                .unwrap_or("*"),
            self.expiration
                .as_ref()
                .map(chrono::DateTime::to_string)
                .as_deref()
                .unwrap_or("*"),
        )
    }
}

#[derive(Debug, Clone)]
pub struct Blocklist {
    list: std::collections::HashSet<BlockEntry>,
    /// Specialisation of adding an all `None` block entry to the list, which
    /// effectively stops all traffic.
    block_all_traffic: bool,
    log: slog::Logger,
}

impl Blocklist {
    pub fn new(log: slog::Logger) -> Self {
        Self {
            list: <_>::default(),
            block_all_traffic: false,
            log,
        }
    }

    /// Returns whether the blocklist contains an address or not.
    pub fn contains(&self, addr: &SocketAddr) -> bool {
        if self.block_all_traffic {
            slog::info!(self.log, "All traffic being blocked.");
            true
        } else if let Some(entry) = self
            .list
            .iter()
            .find(|entry| entry.matches(addr.ip(), addr.port()))
        {
            slog::info!(
                self.log,
                "Blocking traffic from: {socket}, matching rule: {entry}",
                socket = addr,
                entry = entry.to_string()
            );
            true
        } else {
            false
        }
    }

    /// Add a new banned entry to the list.
    pub fn add_ban(&mut self, entry: BlockEntry) {
        if entry == BlockEntry::BAN_ALL {
            self.block_all_traffic = true;
        } else {
            self.list.insert(entry);
        }
    }

    /// Remove a entry from the list.
    pub fn remove_ban(&mut self, entry: &BlockEntry) {
        if *entry == BlockEntry::BAN_ALL {
            self.block_all_traffic = false;
        } else {
            self.list.remove(&entry);
        }
    }
}

impl PartialEq for Blocklist {
    fn eq(&self, rhs: &Self) -> bool {
        self.block_all_traffic == rhs.block_all_traffic && self.list == rhs.list
    }
}
```

src/proxy/admin.rs
```diff
@@ -29,25 +29,22 @@ pub struct Admin {
...
+        Self {
             log: base.new(o!("source" => "proxy::Admin")),
             addr,
-            metrics,
             health: Arc::new(heath),
         }
     }
 
-    pub fn run(&self, mut shutdown_rx: watch::Receiver<()>) {
+    pub fn spawn(self, metrics: Arc<Metrics>, mut shutdown_rx: watch::Receiver<()>) {
```
markmandel, line 44: I know these changes happen mid-thought as we are working on a feature, but can we please avoid slipping in these kind of renames and breaks with current conventions that aren't tied to the current PR? (and this is a large PR at that, so it's easy for them to pass through without consensus being formed/proper review). It feels like this is becoming a bit of a habit.

If we have reason to change conventions, then that's fine - but can we tackle that in its own PR/ticket and draw consensus for it. At the very, very least, let's call it out in a PR (if it's small/medium sized PR), so that it can get the appropriate review.

markmandel:
>  it comes from trying to add this component to Quilkin

My confusion comes from me seeing a BlockList as a Filter. Adding a Filter doesn't change any of the architecture. Maybe a meta point of this conversation is, for #158 (or just in general) - there should be more design discussion and general consensus on how a BlockList (or a feature) is implemented before running down the path of implementation.

> If you want an example of that you can look at #358 where the actual changes are small, but it's spread across 46 files, which makes adding these simple components cumbersome.

Ah - the reasoning for this starts to make more sense. This leads me to another question: "What architectural changes are in your mind that are required for upcoming features?" - I think there may be many that you have that haven't been communicated and/or that we've come to consensus on, so we have a very different mental model of the pain levels here. At this point, I'm not seeing much documented need for major architectural shifts that have been agreed on, so most of what I'm seeing here is more code complexity for not a huge amount of gain.

XAMPPRocky:
Going to close this since the code is out of date, and I want to implement some of #411 which will require more changes to this section.

PR #340, opened 2021-07-19
Add `cargo-deny` to CI by adding to `make test`

### LINKED ISSUE #239: Add cargo-deny config
I was reading over the release steps and I noticed there's a step about ensuring dependencies with a specific licence are properly vendored. Embark has already made a tool for checking licences of a dependency. It might be worth automating that step by having a configuration that warns when a dependency with that licence is found in the graph, and then add it to an exception list once it has been archived.

- [Repo](https://github.com/EmbarkStudios/cargo-deny)
- [Documentation](https://embarkstudios.github.io/cargo-deny/checks/index.html)

### LINKED ISSUE #327: Add Cargo Deny to `make test`
We have a #239 cargo deny configuration, but it's not integrated into our CI or local testing tooling.

If we add a new target to [build/Makefile](https://github.com/googleforgames/quilkin/blob/main/build/Makefile) and make it a dependency of the `test` target, it will also be tracked in CI as well.

markmandel:
This adds `cargo-deny` to `make test`, which also makes sure it is run in Cloud Build for continuous integration.

This found a couple of dependencies that need to be updated because of security advisories. These will be updated in a separate PR.

Closes #327

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -49,6 +49,8 @@ version:
 # Run all tests
 test: ensure-build-image
 	docker run --rm $(common_rust_args) \
+     		--entrypoint=cargo $(BUILD_IMAGE_TAG) deny check
+	docker run --rm $(common_rust_args) \
  		--entrypoint=cargo $(BUILD_IMAGE_TAG) clippy --tests -- -D warnings
 	docker run --rm $(common_rust_args) \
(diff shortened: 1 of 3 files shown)


markmandel:
This should pass once #341 has been merged.

XAMPPRocky (approved):

iffyio (approved):

Issue #318, opened 2021-07-07
Create More Complete End-to-End Testing Framework


XAMPPRocky:
Currently our tests are mostly unit tests, and they touch too much of the API surface that when those other parts change it requires updating a lot of the unit tests making these tests quite brittle. One example of this are the tests for metrics, such as some of the session metrics like measuring bytes. These tests manually construct sessions in their tests, but this isn't needed because those details aren't relevant to what we're testing, we'd get the same value from just sending a request and then looking at the prometheus' statistics after the fact.

I want to expand beyond our current ad-hoc helper functions in `test-utils` and have something that is an easier, more declarative way for our network tests, that requires less boilerplate, and only touches the API surface that it tests. In terms of what the output of this would be, I'm not sure yet, but think we should at least start with what about Quilkin's behaviour do we want to test, and what do we want to ensure is correct?

Some starting examples:
- Metrics being sent and being correct.
- Packets completing a full RTT.
- Packets being dropped when conditions are met.
- Filter's specific behaviour.

XAMPPRocky:
So for our end-to-end tests, another thing that came to mind is that, people who write filters also are also going need a way to reliably test their filters, and it would be nice to remove Rust as a requirement for the tests, since not all of our users will be Rust developers. So I was thinking what if we add a new subcommand (`quilkin test`) that runs a single RTT  of a configuration, so you can be confident that everything at least works. In addition we'd add a `test:` key to the root of the config, that provides some test specific config data such as input.

markmandel:
ooh `quilkin test` is an interesting idea! I usually use `nc` and `ncat` to do this manually, but (a) not everyone is on *nix and (b) it can be annoying to setup.

You mention I think two users here, so I wanted to explore that further:

1. The user that is writing a config, and wants to test the config against a released of quilkin.
2. The user that writes their own Filter and wants to test the Filter, while writing the least amount of Rust they have to (maybe they aren't comfortable, or are exploring a solution).

I feel like `quilkin test` would be extremely useful for No. 1, as well as No. 2 -- so we might want to think through both as seperate end users, and make sure we can cover both scenarios.

I would posit that No. 2 will eventually feel more comfortable with Rust as time goes on, and our e2e tooling will probably work for them, but having a lower onrap for bringing people on is always a good thing in my book.

XAMPPRocky:
[...]

Issue #317, opened 2021-07-07
Move from `slog` to `tracing`


XAMPPRocky:
[`tracing`](https://github.com/tokio-rs/tracing) is a diagnostic built and maintained by the Tokio team. It handles logging, but it provides a lot more as it's design for general instrumentation, for example there's a companion crate that uses the output to provide flamegraphs from the events. So using `tracing` would provide us with a lot more insight into what's happening at runtime than our current logging solution.

markmandel:
I'm curious - a while ago I was looking into https://github.com/tikv/pprof-rs to get better visibility into where bottlenecks are in the codebase. My thought being, that then we would't have to manually instrument the codebase. Would something like that be used in conjunction with something like `tracing`, or are they more of competitors?

(I've been meaning to ping you on what tooling you like for performance tracing for Rust).

XAMPPRocky:
> Would something like that be used in conjunction with something like tracing, or are they more of competitors?

Well these two particular libraries are probably competitors right now, because `pprof` uses the backtraces to figure out what happened, where as `tracing` uses crafted events. For our use case (async) I think `tracing` is way forward for quite a few reasons. One, it's what being focused on by the Rust and Tokio developers (you can read more about that here http://blog.pnkfx.org/blog/2021/04/26/road-to-turbowish-part-1-goals/ ). Two, it's like `log` in that provides a clear separation from providing the events to subscribing to the events (it's essentially log + slog + spans). So for example you would be able to create a flamegraph from the same events that you spit out to your logging output, which IMO is a lot better than backtrace based profiling because that can be quite noisy with async code, since the tasks are passed into tokio several layers deep in stack.

markmandel:
From a logging point of view, as long as we still get structured JSON logging, I'm happy 😄, and it looks like https://docs.rs/tracing-subscriber/0.2.19/tracing_subscriber/fmt/format/struct.Json.html or `tracing-serde` will help solve this. (also our INFO and ERROR level logging are quite quiet, so I don't think there is much adjustment needed).

> where as tracing uses crafted events

That's the only concern I really have - how much work does it take to build out crafted events? Could we possibly miss a bottleneck / potential performance improvement spot if we have to manually craft our events?

That all being said - the integration with opentelemetry and distributed tracing is certainly very interesting (I've been thinking a bit about how we could do tracing across proxies, so this seems prescient):
https://github.com/tokio-rs/tracing/blob/master/examples/examples/opentelemetry.rs

[...]

PR #350, opened 2021-08-04
Move I/O and configuration out of runner::run

XAMPPRocky:
This PR is part of #318, but to make the codebase ready to handle additional subcommands a couple of changes were needed.

This moves the CLI and environment configuration outside of `run`, as we'll need these parts to work in both `test` and `run`, this also makes `run` easier to use in tests, since using it doesn't require arguments or environment to be set.

### DIFF
--- a/examples/quilkin-filter-example/src/main.rs
+++ b/examples/quilkin-filter-example/src/main.rs
@@ -15,7 +15,6 @@
  */
 
 use quilkin::filters::prelude::*;
-use quilkin::runner::run;
 
 use bytes::Bytes;
 use serde::{Deserialize, Serialize};
@@ -73,5 +72,5 @@ impl FilterFactory for GreetFilterFactory {
 
 #[tokio::main]
 async fn main() {
-    run(vec![self::factory()].into_iter()).await.unwrap();
+    quilkin::runner::run(vec![self::factory()].into_iter()).await.unwrap();
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 
-use std::io;
-use std::marker::PhantomData;
 use std::net::SocketAddr;
 
 use base64_serde::base64_serde_type;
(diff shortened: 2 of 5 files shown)


src/config.rs
```diff
@@ -48,6 +46,66 @@ base64_serde_type!(Base64Standard, base64::STANDARD);
...
+}
+
+impl Config {
+    /// Attempts to locate and parse a `Config` located at either `path`, the
+    /// `$QUILKIN_CONFIG` environment variable if set, the current directory,
+    /// or the `/etc/quilkin` directory (on unix platforms only). Returns an
+    /// error if the found configuration is invalid, or if no configuration
+    /// could be found at any location.
+    pub fn find(
+        log: &slog::Logger,
```
XAMPPRocky, line 73: Having to accept `Logger` here is unfortunate, but once #317 is implemented, this argument should be able to be removed.

examples/quilkin-filter-example/src/main.rs
```diff
@@ -73,5 +73,8 @@ impl FilterFactory for GreetFilterFactory {
 
 #[tokio::main]
 async fn main() {
-    run(vec![self::factory()].into_iter()).await.unwrap();
+    let log = quilkin::proxy::logger();
+    let config = quilkin::config::Config::find(&log, None).map(Arc::new)?;
+
+    run(log, config, vec![self::factory()].into_iter()).await.unwrap();
```
iffyio, line 79: I'd prefer if we had a oneline version of this so that a user that doesn't have any custom code doesn't need to know about setup bolierplate. I'm thinking similar to the previous `run` that internally creates a logger and finds the config

src/config.rs
```diff
@@ -48,6 +46,66 @@ base64_serde_type!(Base64Standard, base64::STANDARD);
...
+
+        let config_env = std::env::var(ENV_CONFIG_PATH).ok();
+
+        let config_path = path
+            .or_else(|| config_env.as_deref())
+            .or(Some(CONFIG_FILE))
+            .map(|path| std::path::Path::new(path).canonicalize())
+            .transpose()
+            // Path wll always be `Some` here.
+            .map(Option::unwrap)?;
```
iffyio, line 87: how about ensuring the Option<str> is Some first before we canonicalize? something like
```rust
let config_path = std::path::Path::new(path.or_else(...).unwrap_or(CONFIG_FILE)).canonicalize()?;
```
to avoid the unwrap + comment

src/main.rs
```diff
@@ -14,9 +14,37 @@
...
+                .long("config")
+                .value_name("CONFIG")
+                .help("The YAML configuration file")
+                .takes_value(true),
+        )
+        .get_matches();
+
+    let config = quilkin::config::Config::find(&log, matches.value_of("config")).map(Arc::new)?;
+
+    run(log, config, vec![]).await
```
markmandel, line 49: I like this a lot, as then if people desire to have alternate config builders, they are able to do so. 👍🏻

markmandel (commented):
LGTM. @iffyio any issues on your end?

docs/extensions/filters/writing_custom_filters.md
```diff
@@ -133,11 +133,13 @@ To extend Quilkin's code with our own custom filter, we need to do the following
...
-   use quilkin::{filters::DynFilterFactory, runner::run};
+   use quilkin::{config::Config, filters::DynFilterFactory, runner::run};
 
    #[tokio::main]
-   async fn main() {
-       run(vec![Box::new(GreetFilterFactory) as DynFilterFactory]).await.unwrap();
+   async fn main() -> Result<(), Box<dyn std::error::Error>> {
+       let log = quilkin::proxy::logger();
+       let config = Config::find(&log, None).map(std::sync::Arc::new)?;
+       run(log, config, vec![Box::new(GreetFilterFactory) as DynFilterFactory]).await
```
iffyio, line 142: I think the example should use the shorter version of `run` otherwise we'll need to explain the extra arguments which doesn't seem too useful to the tutorial

src/runner.rs
```diff
@@ -14,80 +14,46 @@
...
-fn version() -> String {
-    format!("{}+debug", VERSION)
-}
-
-#[cfg(not(debug_assertions))]
-fn version() -> String {
-    VERSION.into()
+/// Calls [`run`] with the [`Config`] found by [`Config::find`] and the
+/// default [`FilterSet`].
+pub async fn run_default(
```
iffyio, line 35: can we call this `run` and the other variant e.g `run_with_*`?

iffyio (approved):

PR #354, opened 2021-08-08
Remove initial xds update delay

iffyio:
We don't need to explicitly wait for an update from the xds server during startup since we will eventually get one.
So instead, allow the proxy to startup immediately.

### DIFF
--- a/docs/using.md
+++ b/docs/using.md
@@ -15,11 +15,11 @@ The release binary can be downloaded from the
 
 Quilkin needs to be run with an accompanying [configuration file](./proxy-configuration.md), like so:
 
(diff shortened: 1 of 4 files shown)


XAMPPRocky (approved):

docs/using.md
```diff
@@ -15,11 +15,11 @@ The release binary can be downloaded from the
 
 Quilkin needs to be run with an accompanying [configuration file](./proxy-configuration.md), like so:
 
-`quilkin --filename="configuration.yaml"`
+`quilkin --config="configuration.yaml"`
```
markmandel, line 18: That's a good catch 👍🏻

PR #358, opened 2021-08-10
Refactor Endpoint and Metadata from cluster and config into one type.

XAMPPRocky:
This refactors `Endpoint` and `Metadata` from `cluster` and `config` into their own types that can be used in both places. This makes things like creating static configs in code easier, as reduces the amount of boilerplate required to use these types.  This also removes a redundant translation layer where configuration was first parsed into JSON, before being parsed into YAML, this removes that and decodes Prost into YAML types directly.


src/config.rs
```diff
@@ -384,37 +357,20 @@ static:
...
-                    Some(
-                        serde_yaml::to_value(
-                            vec![(
-                                "quilkin.dev",
-                                vec![("tokens", vec!["MXg3aWp5Ng==", "OGdqM3YyaQ=="])]
-                                    .into_iter()
-                                    .collect::<HashMap<_, _>>(),
-                            )]
+                    EndpointMetadata {
+                        tokens: vec!["1x7ijy6", "8gj3v2i"]
```
iffyio, line 363: how come these aren't base64 anymore, what's the change?

XAMPPRocky: Oh right whoops, I forgot I meant to mention that. The change is that the token metadata's internal representation is now the binary of the token, rather than being the binary of the base64 of the token, and the token is now only converted to and from base64 when converting it to or from a different representation such as YAML, JSON, Protobuf.

src/endpoint.rs
```diff
@@ -14,23 +14,141 @@
...
+        }
+    }
+}
+
+/// Metadata specific to endpoints.
+#[derive(Default, Debug, Deserialize, Serialize, PartialEq, Clone, PartialOrd, Eq)]
+#[non_exhaustive]
+pub struct EndpointMetadata {
+    #[serde(with = "base64_set")]
+    pub tokens: base64_set::Set,
```
iffyio, line 71: The rest of the dynamic metadata seems to be missing?

XAMPPRocky: Dynamic metadata is stored in `Metadata<T>`.

src/config.rs
```diff
@@ -24,23 +24,13 @@ use uuid::Uuid;
...
-pub use self::{
-    builder::Builder,
-    config_type::ConfigType,
-    endpoints::{
-        EmptyListError, Endpoints, RetainedItems, UpstreamEndpoints, UpstreamEndpointsIter,
-    },
-    error::ValidationError,
-};
+
+use crate::endpoint::Endpoint;
```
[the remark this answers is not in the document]
XAMPPRocky: > if we let it start importing modules from other places we end up with a cyclic references

This isn't a thing in Rust though. In Rust the module system is separate from imports, you can reference anything from anywhere and it has no impact on the outputted code. In general you want more general types to be higher in the module tree as child modules can reference private items from parents. Which is why this is in its own module, so that it has the same level visibility to the sibling modules `config`, `filters`, and `cluster`.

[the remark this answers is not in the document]
iffyio: I didn't mean that the compiler won't be able to handle it but rather that we will infact be creating cycles if A->importsB->..importsA which is best avoided. I guess its not as weird if they're sibling modules but still feels weird to me having config import endpoint :man_shrugging:

src/endpoint.rs
```diff
@@ -46,12 +164,8 @@ pub struct UpstreamEndpoints {
...
 impl Endpoints {
     /// Returns an [`Endpoints`] backed by the provided list of endpoints.
-    pub fn new(endpoints: Vec<Endpoint>) -> Result<Self, EmptyListError> {
-        if endpoints.is_empty() {
-            Err(EmptyListError)
-        } else {
-            Ok(Self(Arc::new(endpoints)))
-        }
+    pub fn new(endpoints: Vec<Endpoint>) -> Option<Self> {
+        (!endpoints.is_empty()).then(|| Self(Arc::new(endpoints)))
```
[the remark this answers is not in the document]
XAMPPRocky: > returning a result seems more reasonable than a plain None back when things go wrong?

Returning a result isn't useful when there's only two outcomes, it's either valid or it isn't. Also if you look at where `Endpoints::new` was being used in the codebase, it was always being as option like type, the `Result` aspects were never used (error propagation, etc).

src/endpoint.rs
```diff
@@ -14,23 +14,141 @@
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let items = <Vec<String> as serde::Deserialize>::deserialize(de)?;
+        let set = items.iter().cloned().collect::<Set<String>>();
+
+        if set.len() != items.len() {
+            Err(D::Error::custom(
+                "Found duplicate tokens in endpoint metadata.",
+            ))
```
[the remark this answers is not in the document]
XAMPPRocky: Well I think it's better if our validation starts strict, and only relax it with reason. I agree that it seems harmless to just discard them, but if we're getting incorrect input from an external source it seems safer to reject it.

src/proxy/health.rs
```diff
@@ -39,6 +39,7 @@ impl Health {
         let default_hook = panic::take_hook();
         panic::set_hook(Box::new(move |panic_info| {
             error!(log, "Panic has occurred. Moving to Unhealthy");
+            error!(log, "Info: {}", panic_info);
```
iffyio, line 42: we should probably add the panic info to the first error message rather than log a new one?

src/endpoint.rs
```diff
@@ -14,23 +14,142 @@
...
-pub struct EmptyListError;
+use std::{net::SocketAddr, sync::Arc};
 
-#[derive(Debug)]
-pub struct AllEndpointsRemovedError;
+use serde::{Deserialize, Serialize};
 
-#[derive(Debug)]
-pub struct IndexOutOfRangeError;
+type MetadataMap = crate::metadata::Metadata<Metadata>;
```
iffyio, line 23: The name sounds misleading since this isn't a map?

src/endpoint.rs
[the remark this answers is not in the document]
iffyio: SGTM!

src/endpoint.rs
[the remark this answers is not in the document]
iffyio: Yeah I guess in simple cases an option can be alright. Could we not use a boolean expr for this? it reads more cryptic to me in this case than an if/else

markmandel (commented):
Totally forgot to submit my questions / comments from the other day 🤦🏻‍♂️

markmandel (commented):
This LGTM 👍🏻 (outside of the things @iffyio has mentioned).

Nice change!

src/proxy/health.rs
markmandel: As a request: If there are small refactorings or additions to a PR that aren't specifically tied to the overall goal of the PR, it would be nice to either split that out into its own small PR we can do a quick review on, or throw a comment on it when submitting the PR / make a note of it in the description so that it's easy to spot.

Going through the code, I didn't even notice that this PR edited `health.rs`, I was just focusing on things that touched on Endpoints, so I totally missed the change.

(Nice catch @iffyio)

XAMPPRocky: Yeah I didn't even intend to include this, we have an issue where some tests when they panic they hang, and I needed to add this in order to be able figure out what was panicking, after I fixed the issue I forgot about it.

src/endpoint.rs
```diff
@@ -14,43 +14,62 @@
...
-    pub fn new(endpoints: Vec<Endpoint>) -> Result<Self, EmptyListError> {
-        if endpoints.is_empty() {
-            Err(EmptyListError)
-        } else {
-            Ok(Self(Arc::new(endpoints)))
+    pub fn new(endpoints: Vec<Endpoint>) -> Option<Self> {
+        match endpoints.is_empty() {
+            true => None,
+            false => Some(Self(Arc::new(endpoints))),
         }
```
iffyio, line 73:
```suggestion
        if endpoints.is_empty() {
            None
        } else {
            Some(Self(Arc::new(endpoints))),
        }
```
nit: matching on a bool feels overkill

iffyio (approved):

PR #359, opened 2021-08-10
Include endpoint token and metadata from cluster update

iffyio:
Currently we're only passing along the endpoint address
while other info is empty


markmandel:
Nice catch on that one. Should we also add some unit tests here to check that the information is passed all the way through?

iffyio:
> Nice catch on that one. Should we also add some unit tests here to check that the information is passed all the way through?

Added :+1:

src/cluster/cluster_manager.rs
```diff
@@ -158,9 +154,11 @@ impl ClusterManager {
...
                         match update {
                             Some(update) => {
-                                Self::update_cluster_update_metrics(&metrics, &update);
-                                let update = Self::create_endpoints_from_update(&update);
                                 debug!(log, "Received a cluster update.");
+                                let num_clusters = update.len() as i64;
+                                let update = Self::create_endpoints_from_update(update);
+                                let num_endpoints = update.as_ref().map(|ep| ep.as_ref().len() as i64).unwrap_or_default();
+                                Self::update_cluster_update_metrics(&metrics, num_clusters, num_endpoints);
                                 cluster_manager.write().update(update);
```
XAMPPRocky, line 162: Can we move this to its own `handle_cluster_update` function?

XAMPPRocky (approved):
LGTM

src/cluster/cluster_manager.rs
iffyio: Done!

Issue #233, opened 2021-04-28
xDS Example


markmandel:
It would be lovely to have some kind of simple [xDS](https://github.com/googleforgames/quilkin/blob/main/docs/xds.md) example in https://github.com/googleforgames/quilkin/tree/main/examples

One potential idea I had, was to do something on the client side proxy, that adds the hosted endpoint to the client proxy. I thought this might be a simple enough example. Just a thought though.

markmandel:
I'm lumping into this ticket the work converting xDS into Quillkin itself - since it solves both problems at once.

markmandel:
I have this locally with Agones + Quilkin + Xonotic. Just waiting on #574 to review and see if simplifies any of the configuration.

Issue #346, opened 2021-07-22
Reducing Configuration Representation Boilerplate


XAMPPRocky:
Currently most of the boilerplate in writing a filter, is writing a configuration that is statically and dynamically configurable. Right now, you essentially have to write twice, once in Rust and another in Protobuf. I'd like to propose that we try to generate one these representations from the other, so that we can much more easily iterate and add to a filter's configuration. This would come in the form of a tool that is either run separately, or as part of the build process.

- Protobuf -> Rust
  - The issue with this route is that the protobuf parsers in Rust, aren't at the same level quality as the Rust parser libraries. So we'd have to put more work in the parser than I think would be worth it.
- Rust -> Protobuf
  - I think this the most realistic both in terms of the implementation as well as flexibility, as we can do a lot more in Rust than we can in Protobuf.
  
One way we could do this is have a procedural macro for the configuration struct, that generates protobuf files from their type  definitions, and then dump the generated definition into the folder, so you'd have to run cargo build after changing a configuration struct to generate the update protobuf, but that seems reasonable and in return we wouldn't have to manually edit protobuf definitions.

```rust
// src/lib.rs
quilkin::generate_config_protobufs!("proto/quilkin");

// src/filters/capture_bytes.rs

#[derive(quilkin::ConfigEnum)]
enum Strategy {
    #[quilkin(rename = "PREFIX")]
    /// Looks for the set of bytes at the beginning of the packet
    Prefix = 0,
    #[quilkin(rename = "SUFFIX")]
    /// Look for the set of bytes at the end of the packet
    Suffix = 1,
}

#[derive(quilkin::Config)]
struct Config {
    #[quilkin(default)]
    strategy: Strategy,
    /// the number of bytes to capture
    #[quilkin(rename = "size")]
    size: usize,
    /// the key to use when storing the captured bytes in the filter context
    #[quilkin(rename = "metadataKey")]
    #[quilkin(default = "default_metadata_key")]
    metadata_key: String,
    /// whether or not to remove the set of the bytes from the packet once captured
    #[quilkin(default = "default_remove")]
    remove: bool,
}
```

markmandel:
I love this idea, because the double definition of models for each config type annoys me no end. 🥳 

I'm also being the `Rust -> Protobuf` - the other thing we could do is have some opinionated defaults on how the protobuf gets generated (or maybe even start with not even letting people have the option).

[...]

PR #360, opened 2021-08-10
XDS Management Server

### LINKED ISSUE #131: Implement a control plane
The XDS control planes out there today seem to be directed at proxying HTTP and TCP traffic so that it seems simpler to roll our own rather than attempt to adopt one of them to work for our use case.
[go-control-plane](https://github.com/envoyproxy/go-control-plane) seems to make writing an XDS control plane really easy thankfully - it handles running a GRPC server that speaks the XDS protocol with proxies and is backed by a [cache](https://github.com/envoyproxy/go-control-plane#resource-caching) which an implementation needs to populate.

The rough workflow would be that we have our code find out what gameserver/upstreamendpoints are available in the cluster and update the cache. Each quilkin proxy will watch resources by contacting go-control-plane's GRPC server which then feeds it data from the cache whenever it is updated.

Currently wondering what a simple workflow would be with e.g Agones + OpenMatch - say a loop that watches `GameServer`s that have been `Allocated` and populates the cache with their addresses as Endpoints - and it sounds like one difference in this case is that we'll need to ensure that all connected proxies have acknowledged the update containing a `GameServer`'s address before Open Match lets any game client talk to them otherwise a race condition can cause any initial packets from clients to be dropped since the proxies won't know about the new address.
So that this would need some kind of synchronization between the control plane and Open Match? (Say a CRD that will be updated by the control plane server and watched by the director?)

Would this make sense? Is this missing something? Thoughts?

iffyio:
This adds a management server based on go-control-plane that fetches
resources (clusters, filters) from some source and pushes that config
to all connected proxies.

the original plan was to write a tiny server that we can run tests against, where quilkin takes in config from an actual server. Quickly realised the implementation is simple enough that it covers both test and actual scenarios so I figure it'll be the same project (same for fulfulling #233)

The server itself is implemented and works, it can only fetch and live reload resources from disk but adding support for other sources like k8s is a matter of adding an interface implementation that can talk to that source.

Some things are left todo though, so I'm leaving this as a draft in the meantime but figured to gather thoughts/feedback

- [X] ~need to figure out where this should live, either here or in a separate repo and~ have some CI for it too.
- [X] Only debug filter is currently supported. Would love some thoughts on how we want to handle generated proto code, currently I checked in the generated code but not sure if there's a saner approach with go - will port over other filters when its clearer how we want to do this
- [X] no tests ^^ tested manually though, will follow up with unit tests before removing from draft

closes #131


XAMPPRocky (approved):
LGTM, though obviously I have very little opinion on the Go code. 😄 

> need to figure out where this should live, either here or in a separate repo and have some CI for it too.

I think this should live here if possible, only if we want to have separate releases, or it would be difficult to build and release both, should we have a separate repo. It being here simplifies overhead and lets us do codegen more easily.

> Would love some thoughts on how we want to handle generated proto code, currently I checked in the generated code but not sure if there's a saner approach with go - will port over other filters when its clearer how we want to do this

I think this is related to #346 and I'll write up some more thoughts there.

markmandel:
> I think this should live here if possible, only if we want to have separate releases, or it would be difficult to build and release both, should we have a separate repo. It being here simplifies overhead and lets us do codegen more easily.

I concur on this one. Let's keep things in a single repo until we hit such pain that it forces us to move it out 👍🏻 

> Would love some thoughts on how we want to handle generated proto code

Agree on more discussion with #346, but no issue on my end with checking in generated Go code. There's not `build.rs` in Go, so I've no concerns with vendoring this type of code. (I've done this a bunch).

This is super cool!

examples/xds-management-server/pkg/resources/resources.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func makeFilterChain(
+	filterChainResource []FilterConfig,
+) (*envoylistener.FilterChain, error) {
+	var filters []*envoylistener.Filter
+
+	for _, resource := range filterChainResource {
+		switch resource.Name {
+		case "quilkin.extensions.filters.debug.v1alpha1.Debug":
```
markmandel, line 169: 🤔 I wonder if there is a way to make this generic without having to be explicit about type.

Just figuring that If this has a static list of Filters and accompanying models here, then each time a user / we release a new build of Quilkin with a new Filter, we have to update the xDS code and deployment -- if we could be generic here, it might be easier in the long term.

But I can't think of a good way to do that 😦 as per #346 maybe we can generate some kind of go map/yaml/json with accompanying code or something that makes this piece of code do a lookup, rather than have to implement a giant switch. (Thinking out loud at this point).

markmandel (commented):
Some quick feedback on Go:

Running `goimports` over the code, results in a few sorting/formatting changes:

```bash
➜  xds-management-server git:(pr/iu/management-server) goimports -w .
➜  xds-management-server git:(pr/iu/management-server) ✗ gst
On branch pr/iu/management-server
Your branch is up to date with 'upstream/iu/management-server'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   filters/debug/v1alpha1/debug.pb.go
        modified:   pkg/resources/resources.go
```

Running my favourite mega-linter over the code, resulting in one issue:

```bash
➜  xds-management-server git:(pr/iu/management-server) golangci-lint run
pkg/resources/resources.go:17:2: SA1019: package github.com/golang/protobuf/jsonpb is deprecated: Use the "google.golang.org/protobuf/encoding/protojson" package instead. (staticcheck)
        "github.com/golang/protobuf/jsonpb"
        ^
```

Would also love to see some docstrings on the Go functions, especially the public ones 😄 

But this is very cool to see in action 👍🏻

examples/xds-management-server/pkg/resources/resources.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/envoyproxy/go-control-plane/pkg/cache/types"
+	"github.com/envoyproxy/go-control-plane/pkg/cache/v3"
+	"github.com/golang/protobuf/jsonpb"
+	structpb "github.com/golang/protobuf/ptypes/struct"
+
+	quilkinfilter "quilkin.dev/xds-management-server/filters/debug/v1alpha1"
+)
+
+// Endpoint represents an xds endpoint
+type Endpoint struct {
```
markmandel, line 24: Coming from comments in https://github.com/googleforgames/quilkin/issues/346#issuecomment-898384422 - I'm trying to work out what Endpoint, Cluster, FilterConfig, Resource etc are for.

I _think_ it's so that we can specify in a type what our subset of supported xDS is, and thereby not have to deal with all of xDS types in our system?

And then the `make*` commands do all the work to convert our xDS surface into the bigger, more complicated xDS types.

Is that correct?

examples/xds-management-server/pkg/resources/resources.go
iffyio: I did some digging and reworked this now PTAL! now we capture filter configs as a generic `interface` object and marshal it directly into an xds filter so we don't need to know which particular filter it is (and not switching/lookup to do either). Only needed to [change the config to match the xds proto](https://github.com/googleforgames/quilkin/blob/0d9a3111d1975c8c1a2ccafce00de83ceb1c73b5/examples/xds-management-server/config.json#L17-L23) in order to make it directly serializable but its the same format istio uses as well to describe envoy filter configs so its not too awkward at least https://istio.io/latest/docs/reference/config/networking/envoy-filter/

examples/xds-management-server/pkg/resources/resources.go
```diff
@@ -0,0 +1,199 @@
...
+	var filters []*envoylistener.Filter
+
+	for _, resource := range filterChainResource {
+		switch resource.Name {
+		case "quilkin.extensions.filters.debug.v1alpha1.Debug":
+			configBytes, err := json.Marshal(resource.Config)
+			if err != nil {
+				return nil, fmt.Errorf("failed to JSON marshal filter config: %w", err)
+			}
+			var filterConfig debugFilterConfig
```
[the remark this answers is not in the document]
iffyio: Yes that made more sense! Removed the config in the latest commit

examples/xds-management-server/pkg/resources/resources.go
iffyio: Yes that was the idea, so that we only have one place to do the mapping to xds types.

examples/xds-management-server/cmd/controller.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/alecthomas/kong"
+	log "github.com/sirupsen/logrus"
+
+	"quilkin.dev/xds-management-server/pkg/providers"
+	"quilkin.dev/xds-management-server/pkg/server"
+	"quilkin.dev/xds-management-server/pkg/snapshot"
+)
+
+type flags struct {
+	File string `name:"file" help:"Resource config file path." type:"path" default:"config.json"`
```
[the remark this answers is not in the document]
iffyio: Done!

examples/xds-management-server/cmd/controller.go
```diff
@@ -0,0 +1,94 @@
...
+
+	logger := &log.Logger{}
+	logger.SetOutput(os.Stdout)
+	logger.SetLevel(log.DebugLevel)
+	logger.SetFormatter(&log.JSONFormatter{})
+
+	ctx, shutdown := context.WithCancel(context.Background())
+	defer shutdown()
+
+	k8sConfig, err := k8s.GetK8sConfig(logger, flags.KubeHost, flags.KubeConfigPath)
```
markmandel, line 42: Fun question for this: Does this mean there will be one xDS server per Kubernetes cluster?

An interesting assumption 😁 (lots of pros and cons I'm sure) - we should probably make a note of this in the README though.

Not got any strong opinions, especially as this is an example, but just highlighting it so we make a note of it somewhere in some documentation.

examples/xds-management-server/pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,119 @@
...
+	clusterCh <-chan []cluster.Cluster,
+	filterChainCh <-chan filterchain.ProxyFilterChain,
+	snapshotCache cache.SnapshotCache,
+	updateInterval time.Duration,
+) {
+	logger = logger.WithFields(log.Fields{
+		"component": "SnapshotUpdater",
+	}).Logger
+
+	updateTicker := time.NewTicker(updateInterval)
```
markmandel, line 53: Non blocking suggestion: Could use something like github.com/jonboulle/clockwork to mock out time, so you can control the flow of time.

examples/xds-management-server/pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,119 @@
...
+	log "github.com/sirupsen/logrus"
+
+	"quilkin.dev/xds-management-server/pkg/resources"
+)
+
+// RunSnapshotUpdater starts a goroutine that listens for resource updates,
+// uses the updates to generate an xds config snapshot and updates the snapshot from the provided channel and generates xds config snapshots.
+// cache with the latest snapshot for each connected node.
+// It returns the snapshot cache which can be passed on to the xds server impl.
+func RunSnapshotUpdater(
```
markmandel, line 20: Looks like everywhere else (still digging in, but I think this is true), you've used the pattern of New() to initalise and Run() to start the go routine (I really like this pattern, which you have likely noticed 😁)

For the sake of consistency, I think it would be nice to stick to that here too.

Also, if you choose to do the time mock, it's easier to overwrite the local clock variable at test time without having to change the constructor.

examples/xds-management-server/pkg/k8s/k8s.go
```diff
@@ -0,0 +1,52 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+
+	//  load the gcp plugin (required to authenticate against GKE clusters from outside the cluster)
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+)
+
+// GetK8sConfig finds and returns a kubernetes config
+func GetK8sConfig(logger *log.Logger, kubeHost, kubeConfigPath string) (*rest.Config, error) {
```
markmandel, line 17: Fairly sure https://pkg.go.dev/k8s.io/client-go/tools/clientcmd will do all of this 😃 (see the example at the top) - although it uses KUBECONFIG as the env var, which is likely [what you want anyay](https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#set-the-kubeconfig-environment-variable)

examples/xds-management-server/pkg/server/server.go
```diff
@@ -0,0 +1,63 @@
...
+	"google.golang.org/grpc"
+)
+
+const (
+	grpcMaxConcurrentStreams = 1000000
+)
+
+// Run starts a go-control-plane, xds grpc server in a background goroutine.
+// The server is bounded by the provided context.
+func Run(
```
markmandel, line 22: I mentioned this in snapshot.go - but I'd love to see this split into a New(args...) and Run(ctx) pattern. It'll mean you will need a new struct - but I expect it will make it easier for our end users to extend in the long run.

examples/xds-management-server/config.yaml
```diff
@@ -0,0 +1,15 @@
+clusters:
```
markmandel, line 1: Add a comment at the top, stating that this is the source for a file provider?

I was staring at this for a bit trying to work out what it was for 😁

examples/xds-management-server/pkg/providers/file.go
```diff
@@ -0,0 +1,309 @@
...
+// NewFileProvider creates a new FileProvider.
+func NewFileProvider(configFilePath string, proxyIDCh chan string) *FileProvider {
+	return &FileProvider{
+		configFilePath: configFilePath,
+		proxyIDCh:      proxyIDCh,
+	}
+}
+
+// Run runs the FileProvider.
+func (p *FileProvider) Run(ctx context.Context, logger *log.Logger) (
```
markmandel, line 46: nit: it would probably be cleaner to have logger be passed to `NewfileProvider` and it could store it on itself.

Then you have `Run(ctx)`, which is nice and neat 👍🏻

examples/xds-management-server/pkg/providers/file.go
```diff
@@ -0,0 +1,309 @@
...
+	<-chan []cluster.Cluster, <-chan filterchain.ProxyFilterChain, <-chan error) {
+	clusterCh := make(chan []cluster.Cluster, 10)
+	filterChainCh := make(chan filterchain.ProxyFilterChain, 10)
+	resourcesCh := make(chan resources)
+
+	errorCh := make(chan error)
+
+	go runUpdater(
```
markmandel, line 54: A thought: Both `runUpdater` and `runConfigFileProvier` could be methods on `p` - that would save you passing in both the `logger` and `proxyIDCh` since they would be accessible.

examples/xds-management-server/pkg/providers/file.go
```diff
@@ -0,0 +1,309 @@
...
+	proxyIDCh <-chan string,
+	resourcesCh <-chan resources,
+	clusterCh chan<- []cluster.Cluster,
+	filterChainCh chan<- filterchain.ProxyFilterChain,
+) {
+	defer func() {
+		close(clusterCh)
+		close(filterChainCh)
+	}()
+	ticker := time.NewTicker(100 * time.Millisecond)
```
markmandel, line 84: Non blocking: 🤔 each time I see this I want to reach for a time mock like github.com/jonboulle/clockwork. (mention this in snapshot.go as well). You're processing relatively fast, so controlling time is likely not an issue during unit tests, but something to consider.

(Possibly another reason to make `runUpdater` a method on `FileProvider`?)

examples/xds-management-server/pkg/cluster/agones/agones.go
```diff
@@ -0,0 +1,169 @@
...
+			IP:       gs.Status.Address,
+			Port:     int(getGameServerPort(gsLogger, gs.Status.Ports)),
+			Metadata: metadata,
+		})
+	}
+
+	return endpoints
+}
+
+func getGameServerPort(logger *log.Entry, ports []agonesv1.GameServerStatusPort) int32 {
```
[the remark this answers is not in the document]
markmandel: Oooh, that's interesting. So each GameServer can have any number of ports open. There's really nothing other than name that might indicate which is the one that Quilkin should connect people to (or if quilkin should somehow route to multiple ports???).

I think for the example we should have a convention (this one seems fine) and just make sure we document it.

meta question: Does Agones ports need metadata labels? 🤔

examples/xds-management-server/pkg/cluster/agones/agones.go
```diff
@@ -0,0 +1,169 @@
...
+// Run spawns a goroutine in the background that watches Agones GameServers
+// and exposes them as endpoints via the returned Cluster channel.
+func (p *Provider) Run(ctx context.Context) (<-chan []cluster.Cluster, error) {
+	gameServerListWatch := cache.NewListWatchFromClient(
+		p.agonesClient.AgonesV1().RESTClient(),
+		"gameservers",
+		p.config.GameServersNamespace,
+		fields.Everything())
+
+	gameServerInformer := cache.NewSharedInformer(gameServerListWatch, &agonesv1.GameServer{}, 0)
```
markmandel, line 56: I would create a SharedInformerFactory ([example](https://github.com/googleforgames/agones/blob/main/cmd/controller/main.go#L161), and then use the event system to track when `GameServers` are added an deleted - rather than polling every 100ms. Probably only need Update and Delete events. ([example](https://github.com/googleforgames/agones/blob/main/pkg/gameservers/controller.go#L155-L165))

Then you can also use a `Lister` to query your GameServers (even if you want to get the full list on each run), and you can know it's backed by the local cache and you aren't hammering on the K8s api.

You might want to create the SharedInformerFactory in controller.go and pass it into the constructor for the agones Provider -- that way you can build out fakes and mock the operations in tests.

markmandel (changes requested):
Ooh, this is so neat. Apologies it took me a while to dig into it.

Note - each of these files need an Apache header please.

I'm digging through it - I think we need a README to cover what it is doing and how to use it / use it as a base etc.

I expect this will need another pass or two, since it's a chunky PR, and I have to run to do some other things, but wanted to get you at least this feedback.

This is exciting stuff!

examples/xds-management-server/cmd/controller.go
```diff
@@ -0,0 +1,94 @@
+package main
```
markmandel, line 1: organisation thought:

Should this be under `/cmd/agones/agones.go` (since it's an agones provider)

Then it's inline with `/cmd/file/file.go` and consistent.

examples/xds-management-server/cmd/controller.go
```diff
@@ -0,0 +1,94 @@
...
+
```
markmandel, line 2: Thought for next time: The Agones integration could have been a seperate PR after the file provider had been merged. Might have cut down on the 4,000 lines 😃

examples/xds-management-server/pkg/cluster/agones/agones.go
iffyio: I'm not following, why do we want to track the events ourselves? we'd still need to build up a cache and check periodically which the cache.Store already does for us? Also I thought the sharedInformer store is a cache :open_mouth: where's the k8s api hammering occurring here?

examples/xds-management-server/cmd/controller.go
iffyio: Not exactly, file.go is a different binary because its more of a demo use case (e.g proxy specific filterchain doesn't make sense there). While the main binary isn't specific to agones or k8s as we add more providers they'll still be in the same binary only with different configs

examples/xds-management-server/cmd/controller.go
iffyio: Its not an assumption, rather this is an mvp and what I _think_ we need to test things out on our end. If it turns out we do need multiple clusters I'll add it but otherwise I'll circle back to it eventually or PR welcome if someone else wants to. We can create an issue to track it if we want

examples/xds-management-server/cmd/controller.go
markmandel: Ah got it - that totally makes sense 👍🏻 I think we should make that clear in a README though, just so people understand which piece is which.

examples/xds-management-server/pkg/cluster/agones/agones.go
markmandel: > I'm not following, why do we want to track the events ourselves? 

My thought with using events, rather than polling - was that then we end up with less no-change events going out to xDS subscriptions, and/or we don't have to track ourselves that changes have been made between each poll ourselves either.

> where's the k8s api hammering occurring here?

Sorry, I wasn't clear -- my concern was that if we somehow misconfigured the cache (although in retrospect that seems unlikely).

The other thing I like about a `SharedInformerFactory` is that you won't have to do all that work to setup the cache, it will do all that for you, so I think it'll make the code simpler.

Also, a `SharedInformerFactory` is a standard way of implementing K8s controllers / interacting with K8s, so it's an API surface that people are used to.

markmandel:
This is looking great 😃

markmandel (commented):
THis is looking great 😃

examples/xds-management-server/pkg/cluster/agones/agones.go
iffyio: > My thought with using events, rather than polling - was that then we end up with less no-change events going out to xDS subscriptions, and/or we don't have to track ourselves that changes have been made between each poll ourselves either.

How do we end up with no-change events going to XDS in the current impl? I'm still not entirely sure I understand the concern correctly but the idea is that we send an XDS update only when the cluster changes (not every gameserver event translates to a cluster change). not sure how we'd do this without remembering the previous cluster state?
Then by periodically checking for updates to send to XDS, we have control over how often we're broadcasting to the proxies (so we batch updates to them) rather than triggering a broadcast for every gameserver change (which we can't control).

examples/xds-management-server/cmd/controller.go
markmandel: Makes perfect sense 👍🏻 just want to make sure we cover this in a README.

examples/xds-management-server/pkg/cluster/agones/agones.go
markmandel: > Then by periodically checking for updates to send to XDS, we have control over how often we're broadcasting to the proxies (so we batch updates to them) rather than triggering a broadcast for every gameserver change (which we can't control).

Oh that is an interesting point! That being said, when would we _not_ want it to be updated in as close to realtime as possible? 🤔 

> but the idea is that we send an XDS update only when the cluster changes

I didn't manage to catch this code on first pass. I'll go hunting for it again.

If we did use the event system, we could track GameServer `add` and `delete` events - which would always be a change in endpoints (we could ignore updates - as GameServer's can't change their IP or ports). On each of those events you would then know that the GameServer has been added or removed, so one could send a delta event rather than a full snapshot, without having to do your own comparison of snapshots. (or do a delta to the internal snapshot, and send at that time).

You wouldn't have to check that the stored snapshots had changed then, as you know that there is a definite change.

The downside here -- we may want to do an full update of all GameServer anyway on xDS server startup / just in case - so we may end up in the same place.

Not 100% wedded to this idea, but just wanted to explore to see if it made things easier / implementation simpler.

iffyio: > we could ignore updates - as GameServer's can't change their IP or ports

Tokens could change too which would be update events.

> On each of those events you would then know that the GameServer has been added or removed, so one could send a delta event rather than a full snapshot

The xds snapshot cache itself currently can only be updated with full snapshots rather than delta events. and with the xds delta protocol too, when a new node joins we need to send a full snapshot to it first before any deltas.

markmandel: > Tokens could change too which would be update events.

Ah yep, I see that now.

> The xds snapshot cache itself currently can only be updated with full snapshots rather than delta events. 

Aaah, in which case, there isn't much gain from using events then. Then I recind my suggestion to use events.

I do think we should use a [SharedInformerFactory](https://pkg.go.dev/agones.dev/agones@v1.17.0/pkg/client/informers/externalversions#NewSharedInformerFactory) - it's less setup to create, and means you can also share the cache if that becomes a requirement in the future.

I found a nice SO post on breaking down the various layers of abstraction:
https://stackoverflow.com/a/59544592

examples/xds-management-server/pkg/cluster/agones/agones.go
```diff
@@ -0,0 +1,169 @@
...
+			continue
+		}
+
+		numPorts := len(gs.Status.Ports)
+		if numPorts == 0 {
+			continue
+		}
+
+		var metadata map[string]interface{}
+		tokenStr, found := gs.Annotations["quilkin.dev/tokens"]
```
markmandel, line 133: This is probably more flexible, but another thought could be to use `GameServer.status.players.ids`.

https://agones.dev/site/docs/reference/agones_crd_api_reference/#agones.dev/v1.PlayerStatus

(It is alpha, but thought I'd float the idea.)

examples/xds-management-server/pkg/filterchain/k8s/k8s.go
```diff
@@ -0,0 +1,188 @@
...
+	informerFactory := informers.NewSharedInformerFactoryWithOptions(
+		k8sClient,
+		0,
+		informers.WithNamespace(podNamespace),
+		informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			options.LabelSelector = labelSelectorProxyRole
+		}),
+	)
+	podInformer := informerFactory.Core().V1().Pods().Informer()
+	go podInformer.Run(ctx.Done())
```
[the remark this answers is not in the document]
markmandel: Just touching base to see if this solved your issue, or maybe this is something I can help with over a video call?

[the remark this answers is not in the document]
iffyio: Ah I haven't found time to try it out yet, will make some time soon to get back to this!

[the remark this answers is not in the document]
iffyio: Thanks for the write up! I've updated the pr now PTAL

examples/xds-management-server/cmd/controller.go
```diff
@@ -0,0 +1,158 @@
...
+		informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			options.LabelSelector = k8sfilterchain.LabelSelectorProxyRole
+		}),
+	)
+	informerFactory.Start(ctx.Done())
+
+	return k8sfilterchain.NewProvider(
+		logger,
+		clock.RealClock{},
+		informerFactory.Core().V1().Pods().Lister(),
```
markmandel, line 100: So you use the Lister here, but not above 😄 Would be good to be consistent between the two.

examples/xds-management-server/pkg/cluster/agones/agones.go
```diff
@@ -0,0 +1,167 @@
...
+	log "github.com/sirupsen/logrus"
+	"k8s.io/client-go/tools/cache"
+	"quilkin.dev/xds-management-server/pkg/cluster"
+)
+
+// Provider implements the Provider interface, exposing Agones GameServers as endpoints.
+type Provider struct {
+	config     Config
+	logger     *log.Logger
+	gsInformer cache.SharedIndexInformer
```
markmandel, line 35:
```suggestion
	gsLister agonesv1.GameServerLister
```

examples/xds-management-server/pkg/cluster/agones/agones.go
```diff
@@ -0,0 +1,167 @@
...
+		clusterCh)
+
+	return clusterCh, nil
+}
+
+func runClusterWatch(
+	ctx context.Context,
+	logger *log.Logger,
+	gameServersPollInterval time.Duration,
+	gameServerStore cache.Store,
```
markmandel, line 74:
```suggestion
	gsLister agonesv1.GameServerLister,
```

examples/xds-management-server/pkg/cluster/agones/agones.go
```diff
@@ -0,0 +1,167 @@
...
+		case <-ctx.Done():
+			logger.Debug("Exiting cluster watch loop: context cancelled")
+
+		}
+	}
+}
+
+func getEndpointsFromStore(
+	logger *log.Logger,
+	gameServerStore cache.Store,
```
markmandel, line 105:
```suggestion
	gsLister agonesv1.GameServerLister,
```

examples/xds-management-server/pkg/cluster/agones/agones.go
```diff
@@ -0,0 +1,167 @@
...
+) []cluster.Endpoint {
+	gameServers := gameServerStore.List()
```
markmandel, line 107:
```suggestion
	gameServers := gsLister.List()
```

examples/xds-management-server/pkg/cluster/agones/agones.go
```diff
@@ -0,0 +1,167 @@
...
+
+	var endpoints []cluster.Endpoint
+	for i := range gameServers {
+		gs := gameServers[i].(*agonesv1.GameServer)
```
markmandel, line 111:
```suggestion
		gs := gameServers[i]
```

markmandel (commented):
All the suggestions I left are about using a Lister rather than an Informer inside the Agones controller code 👍🏻

markmandel (approved):
🔥 😄

PR #363, opened 2021-08-12
Update doc links

iffyio:
Updated missing links in docs.

For links to api references I imagine we don't want to hardcode the version and instead set that based on the current version? for now though I link to the current version since it doesn't seem there's a simple way to do that with mdbook. I found this that we might need to use https://crates.io/crates/mdbook-variables - it sets variables that's substituted during doc generation  - though I had to update it locally to get it to work since its using a pretty old mdbook crate.

Also it seems something about the current deployed docs is broken when rendering rust code blocks: check this for example
https://googleforgames.github.io/quilkin/main/book/filters/capture_bytes.html#configuration-examples
that should only show the yaml and the rust bits should be hidden, it renders properly though when I build locally so maybe we've done something special with the deployed one


docs/src/filters/writing_custom_filters.md
```diff
@@ -392,14 +387,13 @@ However, it usually contains a Protobuf equivalent of the filter's static config
...
-[runner]: #
-[create-filter-args-config]: #CreateFilter::config
-[config-type-dynamic]: #ConfigType::Dynamic
+[Filter]: https://docs.rs/quilkin/{{version}}/quilkin/filters/trait.Filter.html
+[FilterFactory]: https://docs.rs/quilkin/0.1.0/quilkin/filters/trait.FilterFactory.html
+[filter-factory-name]: https://docs.rs/quilkin/0.1.0/quilkin/filters/trait.FilterFactory.html#tymethod.name
+[FilterRegistry]: https://docs.rs/quilkin/0.1.0/quilkin/filters/struct.FilterRegistry.html
+[runner::run]: https://docs.rs/quilkin/0.1.0/quilkin/runner/fn.run.html
+[CreateFilterArgs::config]: https://docs.rs/quilkin/0.1.0/quilkin/filters/prelude/struct.CreateFilterArgs.html#structfield.config
+[ConfigType::dynamic]: https://docs.rs/quilkin/0.1.0/quilkin/filters/enum.ConfigType.html#variant.Dynamic
```
XAMPPRocky, line 396: This is why I think we should move this stuff to rustdoc, it would always go to the correct version.

XAMPPRocky (approved):

docs/src/filters/writing_custom_filters.md
markmandel: We can also just use relative links to our local versions of the API docs - since they are published together.

docs/src/filters/writing_custom_filters.md
```diff
@@ -392,14 +387,13 @@ However, it usually contains a Protobuf equivalent of the filter's static config
...
 
-[Filter]: #
-[FilterFactory]: #
-[filter-factory-name]: #FilterFactory::name
-[FilterRegistry]: #
-[FilterChain]: #
-[runner]: #
-[create-filter-args-config]: #CreateFilter::config
-[config-type-dynamic]: #ConfigType::Dynamic
+[Filter]: https://docs.rs/quilkin/0.1.0/quilkin/filters/trait.Filter.html
```
markmandel, line 390: So that it always stays inline with the released version, we can link directly to our local versions of the api docs, but relatively.

```suggestion
[Filter]: ../../api/quilkin/filters/trait.Filter.html
```
(I think that's the right number of redirects, might want to confirm on the [`gh-pages` branch](https://github.com/googleforgames/quilkin/tree/gh-pages))

And repeat as per below.

Then we never have to update the version 🙂

Because of this, created #365 - to make it easier to preview this locally.

iffyio: do we have any means of versioning what we deploy with mdbook? e.g if we release a 0.2.0 and later a 0.3.0 as far as the mdbook doc is concerned 0.2.0 version no longer existed right? or is that kept around still?

docs/src/filters/writing_custom_filters.md
XAMPPRocky: They aren't published together at mdbook build/test time though, so that would fail any linkchecking.

markmandel: We could also use a seperate link checker - we don't have to be tied just to mdbook's link checker - then we can check links across the entire documentation set. This will be particularly useful as we will probably end up with links going back and forth between rustdoc and mdbook.

I'm particularly partial to https://github.com/wjdp/htmltest for link checking across an entire website.

docs/src/filters/writing_custom_filters.md
markmandel: Yep! @XAMPPRocky did an awesome job there - the github action runs on each tag run ([source](https://github.com/googleforgames/quilkin/blob/main/.github/workflows/docs.yml)), and grabs the tag/branch name and makes it the parent folder (so development is in `main`, but when we make a v0.2.0 release we'll get a `v0.2.0` parent folder).

I tested it out on a fork, and it worked super nice 👍🏻

docs/src/filters/writing_custom_filters.md
markmandel: Created https://github.com/googleforgames/quilkin/issues/367 to track the discussion!

docs/src/filters/writing_custom_filters.md
```diff
@@ -392,17 +387,16 @@ However, it usually contains a Protobuf equivalent of the filter's static config
...
-[runner]: #
-[create-filter-args-config]: #CreateFilter::config
-[config-type-dynamic]: #ConfigType::Dynamic
+[Filter]: ../../../api/quilkin/filters/trait.Filter.html
+[FilterFactory]: ../../../api/quilkin/filters/trait.FilterFactory.html
+[filter-factory-name]: ../../../api/quilkin/0.1.0/quilkin/filters/trait.FilterFactory.html#tymethod.name
+[FilterRegistry]: ../../../api/quilkin/0.1.0/quilkin/filters/struct.FilterRegistry.html
+[runner::run]: ../../../api/quilkin/0.1.0/quilkin/runner/fn.run.html
+[CreateFilterArgs::config]: ../../../api/quilkin/0.1.0/quilkin/filters/prelude/struct.CreateFilterArgs.html#structfield.config
+[ConfigType::dynamic]: ../../../api/quilkin/0.1.0/quilkin/filters/enum.ConfigType.html#variant.Dynamic
```
markmandel, line 396:
```suggestion
[ConfigType::dynamic]: ../../../api/quilkin/filters/enum.ConfigType.html#variant.Dynamic
```

(Also this type doesn't exist anymore, but that may be a problem for another day).

docs/src/filters/writing_custom_filters.md
```diff
@@ -392,17 +387,16 @@ However, it usually contains a Protobuf equivalent of the filter's static config
...
 
 [anchor-static-config]: #static-configuration
-[Filters]: ./filters.md
+[Filters]: ../filters.md
```
markmandel, line 399: This change to `..` needs to be applied to all the below links that also point to `filters.md`.

docs/src/filters/writing_custom_filters.md
```diff
@@ -392,17 +387,16 @@ However, it usually contains a Protobuf equivalent of the filter's static config
...
-[FilterFactory]: #
-[filter-factory-name]: #FilterFactory::name
-[FilterRegistry]: #
-[FilterChain]: #
-[runner]: #
-[create-filter-args-config]: #CreateFilter::config
-[config-type-dynamic]: #ConfigType::Dynamic
+[Filter]: ../../../api/quilkin/filters/trait.Filter.html
+[FilterFactory]: ../../../api/quilkin/filters/trait.FilterFactory.html
+[filter-factory-name]: ../../../api/quilkin/0.1.0/quilkin/filters/trait.FilterFactory.html#tymethod.name
```
markmandel, line 392:
```suggestion
[filter-factory-name]: ../../../api/quilkin/filters/trait.FilterFactory.html#tymethod.name
```

Don' t need to specify the version, it all gets bundled up under the the same parent directory, which has the version snapshot.

docs/src/xds.md
```diff
@@ -70,13 +70,13 @@ Quilkin exposes the following metrics around the management servers and its reso
...
-[filters-doc]: ./extensions/filters/filters.md
+[filters-doc]: ./filters.md
 [listener-resource]: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener.proto#config-listener-v3-listener
 [xds-filters]: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener_components.proto#envoy-v3-api-msg-config-listener-v3-filter
 [xds-filter-chain]: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/listener/v3/listener_components.proto#config-listener-v3-filterchain
 [DiscoveryRequest]: https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/discovery.proto#envoy-api-msg-discoveryrequest
 [xds-variants]: https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#variants-of-the-xds-transport-protocol
-[filter-protos]: ../proto/quilkin/extensions/filters
-[filters-doc]: ./extensions/filters/filters.md
+[filter-protos]: https://github.com/googleforgames/quilkin/tree/v0.1.0/proto/quilkin/extensions/filters
```
markmandel, line 79: 🤔  should we point at main, rather than the version? We could add a note in the docs that you would want to choose the release tagged version when importing?

Save us updating the docs on each release?

markmandel (commented):
Pulled this down and ran a local preview, found a few broken links (Hence #367)

I was hacking away to work out what was needed to be fixed, so I added suggestions below, but I attached a patch if you want to have all my changes.
[docs.patch.txt](https://github.com/googleforgames/quilkin/files/6984786/docs.patch.txt)

(Can also grab it on this [branch](https://github.com/markmandel/quilkin/tree/iu/doc-links))

docs/src/filters/writing_custom_filters.md
iffyio: Cool!

markmandel (approved):

PR #364, opened 2021-08-12
Add a type for session keys

iffyio:
Currently looking at adding some sort of session based state in a filter
and figured to make this a type rather than using a plain tuple of adresses so that
its easier to identify across the repo


XAMPPRocky (approved):

iffyio:
actually scratch that, my use case (ratelimiting) doesn't need per session state since filters don't know what the destination address will be :P only the source will be sufficient, this is probably still a useful change though

markmandel (approved):
No objections here - I think it makes the code more readable 👍🏻 if you are still happy with the change, let's merge.

PR #369, opened 2021-08-17
Move to run and test subcommand setup

XAMPPRocky:
More work towards #318 I wanted to separate this change out to catch anything that breaks from this part. This changes it so `quilkin`'s default mode prints the help text, and now to run the main process you call `quilkin run`. This setup is easier for us, as right now there's not really a good way to run a subcommand as the default command in `clap` if nothing else is present AFAICT.


iffyio (approved):

markmandel:
Just updating the CI, so I expect this will fail on the Docker image test (entrypoint will need an update).

Weird that the last notification was that it failed testing (which it looks like it did with a fail on a `cargo fmt` check) - but Github was reporting that it was fine. Not sure what that's about 🤔

XAMPPRocky:
@markmandel I believe everything is working now.

PR #371, opened 2021-08-19
Add initial version of the test subcommand

### LINKED ISSUE #318: Create More Complete End-to-End Testing Framework

XAMPPRocky:
This PR adds an initial version of the `test` subcommand discussed in #318. Adding a `Testsuite` type which represents a given config and a set of tests to be run against that. In addition, the server now accepts a oneshot sender that is used to indicate when the server is ready to start receiving packets, this removes the need to sleep or wait.

Closes #318 as while this currently this only does basic string equality testing, we can add to it and replace tests incrementally, and I think will make more sense to split things up into feature specific issues.

### DIFF
--- a/src/config/test.rs
+++ b/src/config/test.rs
@@ -1,0 +1,184 @@
+use std::{
+    fmt,
+    path::{Path, PathBuf},
+};
+
+use prometheus::proto::MetricType;
+use serde::{
+    de::{Error, MapAccess, Visitor},
+    Deserialize, Deserializer, Serialize,
+};
+
+use super::{error::TestSuiteDecodeError, Config};
+
+/// The configuration of a Quilkin testsuite.
+pub struct TestSuite {
+    pub config: Config,
+    pub options: TestConfig,
+}
+
+#[derive(Deserialize)]
+pub struct TestConfig {
+    pub config: Option<PathBuf>,
+    pub tests: std::collections::HashMap<String, TestCase>,
+}
+
+impl TestSuite {
+    pub fn find<P: AsRef<Path>>(
+        log: &slog::Logger,
+        path: Option<P>,
+    ) -> Result<Self, TestSuiteDecodeError> {
+        super::find_config_file(log, path)
+            .map_err(From::from)
+            .and_then(|s| Self::from_yaml(&s))
+    }
+
+    /// Attempts to deserialize [`Self`] from a YAML document. A valid source is
+    /// either a combination of [`Config`] document followed by [`TestConfig`]
+    /// document separated by a `---` (YAML document separator), or a
+    /// `TestConfig` document containing a `config` key that points to a valid
+    /// `Config` file.
+    pub fn from_yaml(src: &str) -> Result<Self, TestSuiteDecodeError> {
+        Ok(
+            if let Ok(options) = serde_yaml::from_str::<TestConfig>(src) {
+                let path = options
+                    .config
+                    .as_deref()
+                    .ok_or(TestSuiteDecodeError::MissingConfigInTestOptions)?;
+                let config = serde_yaml::from_reader(std::fs::File::open(path)?)?;
+                Self { config, options }
+            } else {
+                let mut de = serde_yaml::Deserializer::from_str(src);
+                let config =
+                    Config::deserialize(de.next().ok_or(TestSuiteDecodeError::MissingConfig)?)?;
+                let options = TestConfig::deserialize(
+                    de.next().ok_or(TestSuiteDecodeError::MissingTestOptions)?,
+                )?;
+                Self { config, options }
+            },
+        )
+    }
+}
+
+#[derive(Deserialize)]
+pub struct TestCase {
+    /// The data to be given to Quilkin.
+    pub input: Data,
+    /// What we expect Quilkin to send to the game server.
+    pub output: Data,
+    /// Metrics we expect to be set.
+    pub metrics: Option<std::collections::HashMap<String, MetricComparison>>,
+}
+
+#[derive(Deserialize)]
+pub struct MetricComparison {
+    pub name: String,
+    #[serde(deserialize_with = "metric_type_from_str")]
+    pub r#type: MetricType,
+    pub r#value: f64,
+}
+
+fn metric_type_from_str<'de, D>(deserializer: D) -> Result<MetricType, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    const METRIC_TYPES: &[(&str, MetricType)] = &[
+        ("counter", MetricType::COUNTER),
+        ("gauge", MetricType::GAUGE),
+        ("histogram", MetricType::HISTOGRAM),
+        ("summary", MetricType::SUMMARY),
+        ("untyped", MetricType::UNTYPED),
+    ];
+
+    let input = <&str>::deserialize(deserializer)?.to_lowercase();
+
+    METRIC_TYPES
+        .iter()
+        .find(|(key, _)| *key == input)
+        .map(|(_, value)| *value)
+        .ok_or_else(|| {
+            Error::custom(format!(
+                "Invalid Prometheus metric type. Expected: {}",
+                itertools::Itertools::intersperse(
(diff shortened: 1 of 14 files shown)


src/runner.rs
```diff
@@ -61,19 +68,107 @@ pub async fn run_with_config(
...
+    static FEEDBACK_LOOP: Lazy<SocketAddr> = Lazy::new(|| {
+        let socket = std::net::UdpSocket::bind((Ipv4Addr::UNSPECIFIED, 0)).unwrap();
+        let local_addr = socket.local_addr().unwrap();
+        std::thread::spawn(move || loop {
+            let mut packet = [0; 0xffff];
+            let (_, addr) = socket.recv_from(&mut packet).unwrap();
+            let length = packet
+                .iter()
+                .position(|&x| x == 0)
+                .unwrap_or_else(|| packet.len());
```
iffyio, line 107: we can use the `(length, addr)` returned by `recv_from` as the length rather than calculate it manually?

src/runner.rs
```diff
@@ -61,19 +68,107 @@ pub async fn run_with_config(
...
+            let packet = &packet[..length];
+            eprintln!("Received: {:?}", packet);
```
iffyio, line 109:
```suggestion
```
this seems leftover? or if we wanted to we can add an optional debug logging during tests

src/runner.rs
```diff
@@ -61,19 +68,107 @@ pub async fn run_with_config(
...
+    mut testsuite: Testsuite,
+    filter_factories: impl IntoIterator<Item = DynFilterFactory>,
+) -> Result<(), Error> {
+    use once_cell::sync::Lazy;
+
+    static FEEDBACK_LOOP: Lazy<SocketAddr> = Lazy::new(|| {
+        let socket = std::net::UdpSocket::bind((Ipv4Addr::UNSPECIFIED, 0)).unwrap();
+        let local_addr = socket.local_addr().unwrap();
+        std::thread::spawn(move || loop {
+            let mut packet = [0; 0xffff];
```
iffyio, line 102: rather than create a new 64k buffer each iteration, lets move this outside of the loop

src/runner.rs
```diff
@@ -61,19 +68,107 @@ pub async fn run_with_config(
...
+
+        local_addr
+    });
+
+    let log = base_log.new(o!("source" => "run"));
+    let mut receiver = *FEEDBACK_LOOP;
+    receiver.set_ip(Ipv4Addr::LOCALHOST.into());
+
+    {
+        let endpoints = testsuite.config.source.get_static_endpoints_mut().unwrap();
```
iffyio, line 121: I'd prefer we didn't have any concept of mutability in our config, I think we can copy the testconfig and insert the exact `source` we want to use for test?
Coming to think of it, I think this means that all endpoint info in the config are ignored currently (until we have support to point the test to an endpoint)? if so we should call it out in the docs

XAMPPRocky: > Coming to think of it, I think this means that all endpoint info in the config are ignored currently (until we have support to point the test to an endpoint)? if so we should call it out in the docs

I can add docs about it. On using the real endpoints, that would be simple to add, the question is whether that would be useful. As working on your machine you're not always going to be able to access the same IPs that are in the config if it's not publicly accessible.

XAMPPRocky:
@googlebot I consent.

XAMPPRocky:
@iffyio @markmandel Have a final review of the code and let me know what you think, I still have to add documentation, which I'm going to do in a followup PR since this is already a bit big, and I'd like to start using this to write new tests and slowly move over older tests as needed.

src/runner.rs
iffyio: can we copy the config rather than have the `get_static_endpoints_mut` function?

> can add docs about it. On using the real endpoints, that would be simple to add, the question is whether that would be useful. As working on your machine you're not always going to be able to access the same IPs that are in the config if it's not publicly accessible.

Yeah I think that would be useful otherwise we won't be able to do anything beyond unit tests. In the future we'd want to run test against an actual deployment e before rolling out changes for example where we know its IP

src/config/test.rs
```diff
@@ -0,0 +1,184 @@
...
+                .collect::<String>(),
+            ))
+        })
+}
+
+pub enum Data {
+    String(String),
+    Binary(Vec<u8>),
+    Base64(Vec<u8>),
+}
```
iffyio, line 115: what's this used for?

src/config/test.rs
```diff
@@ -0,0 +1,184 @@
...
+    pub metrics: Option<std::collections::HashMap<String, MetricComparison>>,
+}
+
+#[derive(Deserialize)]
+pub struct MetricComparison {
+    pub name: String,
+    #[serde(deserialize_with = "metric_type_from_str")]
+    pub r#type: MetricType,
+    pub r#value: f64,
+}
```
iffyio, line 79: how come we're having metrics here I'm wondering how is it going to be consumed? and it seems like we're introducing our own metrics collector implelmentation, can we use the prometheus ones we export?

XAMPPRocky: This is for defining comparisons between metrics in the test file. For example if you wanted to test that the capture bytes filter captured five bytes you'd write something like the following. I have to manually deserialize to the type, because they don't provide a `FromStr` implementation for `MetricType`. It's not currently being used so I can remove it.

```yaml
tests:
   it_works:
     input: Hello
     output: Hello
     metrics:
     - name: packets_received_total
       type: gauge
       value: 5
```

src/config/test.rs
XAMPPRocky: This allows you accept strings, lists of numbers, or base64 as for the `input` and `output` properties, so you can write any of the following as valid data input for your tests.

```yaml
# Strings
input: Hello
# Bytes
input: [86, 45, 23, 23, 67]
# Base64
input:
   type: base64
   value: MWfu==

Comment (iffyio) [reply src/config/test.rs:115]: Ah that makes sense!

File: src/proxy/server.rs [hunk 7]
```diff
@@ -99,14 +99,19 @@ struct ProcessDownstreamReceiveConfig {
 impl Server {
     /// start the async processing of incoming UDP packets. Will block until an
     /// event is sent through the stop Receiver.
-    pub async fn run(self, mut shutdown_rx: watch::Receiver<()>) -> Result<()> {
+    pub async fn run(
+        self,
+        ready_tx: oneshot::Sender<SocketAddr>,
```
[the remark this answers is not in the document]
XAMPPRocky: Well in my attempts to remove this, and replace it with polling, I start running into an issue where the server shuts down before the tests can complete, and I can't trace where it is causing it to shutdown. So for now, I'm going to move to try and simplify our task management, so that's there's a much clearer and easier to track how we're shutting down.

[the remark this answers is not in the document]
markmandel: Maybe a silly question - but what happens if you don't have a the `ready_tx` at all? 

Out integration tests in Rust don't seem to need to wait for the ready state, so I'm curious as to why this does?

XAMPPRocky:
I want to get back to this, but I'm going to do so as a new PR, as I'm more focused on the xDS and Quilkin API right now, and the conflicts will continue to pile up.

PR #381, opened 2021-08-27
Hash load balancer

gilesheron:
First cut of a hashed load-balancer for Quilkin so all requests from one session go to the same endpoint.

Note that hashes by source IP/port.   Might be good to make it configurable.

Also doesn't implement a consistent hash (so if the set of endpoints changes all flows will re-hash).

### DIFF
--- a/src/filters/load_balancer/config.rs
+++ b/src/filters/load_balancer/config.rs
@@ -21,7 +21,9 @@ use std::convert::TryFrom;
 use serde::{Deserialize, Serialize};
 
 use self::quilkin::extensions::filters::load_balancer::v1alpha1::load_balancer::Policy as ProtoPolicy;
-use super::endpoint_chooser::{EndpointChooser, RandomEndpointChooser, RoundRobinEndpointChooser};
+use super::endpoint_chooser::{
+    EndpointChooser, HashEndpointChooser, RandomEndpointChooser, RoundRobinEndpointChooser,
+};
 use crate::{filters::ConvertProtoConfigError, map_proto_enum};
 
 pub use self::quilkin::extensions::filters::load_balancer::v1alpha1::LoadBalancer as ProtoConfig;
@@ -46,7 +48,7 @@ impl TryFrom<ProtoConfig> for Config {
                     field = "policy",
                     proto_enum_type = ProtoPolicy,
                     target_enum_type = Policy,
-                    variants = [RoundRobin, Random]
+                    variants = [RoundRobin, Random, Hash]
                 )
             })
             .transpose()?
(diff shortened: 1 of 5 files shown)


XAMPPRocky (approved):
Thank you for your PR! This looks good to me, pretty nice solution. You mention it would be nice to add configuration, what would you want to configure about the hashing?

gilesheron:
> Thank you for your PR! This looks good to me, pretty nice solution. You mention it would be nice to add configuration, what would you want to configure about the hashing?

might be worth having an option to hash by source IP only if you want a mode where all sessions from a given user are routed to the same endpoint?

will re-submit as a single commit with the correct email address - once I work out how to do that!

XAMPPRocky:
> might be worth having an option to hash by source IP only if you want a mode where all sessions from a given user are routed to the same endpoint?

I might be missing something, but isn't that what setting `HASH` would do? Or are you suggesting having another option that remembers (ip, port) but doesn't hash?

gilesheron:
> > might be worth having an option to hash by source IP only if you want a mode where all sessions from a given user are routed to the same endpoint?
> 
> I might be missing something, but isn't that what setting `HASH` would do? Or are you suggesting having another option that remembers (ip, port) but doesn't hash?

ah sorry.  Didn't explain myself well.

the current solution hashes on the source SocketAddr.  So all traffic from the same IP/port will go to the same endpoint.  My guess is that for most games that will ensure that all of a given user's traffic will go to the same endpoint, whilst also giving optimal hashing if e.g. multiple users are behind a NAT, and ensuring that a user can reconnect and get a different endpoint if one endpoint has an issue.

the other approach would be to hash on source IpAddr.   that way all traffic from given IP will go to one endpoint.   So if a game has multiple flows per user that will ensure that they land on the same endpoint.

src/filters/load_balancer/config.rs
```diff
@@ -65,13 +65,17 @@ pub enum Policy {
     /// Send packets to endpoints chosen at random.
     #[serde(rename = "RANDOM")]
     Random,
+    /// Send packets to endpoints based on hash of source IP
```
iffyio, line 68:
```suggestion
    /// Send packets to endpoints based on hash of source IP and Port
```

gilesheron: makes sense

iffyio (approved):
LGTM! Thanks!

markmandel:
We were tracking this on Discord, but @thisisnotapril is looking into what the situation is with the CLA.

markmandel:
Manual approval on the CLA - so merging 👍🏻

PR #385, opened 2021-08-29
replace slog with tracing in Filter

rezvaneh:
This PR is an initiator to replace slog with tracing.
As this PR is already big, it aims at replacing only in `src/filters/…` first. 

#317


src/filters/token_router.rs
```diff
@@ -83,24 +81,19 @@ impl FilterFactory for TokenRouterFactory {
...
+    #[instrument]
     fn read(&self, mut ctx: ReadContext) -> Option<ReadResponse> {
         match ctx.metadata.get(self.metadata_key.as_ref()) {
             None => {
                 if self.metrics.packets_dropped_no_token_found.get() % LOG_SAMPLING_RATE == 0 {
-                    error!(
-                        self.log,
-                        "Packets are being dropped as no routing token was found in filter dynamic metadata";
-                        "count" => self.metrics.packets_dropped_no_token_found.get(),
-                        "metadata_key" => self.metadata_key.clone()
```
XAMPPRocky, line 102: Why were these variables removed?

src/filters/token_router.rs
```diff
@@ -118,12 +111,7 @@ impl Filter for TokenRouter {
                 },
                 None => {
                     if self.metrics.packets_dropped_invalid_token.get() % LOG_SAMPLING_RATE == 0 {
-                        error!(
-                            self.log,
-                            "Packets are being dropped as routing token has invalid type: expected Vec<u8>";
-                            "count" => self.metrics.packets_dropped_invalid_token.get(),
-                            "metadata_key" => self.metadata_key.clone()
```
XAMPPRocky, line 125: Same as above.

XAMPPRocky (commented):
Thank you for your PR! This looks good to me, I've left two comments and will checkout and run this locally tomorrow, and then we can merge this.

src/filters/debug.rs
```diff
@@ -16,51 +16,43 @@
...
-        };
 
-        Debug { log }
+    fn new(_: Option<String>) -> Self {
+        Debug {}
     }
 }
 
 impl Filter for Debug {
+    #[instrument]
```
rezvaneh, line 47: I would say better not to use this instrument attribute here!

it creates and enters a span every time the function is called and the parameters passed to the function are recorded as fields in the logs automatically.

if the number of parameters is high, our logs won’t be easily readable anymore!
with the result of this code here, we would have a log similar to:

`INFO caller_func_name:read{self=Debug ctx=ReadContext { endpoints: UpstreamEndpoints { endpoints: Endpoints([Endpoint { address: 127.0.0.1:80, metadata: MetadataView { known: Metadata { tokens: {} }, unknown: Mapping { map: {} } } }]), subset: None }, from: 127.0.0.1:90, contents: [104, 101, 108, 108, 111], metadata: {} }}: quilkin::filters::debug: Read filter event contents="hello"`

but if we try to just enter a span [like here](https://github.com/googleforgames/quilkin/pull/385/files#diff-fd5ade0f959fc048f50224e1dd2c16f8081fefbc5dde1f8f76f81b4909966805R63), the result can be:
`quilkin::filters::debug: Read filter event contents="hello"`

(as a json though)

XAMPPRocky: Makes sense, I know you can configure `instrument` to skip some stuff, so that might be worth looking into, but it's fine if you just enter a span manually.

rezvaneh: was thinking about having the same naming for our logs with that manual span, which that also is configurable in instrument like `#[instrument(name = “extensions::Compress::read”)]`

after conclusion on this, I’ll change other funcs to have them alike!

src/filters/token_router.rs
rezvaneh: I'll get them back after creating a more custom span.

iffyio (changes requested):
Please hold off on merging this :pray: I'm super curious about what the `instrument` and other attributes do especially perf-wise and would love to review the PR first. Will look into it tomorrow hopefully!

Cargo.toml
```diff
@@ -66,6 +66,10 @@ snap = "1.0.3"
 tokio = { version = "1.8.2", features = ["rt-multi-thread", "signal", "test-util", "parking_lot"] }
 tokio-stream = "0.1.2"
 tonic = "0.5.0"
+tracing = {version = "0.1"}
+tracing-subscriber = {version = "0.2"}
+tracing-futures = { version = "0.2.5", default-features = false }
+tracing-test = "0.1"
```
iffyio, line 72: Since this is a test dependency it should be added under `dev-dependencies` instead

Cargo.toml
```diff
@@ -66,6 +66,10 @@ snap = "1.0.3"
 tokio = { version = "1.8.2", features = ["rt-multi-thread", "signal", "test-util", "parking_lot"] }
 tokio-stream = "0.1.2"
 tonic = "0.5.0"
+tracing = {version = "0.1"}
+tracing-subscriber = {version = "0.2"}
+tracing-futures = { version = "0.2.5", default-features = false }
```
iffyio, line 71:
```suggestion
```
this seems unused and we can remove it?

iffyio (changes requested):
(Sorry about the delay in reviewing) had a look at the PR now and this seems to be an expensive library, did a quick load test with 5 noop filters. With `#[instrument(skip(self, ctx))]` set, perf was almost halved.
Then I tried with a trivial span enter and drop on exit and still saw significant perf hits though not as bad with `instrument`.
```rust
impl Filter for Noop {
    fn read(&self, ctx: ReadContext) -> Option<ReadResponse> {
        let span = tracing::span!(tracing::Level::TRACE, "myspan");
        let _enter = span.enter();
        Some(ctx.into())
    }
}
```
So my take on the change is that we should not add this to filters on each read/write (I think similar to how we don't want to log on each read/write) (or we make sure they're compiled out in release mode or something).

XAMPPRocky:
> Tracing shouldn't be on by default

Just to clarify. Tracing instrumentation shouldn't be on by default. E.g. `#[instrument]`, calling `trace`, `debug`, etc. doesn't need a feature flag.

iffyio (approved):

markmandel (approved):

PR #388, opened 2021-08-31
Add github CODEOWNERS file

iffyio:
So that we get pinged when PRs are up for review

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @iffyio @markmandel @xampprocky


XAMPPRocky (approved):

markmandel (approved):
Excellent suggestion!

thisisnotapril:
Did y'all also do the config in the repo to get those auto-assigned?

markmandel:
> Did y'all also do the config in the repo to get those auto-assigned?

I don't think so, since either you or me would need to do it, and I didn't do it.

Issue #373, opened 2021-08-25
Review Docs: Writing Custom Filters for API changes


markmandel:
When I last looked at https://googleforgames.github.io/quilkin/main/book/filters/writing_custom_filters.html

I noticed that there were a few broken links (#367 should totally be a thing), (it looks like we're missing going one level too deep on some of the links).

The main part I saw was that we've improved the experience on "[Working with Filter Configuration](https://googleforgames.github.io/quilkin/main/book/filters/writing_custom_filters.html#working-with-filter-configuration)", and I'm not sure if the current documentation matches what we have via our current API surface.

iffyio:
> The main part I saw was that we've improved the experience on "Working with Filter Configuration", and I'm not sure if the current documentation matches what we have via our current API surface.

Is there anything in particular that's missing, from what I can tell it looks up to date, not sure of any related filter api changes that would need updating?

markmandel:
Just doing a quick check through as I fix up some 404s... so I can be more clear!

Looks like the current approach will work, but the new https://googleforgames.github.io/quilkin/main/api/quilkin/config/enum.ConfigType.html#method.deserialize method definitely makes things easier, and we might want to push people in that direction.

That being said, the current approach works, so not sure if we want to block on this for the release? How do you feel?

markmandel:
To add an extra example, here is what the `LoadBalancerFactory` looks like now

```rust
impl FilterFactory for LoadBalancerFilterFactory {
    fn name(&self) -> &'static str {
        NAME
    }

    fn create_filter(&self, args: CreateFilterArgs) -> Result<Box<dyn Filter>, Error> {
        let config: Config = self
            .require_config(args.config)?
            .deserialize::<Config, ProtoConfig>(self.name())?;

        Ok(Box::new(LoadBalancer {
            endpoint_chooser: config.policy.as_endpoint_chooser(),
        }))
    }
}
```

That `deserialize` is sooo nice 😄

markmandel:
Started updating the example with the new API. Will PR when ready. Haven't touched Rust in a while! 😄

PR #389, opened 2021-09-02
Update custom filter example and add CI

markmandel:
Update the custom filter example to the new API surface, and also setup CI so that it ensures that it always compiles against dev Quilkin, and therefore will stay in sync going forward.

Next I want to block out pieces of code and inject them into our docs via mdbook when updating https://googleforgames.github.io/quilkin/main/book/filters/writing_custom_filters.html as well, rather than have them copy pasted.

Work on #373

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -47,7 +47,10 @@ version:
 	@echo $(package_version)
 
 # Run all tests
-test: ensure-build-image
+test: ensure-build-image test-quilkin test-examples
+
+# test only the quilkin crate
+test-quilkin: ensure-build-image
 	docker run --rm $(common_rust_args) \
      		--entrypoint=cargo $(BUILD_IMAGE_TAG) deny check
 	docker run --rm $(common_rust_args) \
@@ -56,6 +59,13 @@ test: ensure-build-image
  		--entrypoint=cargo $(BUILD_IMAGE_TAG) fmt -- --check
 	docker run --rm $(common_rust_args) \
      		--entrypoint=cargo $(BUILD_IMAGE_TAG) test
+
+# Run tests against the examples
+test-examples: ensure-build-image
+	docker run --rm $(common_rust_args) -w /workspace/examples/quilkin-filter-example \
+		--entrypoint=cargo $(BUILD_IMAGE_TAG) clippy --tests -- -D warnings
+	docker run --rm $(common_rust_args) -w /workspace/examples/quilkin-filter-example \
+		--entrypoint=cargo $(BUILD_IMAGE_TAG) fmt -- --check
 
 # Build all binaries, images and related artifacts
 build: binary-archive build-image
--- a/examples/quilkin-filter-example/Cargo.toml
+++ b/examples/quilkin-filter-example/Cargo.toml
@@ -28,6 +28,7 @@ edition = "2018"
 # quilkin = "0.2.0"
 quilkin = { path = "../../" }
 tokio = { version = "1", features = ["full"] }
+tonic = "0.5.0"
 prost = "0.8"
 prost-types = "0.8"
 serde = "1.0"
--- a/examples/quilkin-filter-example/Cargo.toml
+++ b/examples/quilkin-filter-example/Cargo.toml
@@ -14,6 +14,8 @@
 # limitations under the License.
 #
 
+[workspace]
+
 [package]
 name = "quilkin-filter-example"
 version = "0.1.0"
@@ -22,13 +24,15 @@ repository = "https://github.com/googleforgames/quilkin"
 edition = "2018"
 
 [dependencies]
-quilkin = "0.1.0"
-tokio = { version = "1", features = ["full"]}
-prost = "0.7"
-prost-types = "0.7"
+# If lifting this example, you will want to be explicit about the Quilkin version, e.g.
+# quilkin = "0.2.0"
+quilkin = { path = "../../" }
+tokio = { version = "1", features = ["full"] }
+prost = "0.8"
+prost-types = "0.8"
 serde = "1.0"
 serde_yaml = "0.8"
 bytes = "1.0"
 
 [build-dependencies]
-prost-build = "0.7"
+prost-build = "0.8"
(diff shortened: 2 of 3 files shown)


build/Makefile
```diff
@@ -47,7 +47,10 @@ version:
 	@echo $(package_version)
 
 # Run all tests
-test: ensure-build-image
+test: ensure-build-image test-quilkin test-examples
```
markmandel, line 50: Split up quilkin tests from examples tests here into separate targets, in case someone only wants to test one at a time.

examples/quilkin-filter-example/Cargo.toml
```diff
@@ -22,13 +22,17 @@ repository = "https://github.com/googleforgames/quilkin"
...
+prost-types = "0.8"
 serde = "1.0"
 serde_yaml = "0.8"
 bytes = "1.0"
 
 [build-dependencies]
-prost-build = "0.7"
+prost-build = "0.8"
+
+[workspace]
```
markmandel, line 38: Had to add this `[workspace]` field here, otherwise I would get the error:

`Error during execution of `cargo metadata`: error: current package believes it's in a workspace when it's not:` -- I'm not sure I really get cargo workspaces tbh, so if you have any feedback here, I would appreciate it.

examples/quilkin-filter-example/Cargo.toml
```diff
@@ -22,13 +22,17 @@ repository = "https://github.com/googleforgames/quilkin"
 edition = "2018"
 
 [dependencies]
-quilkin = "0.1.0"
-tokio = { version = "1", features = ["full"]}
-prost = "0.7"
-prost-types = "0.7"
+# If lifting this example, you will want to be explicit abou the Quilkin version, e.g.
+# quilkin = "0.2.0"
+quilkin = { path = "../../" }
```
markmandel, line 27: This was the best way I could think of to ensure that the code stayed in sync. Is there a better way?

examples/quilkin-filter-example/src/main.rs
```diff
@@ -14,16 +14,27 @@
...
+impl TryFrom<ProtoGreet> for Config {
+    type Error = ConvertProtoConfigError;
+
+    fn try_from(p: ProtoGreet) -> Result<Self, Self::Error> {
+        Ok(Config {
+            greeting: p.greeting,
+        })
+    }
+}
+
```
markmandel, line 37: Should/could we using `include_proto!` here instead? I tried it, and it wanted me to have tonic as a dependency, so I'm assuming that's why we avoided it in the example?

examples/quilkin-filter-example/Cargo.toml
XAMPPRocky: I think this is reasonable. This code is never being published to crates.io, so using paths is totally fine.

examples/quilkin-filter-example/Cargo.toml
XAMPPRocky: All cargo projects (even ones with one crate) are workspaces, starting from the `Cargo.toml` and recursively throughout the directory. So if you have a project with multiple crates, cargo needs to figure out if it's part of the parent workspace, or if it's in a new workspace (this will determine what crates to build, where the target directory is, etc). One way to do that is to add `[workspace]` in the `Cargo.toml`, which is explicitly saying "This is the start of a new workspace". The other option is using the `members` and `exclude` fields in the parent `[workspace]`. I would probably go with either of these.

- `members = ["examples/quilkin-filter-example", ...]`, add it to the `members` field, which means it will be built as part of `cargo build --workspace`.
- `exclude = ["examples/quilkin-filter-example", ...]`, add it to the `exclude` field, which means it will **not** be built as part of `cargo build --workspace`, and you need to build it explicitly yourself.

examples/quilkin-filter-example/src/main.rs
XAMPPRocky: I've no issue with using `tonic`, if you depend on `quilkin`, you already depend on `tonic` indirectly.

XAMPPRocky (approved):

examples/quilkin-filter-example/Cargo.toml
markmandel: Thanks for the detailed explanation - that helped a lot!

I'm definitely leaning towards having it be seperate, and that way if you are developing against quilkin you don't _have_ to process the example of every build and test.

I just tried using `exclude` in `quilkin/Cargo.toml`, and I'm still getting the error:

```
Error during execution of `cargo metadata`: error: current package believes it's in a workspace when it's not:
current:   /home/markmandel/workspace/quilkin/examples/quilkin-filter-example/Cargo.toml
workspace: /home/markmandel/workspace/quilkin/Cargo.toml

this may be fixable by adding `examples/quilkin-filter-example` to the `workspace.members` array of the manifest located at: /home/markmandel/workspace/quilkin/Cargo.toml
Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
```

Seems like leaving `[workspace]` in the example `Cargo.toml` is the only way.

examples/quilkin-filter-example/src/main.rs
markmandel: Agreed! Got it to work! Update coming soon.

examples/quilkin-filter-example/src/main.rs
```diff
@@ -55,22 +63,16 @@ impl FilterFactory for GreetFilterFactory {
...
-            ConfigType::Dynamic(config) => {
-                let config: greet::Greet = prost::Message::decode(Bytes::from(config.value)).unwrap();
-                config.greeting
-            }
-        };
-        Ok(Box::new(Greet(greeting)))
+        let greeting = self
+            .require_config(args.config)?
+            .deserialize::<Config, ProtoGreet>(self.name())?;
+        Ok(Box::new(Greet(greeting.greeting)))
```
iffyio, line 69: this is now out of sync with the doc, we'll need to update that to match

markmandel: That'll be my next job! I'll take a stab at pulling the code samples from this code directly as well via mdbook. 👍🏻

Issue #266, opened 2021-05-19
[bot] Hide Old Comment On Push


XAMPPRocky:
Also would be nice if the GitHub bot hid its previous comment when you push to a PR.

_Originally posted by @XAMPPRocky in https://github.com/googleforgames/quilkin/issues/264#issuecomment-844248568_

markmandel:
Since it came up on Discord, here is where the bot code is:

https://github.com/googleforgames/quilkin/tree/main/build/ci/github-bot

markmandel:
So I'm pretty happy with how the bot is working with deleting old statuses! I've yet to want go back to a previous build.

So unless anyone objects, I'm happy to close this ticket!

PR #390, opened 2021-09-03
Support clear bot history on new build

devharis:
As per [this](https://github.com/googleforgames/quilkin/issues/266) issue - impl. support to clear bot history on new builds to prevent PR comment bloat.


devharis:
I saw your discussion regarding hiding old comments, so you really need to preserve build history comments? If not - this would support deleting them.

markmandel:
I didn't think to just delete the old comments on each run (rather than minimise), but it's not a bad idea to at least test.

I do like being able to see the history over time (so prefer minimisation), it can also make it easier to identify flaky tests, and I have some vague concerns about race conditions when builds are simultaneous, but I don't see any reason not to at least test it out.

I wonder if we'll hit API request limits in the beginning 😄 

@XAMPPRocky did you have any thoughts? It was you who brought up removing/hiding the old test results, so wanted to check in first before deploying anything new.

XAMPPRocky:
It seems fine to me. My one concern about deleting, is that it would make it harder to look at old builds, which sometimes you want to do, but I'm not against deleting either.

markmandel:
> It seems fine to me. My one concern about deleting, is that it would make it harder to look at old builds, which sometimes you want to do, but I'm not against deleting either.

Yeah, same here. But let's try it! If we hate it, we can change it again. I'll get it built and deployed, and will confirm it works on this PR, and assuming it all works, we can roll it out and see how we go!

markmandel:
Haven't forgotten at this, just been at 100% bandwidth. Do want to get this live so we can take it for a spin, but would like to do it at a time so I can watch in case it goes badly for any reason 😄

build/ci/github-bot/main.go
```diff
@@ -129,6 +132,22 @@ func (g *githubNotifier) SendNotification(ctx context.Context, build *cloudbuild
 	return nil
 }
 
+func (g *githubNotifier) clearBotComments(ctx context.Context, pr int) error {
+	comments, _, err := g.client.Issues.ListComments(ctx, owner, repo, pr, &github.IssueListCommentsOptions{})
+	if err != nil {
+		log.Errorf("Error retriving comment history: %v on PR with id: %s", err, pr)
```
markmandel, line 138: Small spelling error: 
```suggestion
		log.Errorf("Error retrieving comment history: %v on PR with id: %s", err, pr)
```

markmandel:
Deploying now! Let's see how we go!

markmandel:
Looks like it worked! Let's let it bake for a few days (also would love to get that typo fixed), and if it's going swimmingly, let's merge 👍🏻

markmandel (approved):

markmandel:
Let's merge it - it's looking good 👍🏻

PR #391, opened 2021-09-06
Use hostname as default proxy id

iffyio:
If we don't provide a default proxy id this falls back to using the
hostname instead of uuid (and if that fails, then we use a uuid).

For environments like kubernetes, we would like the proxy's id to match
the pod id so that the management server can easily find pod information
for the proxy that is connected to it (the proxy passes its ID when it
connects to the server)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,7 @@ slog-async = "2.6.0"
 slog-json = "2.3.0"
 slog-term = "2.5.0"
 snap = "1.0.3"
+sys-info = "0.9.0"
 tokio = { version = "1.8.2", features = ["rt-multi-thread", "signal", "test-util", "parking_lot"] }
 tokio-stream = "0.1.2"
 tonic = "0.5.0"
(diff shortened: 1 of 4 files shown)


XAMPPRocky (approved):

iffyio:
It seems this didn't compile for mac on our ci so I've updated the pr to only use host name on linux all platforms will use uuids

markmandel (approved):

XAMPPRocky:
We should probably file an issue with `sysinfo` if hostname is not working on macOS, as it's supposed to work for all main platforms.

iffyio:
> We should probably file an issue with `sysinfo` if hostname is not working on macOS, as it's supposed to work for all main platforms.

yeah I'm not sure if its rather the lib or what we're doing in CI, when I googled yesterday the issue seemed to be something about not using xcode's gcc to compile so maybe its our cross compilation setup that doesn't support it.
```
warning: cc: error: x86_64: No such file or directory
warning: cc: error: unrecognized command line option '-arch'; did you mean '-march='?

error: failed to run custom build command for `sys-info v0.9.0`
```

markmandel:
🤔 or maybe an issue with https://github.com/joseluisq/rust-linux-darwin-builder ?

There are a set of environment variables that can also be used to specify which compiler to use.

PR #393, opened 2021-09-07
Exit on SIGTERM

iffyio:
we currently only listen for SIGINT while k8s sends
a SIGTERM to shutdown a process, so we need to listen for that
too

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -61,11 +61,26 @@ pub async fn run_with_config(
         .validate()?
         .build();
 
+    #[cfg(target_os = "linux")]
+    let mut sig_term_fut = signal::unix::signal(signal::unix::SignalKind::terminate())?;
+    #[cfg(not(target_os = "linux"))]
+    let sig_term = std::future::pending();
+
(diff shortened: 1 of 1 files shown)


src/runner.rs
```diff
@@ -61,11 +61,26 @@ pub async fn run_with_config(
...
+    let sig_term = std::future::pending();
+
     let (shutdown_tx, shutdown_rx) = watch::channel::<()>(());
+    let signal_log = log.clone();
     tokio::spawn(async move {
+        #[cfg(target_os = "linux")]
+        let sig_term = sig_term_fut.recv();
+        tokio::select! {
+            _ = signal::ctrl_c() => {
+                info!(signal_log, "Received SIGINT")
```
markmandel, line 76: Mega nit: Should this be `debug!`? Not sure if production really cares which signal it gets?

Otherwise, this is a great catch 👍🏻

XAMPPRocky (approved):

markmandel (approved):

markmandel:
(Totally random - but why is the notification bot not working for this PR?)

PR #395, opened 2021-09-14
Docs: Updated Custom Filters

### LINKED ISSUE #373: Review Docs: Writing Custom Filters for API changes

markmandel:
Updates the custom filter documentation to the latest API changes.

This also includes using mdbook's `include` preprocessor to inject the example code into the documentation.

Note: Because of the use of `include` preprocessor I had to remove the Markdown lists, as they wouldn't format correctly
(See: https://github.com/rust-lang/mdBook/issues/1564).

Closes #373

### DIFF
--- a/docs/src/filters/writing_custom_filters.md
+++ b/docs/src/filters/writing_custom_filters.md
@@ -3,11 +3,11 @@
 Quilkin provides an extensible implementation of [Filters] that allows us to plug in custom implementations to fit our needs.
 This document provides an overview of the API and how we can go about writing our own [Filters].
 
-#### API Components
+## API Components
 
 The following components make up Quilkin's implementation of filters.
 
-##### Filter
+### Filter
 
 A [trait][Filter] representing an actual [Filter][built-in-filters] instance in the pipeline.
 
@@ -15,15 +15,15 @@ A [trait][Filter] representing an actual [Filter][built-in-filters] instance in 
 - Both methods are invoked by the proxy when it consults the [filter chain] - their arguments contain information about the packet being processed.
 - `read` is invoked when a packet is received on the local downstream port and is to be sent to an upstream endpoint while `write` is invoked in the opposite direction when a packet is received from an upstream endpoint and is to be sent to a downstream client.
 
-##### FilterFactory
+### FilterFactory
 
 A [trait][FilterFactory] representing a type that knows how to create instances of a particular type of [Filter].
 
 - An implementation provides a `name` and `create_filter` method.
 - `create_filter` takes in [configuration][filter configuration] for the filter to create and returns a new instance of its filter type.
-`name` returns the Filter name - a unique identifier of filters of the created type (e.g quilkin.extensions.filters.debug.v1alpha1.Debug).
+`name` returns the Filter name - a unique identifier of filters of the created type (e.g. quilkin.extensions.filters.debug.v1alpha1.Debug).
 
-##### FilterRegistry
+### FilterRegistry
 
 A [struct][FilterRegistry] representing the set of all filter types known to the proxy.
 It contains all known implementations of [FilterFactory], each identified by their [name][filter-factory-name].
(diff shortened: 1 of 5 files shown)


docs/src/filters/writing_custom_filters.md
```diff
@@ -3,27 +3,27 @@
 Quilkin provides an extensible implementation of [Filters] that allows us to plug in custom implementations to fit our needs.
 This document provides an overview of the API and how we can go about writing our own [Filters].
 
-#### API Components
+## API Components
```
markmandel, line 6: Just noting: Since I had to remove the lists, I moved the headers to be in level order. PTAL, let me know what you think.

docs/src/filters/writing_custom_filters.md
```diff
@@ -38,109 +38,91 @@ These components come together to form the [filter chain].
...
+#### 2. Implement the filter traits
+
+It's not terribly important what the filter in this example does so let's write a `Greet` filter that appends `Hello` to every packet in one direction and `Goodbye` to packets in the opposite direction.
+
+We start with the [Filter] implementation
+
+```rust,no_run,noplayground
+struct Greet;
+
+impl Filter for Greet {
```
markmandel, line 73: I couldn't include this, so just copied it inline. In theory it could go out of date, but I couldn't think of a better way.

markmandel:
Looks like I broke a stack of the docs tests. I'll swing back to this tomorrow and see if I can fix them / if they should be ignored.

docs/src/filters/writing_custom_filters.md
```diff
@@ -38,109 +38,91 @@ These components come together to form the [filter chain].
...
+        NAME
+    }
+    fn create_filter(&self, args: CreateFilterArgs) -> Result<Box<dyn Filter>, Error> {
+        Ok(Box::new(Greet))
+    }
+}
+```
+
+> As a convention we implement a `pub factory() -> DynFilterFactory` function, to 
+> make the next integration step easier.
```
iffyio, line 106: I don't think this is correct? the `factory()` function or constructor wouldn't be visible at all from user code and what's passed into quilkin is the `DynFilterFactory` so it doesn't matter how the object is generated?
```suggestion
```

markmandel: Just so I'm clear, are you saying the description isn't correct? 

If I look at all our Filter implementations, they all have a `pub factory() -> DynFilterFactory` - so I thought it would be good to call out that we conventionally do that, and why.

You are 100% correct that the end use could create a constructor any way, but I thought it would be good to show our usual convention. Maybe the description should be:

> As a convention we generally implement a `pub factory() -> DynFilterFactory` function as a constructor, but a Factory 
> can be constructed as you see fit.

To explain that while you may see this common pattern, it's more by convention, rather than technical limitation. WDYT?

markmandel:
Updated, so now it should pass -- basically put the include sections back the way they were before, and updated the include sections so that they have a `// src/main.rs` (or similar) comment like they did previously.

Also updated the initial examples include to use `use quilkin::filters::prelude::*;` - since that is also what we're generally using now as well.

docs/src/filters/writing_custom_filters.md
markmandel: This should no longer apply, as now it matches how it was before.

docs/src/filters/writing_custom_filters.md
iffyio: Ah, the incorrect part was mostly the 'to make the next integration step easier' bit since the function itself didn't matter.

I think in this case the fact that the quilkin codebase used a function called factory vs anything else doesn't matter at all so it can add confusion to call it out in the first place (as a user I'd find myself reading into what that means and quickly realise it doesn't matter which is wasted time), the user would declare and call the function within their own crate so the function can be called anything or they could construct the factory by some other means (I don't see it as a convention here since both crates don't need to worry about it in the first place).

iffyio (approved):

docs/src/filters/writing_custom_filters.md
markmandel: That's fair. Sounds good, I'll remove the comment 👍🏻 

Worst case, if people start asking, we can add it back in 😄

PR #396, opened 2021-09-14
Add a /config endpoint

### LINKED ISSUE #394: Add admin endpoints to introspect proxy config
It would be very useful for operations to be able to be able to view the current config that a proxy is running at a given point in time.

Let's add two admin endpoints /clusters and /filterchain that returns a simplified JSON of the current cluster and filterchain respectively back to the caller. e.g:

```
$ curl quilkin:8080/clusters
[{
  "name": "default-cluster",
  "endpoints": [{
    "address": {
      "socketAddress": {
          "address": "127.0.0.1",
          "portValue": 1234
      }
    }
  }]
]}
```

```
$ curl quilkin:8080/filterchain
{
  "filters": [{
    {
      "name": "quilkin.extensions.filters.debug.v1alpha1.Debug",
      "config": {
        id: "hello"
      }  
    }  
  }]
}
```

iffyio:
That returns a JSON of the current config the proxy is running with

This includes a breaking change, as the FilterFactory to now needs to return
the config in addition to the filter it created.

Resolves #394

### DIFF
--- a/docs/src/filters/writing_custom_filters.md
+++ b/docs/src/filters/writing_custom_filters.md
@@ -20,7 +20,7 @@ A [trait][Filter] representing an actual [Filter][built-in-filters] instance in 
 A [trait][FilterFactory] representing a type that knows how to create instances of a particular type of [Filter].
 
 - An implementation provides a `name` and `create_filter` method.
-- `create_filter` takes in [configuration][filter configuration] for the filter to create and returns a new instance of its filter type.
+- `create_filter` takes in [configuration][filter configuration] for the filter to create and returns a [CreatedFilter] type containing a new instance of its filter type.
 `name` returns the Filter name - a unique identifier of filters of the created type (e.g quilkin.extensions.filters.debug.v1alpha1.Debug).
 
 ##### FilterRegistry
@@ -95,7 +95,7 @@ To extend Quilkin's code with our own custom filter, we need to do the following
    # struct Greet;
    # impl Filter for Greet {}
    # use quilkin::filters::Filter;
-   use quilkin::filters::{CreateFilterArgs, Error, FilterFactory};
+   use quilkin::filters::{CreatedFilter, CreateFilterArgs, Error, FilterFactory};
 
    struct GreetFilterFactory;
    impl FilterFactory for GreetFilterFactory {
@@ -103,8 +103,9 @@ To extend Quilkin's code with our own custom filter, we need to do the following
            // We provide the name of filter that we defined earlier.
            NAME
        }
-       fn create_filter(&self, _: CreateFilterArgs) -> Result<Box<dyn Filter>, Error> {
-           Ok(Box::new(Greet))
+       fn create_filter(&self, _: CreateFilterArgs) -> Result<CreatedFilter, Error> {
+           let filter: Box<dyn Filter> = Box::new(Greet);
+           Ok((serde_json::Value::Null, filter).into())
        }
    }
    ```
@@ -124,13 +125,13 @@ To extend Quilkin's code with our own custom filter, we need to do the following
    Add a main function that starts the proxy.
    ```rust, no_run
    // src/main.rs
-   # use quilkin::filters::{CreateFilterArgs, Filter, Error, FilterFactory};
+   # use quilkin::filters::{CreatedFilter, CreateFilterArgs, Filter, Error, FilterFactory};
    # struct GreetFilterFactory;
    # impl FilterFactory for GreetFilterFactory {
    #     fn name(&self) -> &'static str {
    #         "greet.v1"
    #     }
-   #     fn create_filter(&self, _: CreateFilterArgs) -> Result<Box<dyn Filter>, Error> {
+   #     fn create_filter(&self, _: CreateFilterArgs) -> Result<CreatedFilter, Error> {
    #         unimplemented!()
    #     }
    # }
@@ -232,7 +233,7 @@ The [Serde] crate is used to describe static YAML configuration in code while [P
    # struct Config {
    #     greeting: String,
    # }
-   # use quilkin::filters::{CreateFilterArgs, Error, FilterFactory, Filter, ReadContext, ReadResponse, WriteContext, WriteResponse};
+   # use quilkin::filters::{CreatedFilter, CreateFilterArgs, Error, FilterFactory, Filter, ReadContext, ReadResponse, WriteContext, WriteResponse};
    # struct Greet(String);
    # impl Filter for Greet { }
    use quilkin::config::ConfigType;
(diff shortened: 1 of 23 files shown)


iffyio:
example output
```yaml
{
  "clusters": [
    {
      "name": "default-quilkin-cluster",
      "endpoints": [
        {
          "address": "127.0.0.1:4321",
          "metadata": {
            "quilkin.dev": {
              "tokens": [
                "AA=="
              ]
            }
          }
        }
      ]
    }
  ],
  "filterchain": {
    "filters": [
      {
        "name": "quilkin.extensions.filters.capture_bytes.v1alpha1.CaptureBytes",
        "config": {
          "metadataKey": "quilkin.dev/captured_bytes",
          "remove": true,
          "size": 1,
          "strategy": "SUFFIX"
        }
      },
      {
        "name": "quilkin.extensions.filters.token_router.v1alpha1.TokenRouter",
        "config": {
          "metadataKey": "quilkin.dev/captured_bytes"
        }
      }
    ]
  }
}
```

iffyio:
Also this has a bunch of conflicts with #395 in the custom filter doc at least, I'll resolve them after that's merged

src/config/config_type.rs
```diff
@@ -49,7 +74,63 @@ impl ConfigType<'_> {
...
+            .and_then(|config_json| {
+                serde_json::from_str::<serde_json::Value>(config_json.as_str()).map_err(|err| {
+                    Error::DeserializeFailed(format!(
+                        "filter `{}`: failed to deserialize into json: {}",
+                        filter_name,
+                        err.to_string()
+                    ))
+                })
+            })
+            .map(|config_json| (config_json, config))
```
XAMPPRocky, line 104: If you take it by reference here, you don't need to pass it through the function and back to the caller, and we can also remove level of serialisation.

```suggestion
    fn get_json_config<T>(filter_name: &str, config: &T) -> Result<serde_json::Value, Error>
    where
        T: serde::Serialize + for<'de> serde::Deserialize<'de>,
    {
        serde_json::to_value(config.clone())
            .map_err(|err| {
                Error::DeserializeFailed(format!(
                    "filter `{}`: failed to serialize config to json: {}",
                    filter_name,
                    err.to_string()
                ))
            })
```

src/config/config_type.rs
```diff
@@ -30,17 +30,42 @@ pub enum ConfigType<'a> {
...
 
 impl ConfigType<'_> {
-    /// Deserializes the configuration to `T` based on the input type. Errors if
-    /// the data produces an invalid config.
-    pub fn deserialize<T, P>(self, filter_name: &str) -> Result<T, Error>
+    /// Deserializes takes two type arguments `Static` and `Dynamic` representing
+    /// the types of a static and dynamic configuration respectively.
+    ///
+    /// If the configuration input is a [ConfigType::Static], then it is directly
+    /// marshalled into the returned `Static` instance.
```
XAMPPRocky, line 37: Very minor nit: The "Marshall" terminology isn't what you'd use in Rust. You'd normally just say "serialisation/deserialisation", I guess it's because in Rust they're not really objects in the same way they are in Go or Python. I don't think it's too important though.

src/filters/factory.rs
```diff
@@ -24,6 +24,23 @@ use crate::{
 /// An owned pointer to a dynamic [`FilterFactory`] instance.
 pub type DynFilterFactory = Box<dyn FilterFactory>;
 
+/// The value returned by [`FilterFactory::create_filter`].
+pub struct CreatedFilter {
+    /// The configuration used to create the filter.
+    pub config: serde_json::Value,
```
XAMPPRocky, line 30: Since most of the time we create a new a filter the configuration doesn't change, as a future improvement, we could use `Arc` here instead to share memory across filter instances.

src/filters/chain.rs
```diff
@@ -124,7 +128,7 @@ impl Filter for FilterChain {
             .try_fold(ctx, |ctx, ((_, filter), histogram)| {
                 Some(ReadContext::with_response(
                     ctx.from,
-                    histogram.observe_closure_duration(|| filter.read(ctx))?,
+                    histogram.observe_closure_duration(|| filter.filter.read(ctx))?,
```
XAMPPRocky, line 131: Along with changing the type to `FilterInstance` I would change these to call the top level variable `instance`. 
```suggestion
                    histogram.observe_closure_duration(|| instance.filter.read(ctx))?,
```

src/config/config_type.rs
```diff
@@ -30,17 +30,42 @@ pub enum ConfigType<'a> {
...
+        self,
+        filter_name: &str,
+    ) -> Result<(serde_json::Value, Static), Error>
     where
-        P: prost::Message + Default,
-        T: for<'de> serde::Deserialize<'de> + TryFrom<P, Error = ConvertProtoConfigError>,
+        Dynamic: prost::Message + Default,
+        Static: serde::Serialize
+            + for<'de> serde::Deserialize<'de>
+            + TryFrom<Dynamic, Error = ConvertProtoConfigError>,
```
XAMPPRocky, line 56: This doesn't have to be in this PR, but I would like to combine these two type parameters into single trait, so that we can enforce it as part of the type system. Then you also only need to provide a single parameter.

```rust
trait FilterConfig {
   type Dynamic: prost::Message + Default;
   type Static: serde::Serialize
             + for<'de> serde::Deserialize<'de>
             + TryFrom<Self::Dynamic, Error = ConvertProtoConfigError>;
}

// src/filters/debug.rs
impl FilterConfig for Config {
    type Static = Self;
    type Dynamic = ProtoDebug;
}
```

src/filters/factory.rs
```diff
@@ -24,6 +24,23 @@ use crate::{
...
+    /// The created filter.
+    pub filter: Box<dyn Filter>,
+}
+
+impl From<(serde_json::Value, Box<dyn Filter>)> for CreatedFilter {
```
XAMPPRocky, line 35: I'm not sure about this `From` implementation, if we add new things to `CreatedFilter` we'll have to change it, and it feels odd that `(Value, Box)` would work but `(Box, Value)` won't. What would you think about having a `new` method instead?

src/proxy/admin.rs
```diff
@@ -77,14 +93,13 @@ impl Admin {
...
-    metrics: Arc<Metrics>,
-    health: Arc<Health>,
-) -> Response<Body> {
+fn handle_request(request: Request<Body>, args: HandleRequestArgs) -> Response<Body> {
     match (request.method(), request.uri().path()) {
-        (&Method::GET, "/metrics") => metrics.collect_metrics(),
-        (&Method::GET, "/live") => health.check_healthy(),
+        (&Method::GET, "/metrics") => args.metrics.collect_metrics(),
+        (&Method::GET, "/live") => args.health.check_healthy(),
+        (&Method::GET, "/config_dump") => {
```
XAMPPRocky, line 100:
```suggestion
        (&Method::GET, "/config") => {
```

XAMPPRocky (commented):
Looks good to me, mostly just minor naming concerns, some code changes, and some potential improvements that can come later.

markmandel:
Since #395 is about ready to be merged (I need to make a small change, and it's good to go) -- since this is a particularly useful feature, should we wait until it's merged and bundle it into the 0.2.0 release?

iffyio:
> Since #395 is about ready to be merged (I need to make a small change, and it's good to go) -- since this is a particularly useful feature, should we wait until it's merged and bundle it into the 0.2.0 release? (I'm leaning towards "yes", personally)

yes please :pray: also so we can avoid using a custom branch in our deployment internally until the next release

markmandel:
Added to the milestone!

docs/src/filters/writing_custom_filters.md
```diff
@@ -206,14 +207,70 @@ First let's create the config for our static configuration:
 {{#include ../../../examples/quilkin-filter-example/src/main.rs:filter}}
 ```
 
+###### 4. Finally, update `GreetFilterFactory` to extract the greeting from the passed in configuration and forward it onto the `Greet` Filter.
```
markmandel, line 210: I really like this being back in - since it slowly introduces people to the config option, and then we can move into `deserialize`, which is the more complex topic. 👍🏻

markmandel (approved):
Looks great to me!

If @XAMPPRocky is also happy, I'm very happy to have this merged 🥳

XAMPPRocky (approved):

PR #397, opened 2021-09-19
Fix duplicate metric registration for filters

iffyio:
Some filter currently call `register` (rather than
`register_if_not_exists`) which throws an error if the metric
has already been registered - this meant that we couldn't use
multiple instances of those filter in the filter chain

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -43,25 +43,22 @@ pub fn filter_opts(name: &str, filter_name: &str, description: &str) -> Opts {
     opts(name, &format!("filter_{}", filter_name), description)
 }
 
-pub trait CollectorExt: Collector + Clone + Sized + 'static {
-    /// Registers the current metric collector with the provided registry.
-    /// Returns an error if a collector with the same name has already
-    /// been registered.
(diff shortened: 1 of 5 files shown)


src/metrics.rs
```diff
@@ -43,25 +43,22 @@ pub fn filter_opts(name: &str, filter_name: &str, description: &str) -> Opts {
...
+/// Returns an error if a collector with the same name has already
+/// been registered.
+fn register_metric<T: Collector + Sized + 'static>(
+    registry: &Registry,
+    collector: T,
+) -> Result<()> {
+    registry.register(Box::new(collector))
+}
 
+pub trait CollectorExt: Collector + Clone + Sized + 'static {
```
XAMPPRocky, line 56: Nice change.

XAMPPRocky (approved):

iffyio:
@markmandel we can also include this fix in the upcoming release?

markmandel:
> @markmandel we can also include this fix in the upcoming release?

Let's do it! Going to start the release today, so this sneaks in nicely!

PR #417, opened 2021-10-10
Add a rate limited logger

iffyio:
We currently have a `LOG_SAMPLING_RATE` variable which is a more naive
solution to limit the messages we log. Instead, this
makes sure that we don't log over a threshold across the proxy in a
single time window.

Impl wise its a tiny wrapper (log.rs) around the log macros that check if we're
allowed to log a message and drops it otherwise. The remaining changes
update usages so even though this touches a few files its not as large a
change

This has a breaking change, now the function to create a `Builder` from `Config` takes in a logger

### DIFF
--- a/benches/throughput.rs
+++ b/benches/throughput.rs
@@ -31,10 +31,11 @@ static SERVER_INIT: Lazy<()> = Lazy::new(|| {
                 )],
             )
             .build();
-        let server = quilkin::Builder::from(std::sync::Arc::new(config))
-            .validate()
-            .unwrap()
-            .build();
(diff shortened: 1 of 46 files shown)


markmandel:
Just bumping this, as it's approved, but it just needs an update to fix the conflicts.

iffyio:
> Just bumping this, as it's approved, but it just needs an update to fix the conflicts.

Ah right I'll get back to it but probably after #385 is merged since this needs a complete rework for that. will probably open a new pr entirely in place of this

PR #430, opened 2021-11-02
Refactor AdsClient and switch backoff with tryhard

XAMPPRocky:
While looking at investigating how our code handles shutdown signals and seeing how we could improve it, and the code in the `AdsClient` was particularly gnarly to understand, so I've split up the responsibilities of the `AdsClient` into distinct types to split the code up, similar to how it was done in #339. Now instead of all the operations being associated functions on `AdsClient`, the responsiblities are split up between `RpcSender`, `RpcReceiver`, `RpcSession`, and `AdsClient`. Like #339 there's no actual behaviour change here.

- `RpcSender` is responsible for sending discovery requests.
- `RpcReceiver` is responsible for handling discovery responses.
- `RpcSession` encapsulates both `RpcSender` and `RpcReceiver` into a single session with a server.
- `AdsClient` handles `RpcSession` initiation, failure, and retrying.

This PR also replaces using `backoff` for the handling backoff on failure with `tryhard`. The motivation being that `tryhard` is a lot simpler conceptually and in implementation than `backoff`, with `backoff` we had to control the backoff ourselves, and pass state back and forth to ensure we can properly retry. Where as now with `tryhard` we don't have to do any of that, we just run the future, and that is all taken of for us.


src/xds/ads_client.rs
```diff
@@ -131,145 +90,192 @@ impl AdsClient {
...
-                backoff,
-                discovery_req_rx: &mut discovery_req_rx,
-                shutdown_rx: shutdown_rx.clone(),
+                shutdown_rx: session_shutdown_rx.clone(),
+            }
+            .run()
+        })
+        .with_config(retry_config);
+
+        tokio::select! {
```
markmandel, line 161: I'll defer the majority of the review to @iffyio who knows this code section far better than I - but I will definitely say that this section is far more clear 👍🏻

src/xds/ads_client.rs
```diff
@@ -131,145 +90,192 @@ impl AdsClient {
...
+            RpcSession {
+                discovery_req_rx,
                 log: log.clone(),
                 metrics: metrics.clone(),
-                server_addr: server_addr.clone(),
                 node_id: node_id.clone(),
+                addr: server_iter
+                    .next()
+                    .map(|server| server.address.to_string())
+                    .unwrap(),
```
iffyio, line 153: lets add a comment why this unwrap is safe?

src/xds/cluster.rs
```diff
@@ -339,13 +339,6 @@ impl ClusterManager {
         Ok(existing_endpoints)
     }
 
-    // Notify that we are about to reconnect the GRPC stream.
-    pub(in crate::xds) fn on_reconnect(&mut self) {
-        // Reset any last seen version and nonce since we'll be working
-        // with a new connection from now on with a clean slate.
-        self.last_seen_cluster_load_assignment_version = None
```
iffyio, line 346: this seems gone entirely?

iffyio (approved):

src/xds/cluster.rs
XAMPPRocky: Yes, because now on failure, we just reconstruct a new `ClusterManager` altogether.

https://github.com/googleforgames/quilkin/blob/2f2667b8881e99f3ccb8ee8c0a4df8e74ac23d96/src/xds/ads_client.rs#L129-L158

src/xds/ads_client.rs
```diff
@@ -131,145 +90,192 @@ impl AdsClient {
...
             };
 
-            let args = RpcSessionArgs {
+            RpcSession {
+                discovery_req_rx,
                 log: log.clone(),
                 metrics: metrics.clone(),
-                server_addr: server_addr.clone(),
                 node_id: node_id.clone(),
+                addr: server_iter
```
XAMPPRocky, line 150:
```suggestion
                // server_iter is guaranteed to always have at least one entry.
                addr: server_iter
```

src/xds/cluster.rs
iffyio: Ah got it :+1:

markmandel (approved):

PR #432, opened 2021-11-03
Documentation for Firewall filter

### LINKED ISSUE #158: Blocklist Filter

### LINKED ISSUE #343: Allowlist filter
In #158 you discussed a blocklist filter. It would be nice to have an allowlist filter or some way to accept connections only from a given address and port pair. 

In a telco use case in the initiating SDP messages, there is information about the client's sender socket like address and port. After these messages, we want to tell Quilkin to accept connections from that address and port pair and from nowhere else. This way you wouldn't have an open port for everyone, but only for the trusted one.

More info about a possible telco use-case: There is a separate control plane and a media plane. When the offer and answer come in from the caller and the callee we dynamically (via an XDS server) create a media plane pipeline for every single RTP/RTCP stream. We know where we should expect the streams from, due to the SDP messages sent previously on the control plane. We want to accept packets only from that single address port pair on the listeners(not sure how to call them here).

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Finishes off the Firewall Filter!

Closes #158
Closes #343

While writing the docs, I found a typo in the name of the Filter, so
included that as well.

### DIFF
--- a/docs/src/filters/firewall.md
+++ b/docs/src/filters/firewall.md
@@ -1,0 +1,96 @@
+# Firewall
+
+The `Firewall` filter's job is to allow or block traffic depending on if the incoming traffic's IP and port matches
+the rules set on the Firewall filter.
+
+#### Filter name
+```text
+quilkin.extensions.filters.firewall.v1alpha1.Firewall
+```
+
+### Configuration Examples
+```rust
+# let yaml = "
+version: v1alpha1
+static:
+  filters:
+    - name: quilkin.extensions.filters.firewall.v1alpha1.Firewall
+      config:
+        on_read:
+          - action: ALLOW
+            source: 192.168.51.0/24
+            ports:
+               - 10
+               - 1000-7000
+        on_write: 
+          - action: DENY
+            source: 192.168.51.0/24
+            ports:
+               - 7000
+  endpoints:
+    - address: 127.0.0.1:7001
+# ";
+# let config = quilkin::config::Config::from_reader(yaml.as_bytes()).unwrap();
+# assert_eq!(config.source.get_static_filters().unwrap().len(), 1);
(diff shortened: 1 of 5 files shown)


iffyio (approved):

docs/src/filters/firewall.md
```diff
@@ -0,0 +1,96 @@
...
+               - 7000
+  endpoints:
+    - address: 127.0.0.1:7001
+# ";
+# let config = quilkin::config::Config::from_reader(yaml.as_bytes()).unwrap();
+# assert_eq!(config.source.get_static_filters().unwrap().len(), 1);
+# quilkin::Builder::from(std::sync::Arc::new(config)).validate().unwrap();
+```
+
+### Configuration Options
```
XAMPPRocky, line 38: Should this just link the `Config` struct in `rustdoc` instead?

markmandel: Both seems good! Then if you know Rust, you can look there, but if you don't you have an alternate documentation:

image

What do you think of that format? I must say, I quite like having both, gives you a few ways of working out what is possible depending on how you are interacting with Quilkin.

Obligatory:

XAMPPRocky (approved):

PR #435, opened 2021-11-10
Move xds-management-server to the top-level

XAMPPRocky:
After merging #360 I noticed that the project was in `examples/` which doesn't seem right, as it's designed to be used, not just as an example. So this just moves it to a top-level `xds/` folder instead.


markmandel:
🤔 @iffyio be curious to hear your thoughts on this.

My thought was that this would be _a_ way you could integrate Agones, but possibly not _the_ way? I.e. if you wanted to base your integration on this you could (which led me to have this in the `examples` folder), if we move it up to the top level, are we expected to put out a Docker image, etc, and make it a more complete solution?

I just want to manage expectations appropriately.

iffyio:
Ah so this is one server to use indeed, the idea isn't for folks to extend it. If someone wants something different they'll need to either roll their own implementation separately like we did here or fork this to add their change. We'll need docker images etc for this yeah

markmandel:
> Ah so this is one server to use indeed, the idea isn't for folks to extend it. If someone wants something different they'll need to either roll their own implementation separately like we did here or fork this to add their change. We'll need docker images etc for this yeah

Got it. Awesome, then if that's the plan, let's definitely move this up a level.

markmandel (approved):

iffyio (approved):

PR #436, opened 2021-11-12
Add `metadata::Value`

### LINKED ISSUE #433: Changing dynamic metadata from `Any`.
Currently, out metadata type is `Box<std::any::Any + Send>`, which allows for it to be downcast to a concrete type. This is suitable when both the creator and reader know the concrete type of the metadata. However there are more cases where we want to be able to do something with the metadata without knowing the concrete type.

One example of this is if you where to have dynamic or scripted filters, another case is attempting to implement the `quilkin.match` filter described in #401. In order to be able to branch, we need to be able to compare the values in the config with the values in the metadata map, without knowing really knowing the concrete type. So we need a common type that can be compared

I'd like to instead propose that we switch from using `Box<std::any::Any>` to `serde_yaml::Value` as the base metadata type. This does potentially add a serialisation step, however we should discourage ourselves and users from serialising simple metadata, and to instead create `serde_yaml::Values` directly.

XAMPPRocky:
closes #433

### DIFF
--- a/src/filters/token_router.rs
+++ b/src/filters/token_router.rs
@@ -28,6 +28,7 @@ use crate::{
     config::LOG_SAMPLING_RATE,
     endpoint::RetainedItems,
     filters::{metadata::CAPTURED_BYTES, prelude::*},
+    metadata::Value,
 };
 
 use metrics::Metrics;
@@ -111,10 +112,10 @@ impl Filter for TokenRouter {
                 self.metrics.packets_dropped_no_token_found.inc();
                 None
             }
-            Some(value) => match value.downcast_ref::<Vec<u8>>() {
-                Some(token) => match ctx
+            Some(value) => match value {
+                Value::Bytes(token) => match ctx
                     .endpoints
-                    .retain(|e| e.metadata.known.tokens.contains(token))
+                    .retain(|e| e.metadata.known.tokens.contains(&**token))
                 {
                     RetainedItems::None => {
                         self.metrics.packets_dropped_no_endpoint_match.inc();
@@ -122,7 +123,7 @@ impl Filter for TokenRouter {
                     }
                     _ => Some(ctx.into()),
                 },
-                None => {
+                _ => {
                     if self.metrics.packets_dropped_invalid_token.get() % LOG_SAMPLING_RATE == 0 {
                         error!(
                             count = ?self.metrics.packets_dropped_invalid_token.get(),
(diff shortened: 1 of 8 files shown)


src/filters/token_router.rs
```diff
@@ -111,18 +112,18 @@ impl Filter for TokenRouter {
...
                         self.metrics.packets_dropped_no_endpoint_match.inc();
                         None
                     }
                     _ => Some(ctx.into()),
                 },
-                None => {
+                _ => {
                     if self.metrics.packets_dropped_invalid_token.get() % LOG_SAMPLING_RATE == 0 {
                         error!(
                             count = ?self.metrics.packets_dropped_invalid_token.get(),
```
iffyio, line 129: the error message here has a hardcoded `Vec<u8>` which is now out  of date

iffyio (commented):
LGTM! but let's update the section of filter metadata in the docs to reflect this change

src/filters/token_router.rs
XAMPPRocky: I'll update the error message, but I wonder if we should be logging an error here, instead of panicking and terminating the process. Since the type not matching would be the result of programmer error, not application error.

Like would there ever be a situation where the type would be correct 90% of the time? It seems like it would either, always be correct or always be wrong.

markmandel: > panicking and terminating the process

Maybe I'm missing something, but I don't see a panic here? We drop the packet if the type is not correct, and do a (sampled) log.

src/metadata.rs
```diff
@@ -14,12 +14,53 @@
...
+
 pub const KEY: &str = "quilkin.dev";
 
+#[derive(Clone, Debug, PartialEq, PartialOrd)]
+pub enum Value {
+    Bool(bool),
+    Number(u64),
+    List(Vec<Value>),
+    String(String),
+    Bytes(bytes::Bytes),
```
markmandel, line 32: I really like this 👍🏻

markmandel (approved):
This looks great to me. I'll let you two resolve the final comments, but 👍🏻  from me 😄

src/filters/token_router.rs
XAMPPRocky: @markmandel Right, I'm saying maybe we **should panic** here, because it would a programmer error.

iffyio: yeah its most likely a programmer error (it could also be a buggy control plane e.g after/during an upgrade in which case we might want to keep running so that a new filterchain update is eventually delivered to the proxy) but I don't see a need to panic as a result? we'll end up dropping the packet and logging and metrics. Compared to those, panicking seems harsh and we'd need to document the behavior too which doesn't seem worth it

iffyio (approved):

PR #437, opened 2021-11-12
xds-server updates

iffyio:
- Adds health probes and configurable logging
- Fixes a bug where we pass each endpoint as its own locality
  rather than all in the same locality
- Add support for routing tokens - if a routing token annotation
  is set, then the server configures the proxy with a
  `CaptureBytes=>TokenRouter` in the filter chain.
- Adds a simple dockerfile that can be used to build the server


xds/ci/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM golang:1.16-alpine
```
XAMPPRocky, line 1: Licence header needed.

XAMPPRocky (approved):
LGTM, though I'll obviously defer to @markmandel on the Go code.

xds/pkg/server/server.go
```diff
@@ -33,40 +38,99 @@ const (
 	grpcMaxConcurrentStreams = 1000000
 )
 
+// HealthCheck implements the health check logic exposed by the server.
+// It is implemented by internal components of the server and returns
+// an error if the component is unhealthy.
+type HealthCheck func(ctx context.Context) error
```
markmandel, line 44: Just a thought: I've used https://github.com/heptiolabs/healthcheck in the past (not that it's been updated in a while) to build out composable health checks. Just a thought that it might save some work building this out ourselves.

But it's also easy enough to do that it's not a huge deal, just thought I'd throw the idea out there.

xds/pkg/server/server.go
```diff
@@ -33,40 +38,99 @@ const (
...
-		port:          port,
+		xdsPort:       xdsPort,
 		snapshotCache: snapshotCache,
 		nodeIDCh:      nodeIDCh,
+		healthChecks:  healthChecks,
+		adminPort:     adminPort,
 	}
 }
 
+func (s *Server) startAdminServer(ctx context.Context, address string) {
```
markmandel, line 84:
```suggestion
func (s *Server) Run(ctx context.Context) {
```

Move address into `New`, and then this object matches the New/Run pattern as everywhere else.

xds/pkg/server/server.go
```diff
@@ -33,40 +38,99 @@ const (
...
+	healthChecks []HealthCheck
+	// adminPort is the port the admin server runs on.
+	adminPort int16
 }
 
 // NewServer returns a new Server
 func New(
 	logger *log.Logger,
-	port int16,
+	xdsPort int16,
```
markmandel, line 66: This should fail linting 🤔 , since the function name doesn't match the comment (and it should be `NewServer`)

xds/cmd/controller.go
```diff
@@ -70,18 +97,53 @@ func createAgonesClusterProvider(
 		externalversions.WithNamespace(flags.GameServersNamespace))
 	informerFactory.Start(ctx.Done())
 
+	gameServerInformer := informerFactory.Agones().V1().GameServers().Informer()
+	err = gameServerInformer.SetWatchErrorHandler(func(r *cache.Reflector, err error) {
```
markmandel, line 101: I've not seen `SetWatchErrorHandler` before!

That makes sense, given you aren't writing back to the K8s control plane, you can't pick up disconnections otherwise 👍🏻

Question: Wouldn't the health check ping on the k8s control plane capture this, without the extra handler? (I mean, you get more logs this way, but just thinking about it, once you know it's disconnected, it's because the whole thing has gone away - it would be very weird if you disconnected from just GameServers and not Pods for example -- it's the same backend).

xds/cmd/controller.go
```diff
@@ -93,20 +155,52 @@ func createFilterChainProvider(
...
-		flags.ProxyPollInterval)
+		flags.ProxyPollInterval), createFilterChainProviderHealthCheck(flags.ProxyNamespace, k8sClient)
+}
+
+func createFilterChainProviderHealthCheck(
+	podNamespace string,
+	k8sClient *kubernetes.Clientset,
+) server.HealthCheck {
+	return func(ctx context.Context) error {
+		if _, err := k8sClient.
```
markmandel, line 182: I don't think you need to check both. I think it would be sufficient to check just GameServers (which also checks that the CRDs are installed correctly). I don't think there's a scenario where you can get at one not the other. 

To be devil's advocate for myself -- RBAC rules could restrict access, but you will see that in the logs pretty quick. It's not likely be a runtime issue like a K8s control plane going down. It's all one etcd instance behind Pods and GameServers anyway.

Thoughts?

xds/pkg/server/server.go
iffyio: Yeah I think its simple enough in this case to not drag the dependency

xds/pkg/server/server.go
iffyio: updated to create address within the function, this is a helper function that is already being called from Server.Run

xds/pkg/server/server.go
iffyio: not sure how to configure it, golint fails as expected but golangci-lint doesn't do this check

xds/cmd/controller.go
iffyio: removed

xds/cmd/controller.go
iffyio: yes its mainly for rbac, I'd rather the pod reports itself as unhealthy if its not able to function properly especially during install/startup or if someone messes with policies at runtime. it seems like an easy win here to be able to detect it

xds/pkg/server/server.go
markmandel: Here's the config we use for Agones:
https://github.com/googleforgames/agones/blob/97dbf4677ac1c5223e434bca8ffc84654f6ae851/.golangci.yml#L47-L65

There's lots more linters 😃 (want to do it here, or in aseparatee PR?)

markmandel: TIL ([source](https://go.dev/blog/package-names#:~:text=When%20a%20function%20returns%20a%20value%20of%20type%20pkg.T%2C%20where%20T%20is%20not%20Pkg%2C%20the%20function%20name%20may%20include%20T%20to%20make%20client%20code%20easier%20to%20understand.%20A%20common%20situation%20is%20a%20package%20with%20multiple%20New%2Dlike%20functions%3A))

> When a function returns a value of type pkg.T, where T is not Pkg, the function name may include T to make client code easier to understand. A common situation is a package with multiple New-like functions:

So since this returns a Server, `New` is perfectly acceptable as a constructor name!

xds/ci/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+COPY go.mod ./
+COPY go.sum ./
+RUN go mod download
+
+COPY cmd/ cmd/
+COPY pkg/ pkg/
+
+RUN go build -o /server cmd/controller.go
+
+FROM alpine:latest
```
markmandel, line 29: This should run as a non-root user.

I would also recommend pinning an alpine version, I think major.minor is fine though, since alphine 

Actually, if we are going down this path, our other image is distroless -- I would suggest keeping one base image rather than two. distroless has a very low surface area, and also makes it very easy to run as non root in an already setup best-practice environment.

Go example:
https://github.com/GoogleContainerTools/distroless/blob/main/examples/go/Dockerfile
Although we would want to run gcr.io/distroless/static:nonroot

Local example:
https://github.com/googleforgames/quilkin/blob/6a0762f4d96e33818d3694036b2b166e7a1f4d14/image/Dockerfile#L17

xds/pkg/server/server.go
```diff
@@ -33,40 +38,101 @@ const (
...
 }
 
+func (s *Server) startAdminServer(ctx context.Context) {
+	address := fmt.Sprintf(":%d", s.adminPort)
+
+	srv := &http.Server{Addr: address}
+
+	http.Handle("/metrics", promhttp.Handler())
+
+	http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
```
markmandel, line 91: Is this meant to be a liveness probe or a readiness probe?

It would be good if this was `/live` or `/ready` so that it was clear.

I'm assuming that it's a readiness probe, in that if the k8s control plane goes down, we don't want to keep restarting this process because the health check is failing.

In that vein, it's probably useful to have both, as if the process can't respond to a simple http response, it should probably be restarted.

But if the k8s api is down we should keep logging this via liveness, but not restart the process.

iffyio: Updated to /ready, I assume if folks want to use it for liveness they can point to the same endpoint as well?

xds/ci/Dockerfile
iffyio: Fixed!

xds/pkg/server/server.go
iffyio: Updated to use the same linters, (though none of those still catches the functionname/comment matching)

xds/pkg/server/server.go
markmandel: Yeah they could. But I don't think people will want to restart the container if the K8s control plane goes down (but it's a choice 😁)

Doesn't have to be this PR, but I'd be tempted to have a `/live` endpoint as well that was just a Http Handler that returns 200. 

Reason being that if the process does freeze (and can't respond to just a simple http request) for whatever reason, then it will restart and hopefully recover again.

xds/pkg/server/server.go
markmandel: Weird. 🤔  Seems fine now though.

markmandel (approved):
Love it!

markmandel:
> Doesn't have to be this PR, but I'd be tempted to have a /live endpoint as well that was just a Http Handler that returns 200.

Only a minor thought for me - so if you want to merge as is, please feel free. Just leaving it open if it resonated and you wanted to do it with this PR 👍🏻

Issue #292, opened 2021-06-04
Metric: Total Packet Processing time


markmandel:
We now have per-filter metrics processing time (#262), but we don't currently track the total processing time from packet receive to send from local port to endpoint to back again.

I don't think we can use a sum of the filter processing times, since we would want to observe the general Quilkin overhead that gets added on top of filters (also, sometimes, there is no filter!).

_Design idea:_
On receipt of a incoming packet, we can attach at `prometheus::HistogramTimer` to the data structure we create, and then called `observe_duration()` just prior to the packet being sent.

markmandel:
Working on this next. Been wanting this for a while.

Worklist:

* [ ] Metrics for data coming into the local post (read)
* [ ] Metrics for data coming into an Endpoint (write)
* [ ] Example chart for the Grafana dashboard with these metrics.

PR #441, opened 2021-11-18
Metric: Total Packet processing time

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

* Added processing time histogram into proxy metrics.
* Moved event label keys and values into centralised const.
* Renamed Packer to UpstreamPacket
* Create DownstreamPacket struct
* Added SessionArg to Session constructor
* Updating documentation.

Work on #292

### DIFF
--- a/docs/src/proxy.md
+++ b/docs/src/proxy.md
@@ -41,10 +41,11 @@ A session represents ongoing communication flow between a client and an [Upstrea
 
 The proxy exposes the following general metrics (See the metrics sub-sections for metrics specific to other Quilkin components, e.g for metrics related to packet flow see [sessions metrics][session-metrics], or metrics exported by individual filters can be found in the documentation for each filter):
 
-- `quilkin_proxy_processing_time_seconds{event}` (Histogram)
+- `quilkin_proxy_packet_processing_duration_seconds{event}` (Histogram)
 The total duration of time in seconds that it took to process a packet.
-  * The `event` label is either `read` or `write` which corresponds to either `on_read` or `on_write` execution within 
-    the proxy.
+  * The `event` label is either:
+      * `read`: when the proxy receives data from a downstream connection on the listening port.
+      * `write`: when the proxy sends data to a downstream connection via the listening port.
 
 - `quilkin_proxy_packets_dropped_total{reason}` (Counter)
 
--- a/docs/src/proxy.md
+++ b/docs/src/proxy.md
@@ -41,6 +41,11 @@ A session represents ongoing communication flow between a client and an [Upstrea
 
 The proxy exposes the following general metrics (See the metrics sub-sections for metrics specific to other Quilkin components, e.g for metrics related to packet flow see [sessions metrics][session-metrics], or metrics exported by individual filters can be found in the documentation for each filter):
 
+- `quilkin_proxy_processing_time_seconds{event}` (Histogram)
+The total duration of time in seconds that it took to process a packet.
+  * The `event` label is either `read` or `write` which corresponds to either `on_read` or `on_write` execution within 
+    the proxy.
+
 - `quilkin_proxy_packets_dropped_total{reason}` (Counter)
 
   The total number of packets (not associated with any session) that were dropped by proxy.
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -18,6 +18,15 @@ use prometheus::core::Collector;
 pub use prometheus::Result;
 use prometheus::{HistogramOpts, Opts, Registry, DEFAULT_BUCKETS};
 
+/// "event" is used as a label for Metrics that can apply to both Filter
+/// `read` and `write` executions.
+pub const EVENT_LABEL: &str = "event";
+
+/// Label value for [EVENT_LABEL] for `read` events
+pub const EVENT_READ_LABEL_VALUE: &str = "read";
+/// Label value for [EVENT_LABEL] for `write` events
+pub const EVENT_WRITE_LABEL_VALUE: &str = "write";
+
 /// Create a generic metrics options.
 /// Use [filter_opts] instead if the intended target is a filter.
 pub fn opts(name: &str, subsystem: &str, description: &str) -> Opts {
--- a/src/proxy/server.rs
+++ b/src/proxy/server.rs
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
-use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
+use std::net::{Ipv4Addr, SocketAddrV4};
 use std::sync::Arc;
 
+use prometheus::HistogramTimer;
 use slog::{debug, error, info, trace, warn, Logger};
 use tokio::net::UdpSocket;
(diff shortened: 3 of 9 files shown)


src/proxy/sessions/session.rs
```diff
@@ -100,36 +111,44 @@ impl Packet {
...
         &self.contents
     }
+
+    /// stop and record the stored [HistogramTimer]
+    pub fn stop_and_record(self) {
+        self.timer.stop_and_record();
+    }
+}
+
+pub struct SessionArgs {
```
markmandel, line 121: I did this because Clippy complained there were too many arguments going to `Session::new(....)` -- I'm not sure if this actually improves things? It seems to just push the problem down a level?

I'd love to hear if there is a better way of doing this?

src/metrics.rs
```diff
@@ -18,6 +18,15 @@ use prometheus::core::Collector;
 pub use prometheus::Result;
 use prometheus::{HistogramOpts, Opts, Registry, DEFAULT_BUCKETS};
 
+/// "event" is used as a label for Metrics that can apply to both Filter
+/// `read` and `write` executions.
+pub const EVENT_LABEL: &str = "event";
```
markmandel, line 23: Is there a way to do this with a type system? I couldn't think of one.

docs/src/proxy.md
```diff
@@ -41,6 +41,11 @@ A session represents ongoing communication flow between a client and an [Upstrea
 
 The proxy exposes the following general metrics (See the metrics sub-sections for metrics specific to other Quilkin components, e.g for metrics related to packet flow see [sessions metrics][session-metrics], or metrics exported by individual filters can be found in the documentation for each filter):
 
+- `quilkin_proxy_processing_time_seconds{event}` (Histogram)
```
iffyio, line 44:
```suggestion
- `quilkin_proxy_packet_processing_duration_seconds{event}` (Histogram)
```
we can clarify that it only concerns packets, then `duration` is more conventional for prometheus histograms

docs/src/proxy.md
```diff
@@ -41,6 +41,11 @@ A session represents ongoing communication flow between a client and an [Upstrea
...
+The total duration of time in seconds that it took to process a packet.
+  * The `event` label is either `read` or `write` which corresponds to either `on_read` or `on_write` execution within
```
iffyio, line 46: > corresponds to either `on_read` or `on_write` execution within the proxy.


this reads a bit confusing, I don't think we use `on_read`/`on_write` in our docs/code but rather `read`/`write` if this is referring to filters? If this isn't filters specific but on the proxy level we should probably instead use the upstream/downstream terminology we have since read/write is filter specific

src/proxy/server.rs
```diff
@@ -62,17 +68,25 @@ struct RunRecvFromArgs {
...
     session_ttl: Duration,
-    send_packets: mpsc::Sender<Packet>,
+    send_packets: mpsc::Sender<UpstreamPacket>,
     shutdown_rx: watch::Receiver<()>,
 }
 
+/// Packet received from local port
+#[derive(Debug)]
+struct DownstreamPacket {
+    from: EndpointAddress,
```
iffyio, line 78:
```suggestion
    from: EndpointAddress,
```
nit: let's call this `source` rather than `from` - since we already talk about src/dst instead of from/to?

iffyio: my preference would be to use src/dest and we can go through the code base and update usages. but in any case neither change is related to this PR and we can leave this as is

iffyio:
@markmandel I think you might've forgot to push? I couldn't find new commits since last review in the PR

markmandel:
> @markmandel I think you might've forgot to push? I couldn't find new commits since last review in the PR

🤦🏻 🤣 yep, I did forget to push. I'm silly.

src/proxy/server.rs
markmandel: I actually don't have strong opinions. src/dest is fine by me.

src/metrics.rs
XAMPPRocky: You could have a trait with associated constants, though I'm not sure how generally applicable it might be.

```rust
pub trait FilterMetricLabels {
    const LABEL: &'static str;
    const READ: &'static str = "read";
    const WRITE: &'static str = "write";
}

impl FilterMetricsLabel for firewall::Metrics {
    const LABEL: &'static str = "event";
}
```

src/proxy/server.rs
XAMPPRocky: 👍 on source and destination.

src/proxy/server/metrics.rs
```diff
@@ -14,18 +14,41 @@
...
+        let event_labels = &[EVENT_LABEL];
+
+        let processing_time = HistogramVec::new(
+            histogram_opts(
+                "packet_processing_duration_seconds",
+                subsystem,
+                "Total processing time for a packet",
+                // Less than a millisecond is good, so starting at a quarter of that.
+                // Any processing that goes over a second is way too long, so ending there.
+                Some(exponential_buckets(0.00025, 2.0, 13).unwrap()),
```
XAMPPRocky, line 46: I'd maybe make these numbers `const`s, as even with the comment they feel pretty "magic", like why `13`?

src/proxy/sessions/session.rs
XAMPPRocky: Instead of `SessionArgs`, and passing that to `Session::new` I would instead make `new` a method on `SessionArgs` similar to the pseudo-actor run pattern I've been using elsewhere.


```rust
SessionArgs {
    metrics,
    proxy_metrics,
    filter_manager,
    from,
    dest,
    sender,
    ttl,
}.new(logger)
```

XAMPPRocky (approved):
LGTM, a few comments but nothing that would block merging.

iffyio (approved):

src/proxy/sessions/session.rs
```diff
@@ -100,36 +111,51 @@ impl Packet {
...
-    pub async fn new(
-        base: &Logger,
-        metrics: Metrics,
-        filter_manager: SharedFilterManager,
-        from: EndpointAddress,
-        dest: Endpoint,
-        sender: mpsc::Sender<Packet>,
-        ttl: Duration,
-    ) -> Result<Self> {
+    pub async fn into_session(self, base: &Logger) -> Result<Session> {
```
markmandel, line 134: @XAMPPRocky this is definitely much cleaner 👍🏻 

Clippy complained about `new` as a function name, since it wasn't returning `Self`. I thought `into_session` made sense here (and I couldn't find any Rust naming conventions that said otherwise). PTAL, let me know what you think.

markmandel:
Made some more review updates. I think it looks better now, but please let me know if I misconstrued any feedback. 👍🏻

src/proxy/sessions/session.rs
XAMPPRocky: Yeah, this works, I usually call it `run`, but there's no convention here.

PR #446, opened 2021-11-26
xds-server: Create informers before starting informer factory

cheahjs:
**What type of PR is this?**
/kind bug

Fixes the Kubernetes controller implementation of the xds management server. Without the fix, the controller isn't able to list any `GameServers` or `Pods` in the cluster, and doesn't update connected Quilkin instances with endpoints.

The informer factory only starts informers that were already requested when `Start()` is called. The informers are requested when getting the Lister for a specific type. As we are calling `Lister()` after `Start()`, the informer never watches for the resources, and `List()` will always return the empty list.

### DIFF
--- a/xds/cmd/controller.go
+++ b/xds/cmd/controller.go
@@ -68,9 +68,12 @@ func createAgonesClusterProvider(
 		agonesClient,
 		0,
 		externalversions.WithNamespace(flags.GameServersNamespace))
+
+	gameServerLister := informerFactory.Agones().V1().GameServers().Lister()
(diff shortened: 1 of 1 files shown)


XAMPPRocky (approved):
Thank you for your PR! LGTM, but I'll wait for @markmandel || @iffyio to have a look before merging.

iffyio (approved):
LGTM! The order of things got shifted wrong in the pr it seems, Thanks for finding and fixing it :tada:

PR #447, opened 2021-11-29
Add matches filter

### LINKED ISSUE #401: Add support for version based packet processing
The problem

The CaptureBytes filter is currently the primary means in quilkin to extract information from packets. Packets have a fixed format that determines where to find that info. In practice however that format can change over time or differ between packets - as a result it would be desirable to capture (potentially different) info based on the format of the particular packet in question.

For context, we're currently trying out UnrealEngine integration where the client tags each packet with a routing token but we can't actually put this in production unless we have a backwards compatibilty plan in place otherwise we'll probably have to run specific proxies with different filter chains for our different clients over time to avoid breaking old clients whenever we add changes to the packet format.
We'd also like to use the same set of proxies to handle packets across all our games and not necessarily be tied to having them all use the same packet format.

For example, right now due to some UE limitations, we need to tag our token onto packets as a prefix but we hope to resolve this upstream in the future and then switch to appending the token as a suffix instead. Also, today we only send a routing token but we might want to include more information later on - like when we add support for setting up a connection/authentication.
In any case, we don't want to break previous clients that only use a prefixed token.

Proposal

I think we can extend the CaptureBytes filter to
1. be _version aware_ (i.e know that a packet has a version)
2. capture multiple pieces of information from a packet (currently it can only extract a single piece of info) -

by configuring the filter to first, fetch version information either from the packet itself or from dynamic metadata - and then based on the value of that found version, execute a list of captures.
We should also have support for what to do if the version we got doesn't have any registered list of captures - allow the packet or drop the packet, the former lets us bypass the filter for packets that do not need to be processed by it (e.g some version x packet has no interesting info for the filter to capture).

config wise the filter's schema becomes something like:

```yaml
# The config is one of:
config:
  # A versioned, capture list
  versioned: versioned
  # Short hand for a non-versioned, capture list of 1 - which is how is works today.
  capture: capture
  # A list of captures to perform, non-versioned.
  captures: [ capture ]
```

```yaml
versioned:
  # Where to retrieve what version to use.
  # For each packet the version is always retrieved first.
  version: version
  # A list of (version, captures) pair checked in list order to figure out what capture list
  # to use.
  captures: [ versionedCapture ]
  # What to do if we don't find a capture list to process the packet with.
  onNoVersionMatched: DROP | PASSTHROUGH
```

```yaml
# Version is retrieved from one of
version:
  # A capture from the packet
  packet: capture
  # Or from dynamic metadata (e.g in cases where you have multiple CaptureByte
  # filters in the chain or the version is already known from some other means).
  metadataKey: string
```

```yaml
versionedCapture:
  # If the captured version matches this value
  version: base64Bytes
  # then use this list to capture info from the packet.
  captures: [ capture ]
```

```yaml
# This part remains largely the same as today.
capture:
  strategy: PREFIX | SUFFIX
  # This becomes optional - for cases where we don't necessarily want to store the
  # captured value (like the version maybe could only be captured and thrown away after usage)
  metadataKey: string
  size: int
  remove: bool
```

As an example configuration, the following shows a CaptureBytes that has evolved over time and is backwards compatible with older clients+packet formats.
```yaml
config:
  versioned:
    version:
      packet:
        strategy: PREFIX
        size: 1
        metadataKey: packet-version
        remove: true
    captures:

    # v0 has only a single token prefix
    - version: AA==
      captures:
      - strategy: PREFIX
        size: 16
        metadataKey: routing-token
        remove: true

    # In v1 the token is now a suffix
    - version: AQ==
      captures:
      - strategy: SUFFIX
        size: 16
        metadataKey: routing-token
        remove: true

    # In v2 we have both a token and some other piece of information attached
    - version: Ag==
      captures:
      # First capture the routing token
      - strategy: SUFFIX
        size: 16
        metadataKey: routing-token
        remove: true
      # Then capture the extra info
      - strategy: SUFFIX
        size: 32
        metadataKey: extra-info
        remove: true
```

XAMPPRocky:
Adds a minimal version of the quilkin matches filter discussed in #401.

I believe closes #401, #412 

## Breaking Changes

- `ConfigType::Static` now accepts `serde_yaml::Value` by value instead of by reference.


src/filters/matches.rs
```diff
@@ -0,0 +1,159 @@
...
+/// between read and write is which config and direction is called.
+macro_rules! matches_filter {
+    ($this:ident, $context:ident, $direction:ident, $config:ident) => {{
+        let ctx = $context;
+        let config = match $this.config.$config.as_ref() {
+            Some(value) => value,
+            None => return Some(ctx.into()),
+        };
+
+        for branch in &config.branches {
```
iffyio, line 43: Let's increment counter metrics here for when we find a match vs not, for visibility

src/filters/matches.rs
```diff
@@ -0,0 +1,159 @@
...
+            if ctx.metadata.get(&config.metadata_key) == Some(&branch.value) {
+                let args = CreateFilterArgs::fixed(
+                    $this.filter_registry.clone(),
+                    $this.metrics_registry.clone(),
+                    branch.config.as_ref(),
+                )
+                .with_metrics_registry($this.metrics_registry.clone());
+
+                let instance = $this.filter_registry.get(&branch.filter, args).ok()?;
```
iffyio, line 52: This looks to be creating a new filter on each packet which wouldn't be the behavior we want? Rather Match would want to e.g instantiate the filters for all its arms once, when it itself is instantiated and that filter can keep state as usual?

src/filters/matches.rs
```diff
@@ -0,0 +1,159 @@
...
+struct Config {
+    on_read: Option<DirectionalConfig>,
+    on_write: Option<DirectionalConfig>,
+}
+
+impl TryFrom<ProtoConfig> for Config {
+    type Error = ConvertProtoConfigError;
+
+    fn try_from(_value: ProtoConfig) -> Result<Self, Self::Error> {
+        todo!()
```
iffyio, line 111: Just flagging so we don't forget to resolve this before merging!

src/filters/matches.rs
```diff
@@ -0,0 +1,159 @@
+crate::include_proto!("quilkin.extensions.filters.matches.v1alpha1");
+
+use self::quilkin::extensions::filters::matches::v1alpha1::Matches as ProtoConfig;
+use crate::filters::{prelude::*, registry::FilterRegistry};
+
+pub const NAME: &str = "quilkin.extensions.filters.matches.v1alpha1.Matches";
+
+/// Creates a new factory for generating capture filters.
```
iffyio, line 8: the comment isn't correct

src/filters/matches.rs
```diff
@@ -0,0 +1,159 @@
...
+pub fn factory() -> DynFilterFactory {
+    Box::from(MatchesFactory::new())
+}
+
+struct Matches {
```
iffyio, line 13: Let's add comments for this and the factory struct describing them since they're public in the rustdoc iirc

iffyio: `Match` could be a better name for the filter? think I remember a while ago @markmandel and I had discussed having a convention for filters and to use verbs like compress/encrypt where possible (not sure how well we've kept to it ^^)

src/filters/matches.rs
```diff
@@ -0,0 +1,159 @@
...
+
+#[derive(Debug, serde::Deserialize, serde::Serialize, PartialEq)]
+struct Branch {
+    value: crate::metadata::Value,
+    filter: String,
+    config: Option<serde_yaml::Value>,
+}
+
+#[cfg(test)]
+mod tests {
```
iffyio, line 129: I'd like for us to add some tests either here or as integration tests etc that exercises more functionality of the filter, like that we call read/write, we skip branches and select the correct branch when they're multiple branches involved and we drop a packet or forward it when there's no match based on the config.

tests/matches.rs
```diff
@@ -0,0 +1,87 @@
...
+
+use quilkin::{
+    config::{Builder, Filter},
+    endpoint::Endpoint,
+    filters::{capture_bytes, matches},
+    test_utils::TestHelper,
+};
+
+/// This test covers both token_router and capture_bytes filters,
+/// since they work in concert together.
```
iffyio, line 29:
```suggestion
```

src/filters/matches.rs
XAMPPRocky: I was initially calling it `match`, the reason I initially went with matches is though is because `match` is a keyword in Rust, and I didn't want to to deal with any hassle that might come from that. Now that I've written it though, it should be easy enough to try and rename.

src/filters/matches.rs
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+}
+
+/// The actual behaviour of the matches filter, the only thing that changes
+/// between read and write is which config and direction is called.
+macro_rules! matches_filter {
+    ($this:ident, $context:ident, $direction:ident, $config:ident) => {{
+        let ctx = $context;
+        let config = match $this.config.$config.as_ref() {
```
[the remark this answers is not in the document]
XAMPPRocky: My reasoning was that it's optional for filters where you just want the default behaviour.

src/filters/matches.rs
```diff
@@ -0,0 +1,159 @@
...
+            config,
+            filter_registry,
+            metrics_registry,
+        }
+    }
+}
+
+/// The actual behaviour of the matches filter, the only thing that changes
+/// between read and write is which config and direction is called.
+macro_rules! matches_filter {
```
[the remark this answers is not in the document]
XAMPPRocky: Yeah, I can give it a shot rewriting this into a function, it seems like it might be possible, just will require a good chunk of generics.

src/metadata.rs
```diff
@@ -61,6 +62,43 @@ impl Value {
...
+impl PartialEq for Value {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Bool(a), Self::Bool(b)) => a == b,
+            (Self::Number(a), Self::Number(b)) => a == b,
+            (Self::List(a), Self::List(b)) => a == b,
+            (Self::String(a), Self::String(b)) => a == b,
+            (Self::Bytes(a), Self::Bytes(b)) => a == b,
+            (Self::String(a), Self::Bytes(b)) | (Self::Bytes(b), Self::String(a)) => a == b,
+            _ => false,
```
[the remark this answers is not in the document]
XAMPPRocky: > its not too far from doing vstring.as_bytes() == vbytes or similar wdyt?

Well it is more involved than that, because we'd have to check the variants as well whenever we wanted to do that check.

I think I'd rather just remove the `_ => false` branch and write out the rest of the variants, so we don't introduce a bug, because that is a good point.

src/filters/matches.rs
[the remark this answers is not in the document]
markmandel: Do we mind if it's mostly copy/paste between `read` and `write`? I know it's duplication, but it's only happening twice, and might just make things more readable? It's not a huge chunk of code.

(Or maybe break things out into several smaller helper functions even?)

src/filters/matches.rs
markmandel: +1 on this one - I figure it would make sense for Matches to container a `FilterChain`, which could be created on `new` and stored. Execution of `read` and `write` on `Match` should be pretty straight forward at that point I think.

src/filters/set.rs
```diff
@@ -52,14 +52,15 @@ impl FilterSet {
     pub fn default_with(filters: impl IntoIterator<Item = DynFilterFactory>) -> Self {
         Self::with(
             std::array::IntoIter::new([
-                filters::debug::factory(),
-                filters::local_rate_limit::factory(),
-                filters::concatenate_bytes::factory(),
-                filters::load_balancer::factory(),
                 filters::capture_bytes::factory(),
```
markmandel, line 55: Nice on the alphabetising! 👍🏻

src/proxy/health.rs
```diff
@@ -38,7 +38,7 @@ impl Health {
         let healthy = health.healthy.clone();
         let default_hook = panic::take_hook();
         panic::set_hook(Box::new(move |panic_info| {
-            error!(log, "Panic has occurred. Moving to Unhealthy");
+            error!(log, "Panic has occurred. Moving to Unhealthy: {}", panic_info);
```
markmandel, line 41: Moving this to a structured format, so it's easier to analyse (code was off the top of my head, may not be 100% correct).

```suggestion
            error!(log, "Panic has occurred. Moving to Unhealthy"; info => format!(panic_info));
```

src/metadata.rs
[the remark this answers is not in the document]
markmandel: > I think I'd rather just remove the _ => false branch and write out the rest of the variants, so we don't introduce a bug, because that is a good point.

👍🏻 I like it. Then if/when new variants show up, they have to be explictly handled.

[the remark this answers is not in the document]
iffyio: Listing them explicitly SGTM!

src/filters/matches.rs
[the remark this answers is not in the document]
iffyio: there really isn't a reasonable default behavior for this filter? In that if the filter is to do something then it needs some config, here the default behavior seems to be the same as not having the filter at all which tells me it isn't really a behavior of the filter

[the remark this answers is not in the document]
XAMPPRocky: Sorry, I meant a default for another filter. For this filter it's optional because you might want to match only in one direction `on_read` or `on_write`, or both. You're right that there's no reasonable default for having neither of those specified.

src/filters/matches.rs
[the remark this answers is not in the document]
XAMPPRocky: I think it's important that these have identical behaviour, which is why it shouldn't be duplicated. While it might be easier to ensure now that these two things do the same thing, if we add features to it, it's easy to overtime add it to one side, and not the other. This reduces some of the need for testing because testing one side's behaviour also tests the other side.

src/filters/matches.rs
XAMPPRocky: I'm going to add metrics in a follow-up PR since this is already pretty big.

XAMPPRocky:
I've updated the tests, and add one new feature to the `Matches` filter, which is the addition of `fallthrough`, which behaves like fallthrough in programming languages, except you have the options for "drop" for dropping packets if none match, "pass" for allowing passthrough, and "filter" which allows you to run a specific filter when nothing matches.

I've not added a test for the `write` side of the filter because currently the only filter we have that populates easily populates metadata is `CaptureBytes` and that does not support capturing from the write side (maybe it should?), when we have something that does, I can add it in a followup PR.

src/filters/matches.rs
XAMPPRocky: I'll rename this in a follow-up PR with the addition of metrics since this PR is already big enough.

src/filters/matches.rs
```diff
@@ -0,0 +1,439 @@
...
+    ) -> Result<Self, Self::Error> {
+        use proto::matches::directional_config::Fallthrough as ProtoFallthrough;
+
+        Ok(match branch {
+            ProtoFallthrough::Pass(_) => Self::Pass,
+            ProtoFallthrough::Drop(_) => Self::Drop,
+            ProtoFallthrough::Filter(filter) => Self::Filter {
+                filter: filter
+                    .filter
+                    .ok_or_else(|| eyre::eyre!("Unexpected missing `filter` field"))?,
```
iffyio, line 298:
```suggestion
                    .ok_or_else(|| eyre::eyre!("missing `filter` field in Fallthrough configuration"))?,
```
or something similar that mentions the context in which this happened

src/filters/matches.rs
```diff
@@ -0,0 +1,439 @@
...
+#[derive(Debug, PartialEq)]
+pub enum Fallthrough {
+    /// The packet will be passed onto the next filter.
+    Pass,
+    /// The packet will be dropped. **Default behaviour**
+    Drop,
+    /// The filter specified in `filter` will be called.
+    Filter {
+        filter: String,
+        config: Option<ConfigType>,
```
iffyio, line 274: I think this can be Option<serde_yaml::Value> rather than a full config type or? since this is only used for the yaml config? in which case it probably allows us to avoid the custom serde we have for fallthrough? (same for Branch)

src/filters/factory.rs
```diff
@@ -63,45 +63,61 @@ pub trait FilterFactory: Sync + Send {
...
-        config: Option<ConfigType<'b>>,
-    ) -> Result<ConfigType<'b>, Error> {
+    fn require_config(&self, config: Option<ConfigType>) -> Result<ConfigType, Error> {
         config.ok_or_else(|| Error::MissingConfig(self.name()))
     }
 }
 
 /// Arguments needed to create a new filter.
-pub struct CreateFilterArgs<'a> {
+pub struct CreateFilterArgs {
```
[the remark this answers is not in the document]
XAMPPRocky: I've added the label, which I believe is what @markmandel is using.

src/filters/matches.rs
XAMPPRocky: It is needed since someone can send a `Fallthrough::Filter` through xDS/protobuf, not just in the static configuration

iffyio: Ah so iiuc its not necessarily it being used through xds but rather that to use [deserialize](https://github.com/googleforgames/quilkin/pull/447/files/c5f4bbe10f1aaea625378d3585d33d61ffca6230#diff-19529b6c5a2c25b8fe5e26633ff818cc2a563656df8c69745b3429ed31d0b937R167-R169) we need to be able to convert from proto to yaml and there isn't a reasonable way to do that for the transparent filter config field? We don't have to use `deserialize` since its only a helper function, we can call the convert function ourselves and instantiate the filter i.e 
```rust
fn create_filter(&self, args: CreateFilterArgs) -> Result<FilterInstance, Error> {
  let (config_json, on_read, on_write) = match args {
    ConfigType::Static(config) => {
      // Instantiate sub filters from static config
    },
    ConfigType::Dynamic(config) => {
      // Instantiate sub filters from dynamic config
    }
  }

  let filter = Matches::new(on_read, on_write)?;
  Ok(FilterInstance::new(
    config_json,
    Box::new(filter) as Box<dyn Filter>,
  ))
}
```

src/filters/factory.rs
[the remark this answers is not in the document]
iffyio: Yeah we need the label too! What I meant was in this case the label isn't enough since the breakage isn't obvious at all from the PR - usually with a release folks will go through the breaking PRs (which is where the label comes in) and it should be clear what broke. Here that won't be possible at all without reading the code which would be harsh, so let's mention that this PR does in fact contain a breaking change since adding a Matches filter isn't breaking in itself.

src/config/config_type.rs
```diff
@@ -101,6 +101,29 @@ impl ConfigType<'_> {
...
+        S: serde::Serializer,
+    {
+        match self {
+            Self::Static(value) => value.serialize(ser),
+            Self::Dynamic(_) => Err(serde::ser::Error::custom(
+                "Protobuf configs can't be serialized.",
+            )),
+        }
+    }
+}
```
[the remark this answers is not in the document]
XAMPPRocky: These are used to deserialize a static `Matches` filter config. Protobuf's can't be used with `serde` regardless, so it makes sense IMO that it only works for static configurations.

docs/src/filters/matches.md
```diff
@@ -0,0 +1,45 @@
...
+          strategy: PREFIX
+          metadataKey: myapp.com/token
+          size: 3
+          remove: false
+    - name: quilkin.extensions.filters.matches.v1alpha1.Matches
+      config:
+          on_read:
+            metadataKey: myapp.com/token
+            branches:
+                - value: abc
```
markmandel, line 32: From digging into the code, `value` can't be a base64 byte array, correct? It has be one of the `Value` types such as String or Int? (bool?). This makes sense, since the dynamic metadata could come from anywhere, and as such, should be a `Value`.

(Just making sure my understanding is correct).

If so, we might want to make that extra clear in the documentation.

src/filters/factory.rs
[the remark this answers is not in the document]
markmandel: Yes please - add a note about this in the description, so when i write the release notes I won't forget 👍🏻

docs/src/filters/matches.md
XAMPPRocky: Yes, it's not base64, it's a string.

markmandel: _nods_ that makes total sense -- I think we should be clear in that in the docs with a comment somewhere, as I could see people assuming it was a byte array and attempting to fill it as such.

XAMPPRocky: Well I mean it should be documented but that's also valid to do that. That's why the `PartialEq` of `Value` compares the value between strings and bytes, so that you can set a string here and use it to match the bytes, since a lot of marker values in my experience are at least string friendly if not human-readable.

markmandel: 100% agree.

docs/src/filters/matches.md
```diff
@@ -0,0 +1,45 @@
...
+                    bytes: eHl6 # "xyz"
+# ";
+# let config = quilkin::config::Config::from_reader(yaml.as_bytes()).unwrap();
+# assert_eq!(config.source.get_static_filters().unwrap().len(), 1);
+# quilkin::Builder::from(std::sync::Arc::new(config)).validate().unwrap();
+```
+
+View the [Matches](./matches.md) filter documentation for more details.
+
+### [Configuration Options](../../api/quilkin/filters/matches/struct.Config.html)
```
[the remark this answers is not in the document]
XAMPPRocky: How are you generating that? Is there a specific format I should be following?

XAMPPRocky:
@markmandel Now with #478, do we want to merge this in, and I can add the configuration schema in that?

markmandel (approved):

markmandel:
> @markmandel Now with #478, do we want to merge this in, and I can add the configuration schema in that?

I like it! Also good to get this in for 0.3.0 👍🏻

PR #457, opened 2021-12-23
Completely remove `slog` and replace with `tracing`

### LINKED ISSUE #317: Move from `slog` to `tracing`

XAMPPRocky:
With the release of https://github.com/tokio-rs/console I wanted to move over to tracing sooner rather than later, so this pulls out all instances of `slog` and replaces them with `tracing`.

resolves #317

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,10 +57,6 @@ rand = "0.8.4"
 serde = { version = "1.0.130", features = ["derive", "rc"] }
 serde_json = "1.0.68"
 serde_yaml = "0.8.21"
-slog = "2.7.0"
-slog-async = "2.7.0"
-slog-json = "2.4.0"
-slog-term = "2.8.0"
 snap = "1.0.5"
 tokio = { version = "1.13.1", features = ["rt-multi-thread", "signal", "test-util", "parking_lot"] }
 tokio-stream = "0.1.8"
@@ -73,6 +69,7 @@ tryhard = "0.4.0"
 eyre = "0.6.5"
 stable-eyre = "0.2.2"
 ipnetwork = "0.18.0"
+futures = "0.3.17"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 sys-info = "0.9.0"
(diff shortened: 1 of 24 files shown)


XAMPPRocky:
@markmandel Failure seems unrelated to the PR, not sure how to fix it.

iffyio (approved):

markmandel:
> @markmandel Failure seems unrelated to the PR, not sure how to fix it.

That's a weird one. When in doubt, retry.

markmandel (approved):

markmandel:
Looks like I can't re-run a build older than three days.

Easy way to rerun a build if you don't want to go to the console is: `git commit --amend --date=now` and then push that over the top. It'll fire a whole new build.

XAMPPRocky:
@markmandel FWIW, I wrote that comment after retrying twice. I'll push an empty commit in the next couple days (on holidays)

markmandel:
> @markmandel FWIW, I wrote that comment after retrying twice. I'll push an empty commit in the next couple days (on holidays)

AAAH. I can't tell anymore, because no history 😄 Sounds like I'll need try this locally and see if I can find where things are broken in the build system 👍🏻 No worries, I'll get on it tomorrow most likely.

markmandel:
I worked it out! Since we no longer have slog-json, we don't have any MPL source code to grab, so the zip doesn't get created!

I'll come up with a fix in my draft branch, and you can include it in the PR 👍🏻

markmandel:
Got it!

You will need to update `image/archive_dependencies.sh` like so (https://github.com/markmandel/quilkin/commit/b6c755007dc890a2f93d70cb80524caa011b4694), since slog-json is no longer a dependency.

I figure it would make more sense to be included in your PR, since the removal of `slog` is tied to this PR, rather than as a separate PR.

PR #458, opened 2021-12-24
Add regex capture and rename capture_bytes to capture

XAMPPRocky:
This PR adds a regex capture strategy to the CaptureBytes filter, and renames it from `CaptureBytes` to capture, to better reflect the new `Capture` trait contract, where a capture strategy can return any valid `metadata::Value` instead of only `Bytes`. This also adds a `/is_present` extension metadata key which is appended ontop of whatever the `metadata_key`  is, this will be useful in combination with #447 to allow to match on whether any successful capture occured, rather than only matching values.


docs/src/filters/token_router.md
```diff
@@ -82,10 +82,12 @@ For example, a configuration would look like:
...
   filters:
-    - name: quilkin.extensions.filters.capture_bytes.v1alpha1.CaptureBytes # Capture and remove the authentication token
+    - name: quilkin.extensions.filters.capture.v1alpha1.Capture # Capture and remove the authentication token
       config:
-          size: 3
-          remove: true
+        strategy:
+            kind: SUFFIX
+            size: 3
+            remove: true
```
iffyio, line 90: is it possible to use the [proposed schema here](https://github.com/googleforgames/quilkin/pull/387#issuecomment-910179544) that groups the fields by strategy? this look like we'll have 1 flat object where all fields are together and we'll have to document which belongs to which strategy

src/filters/capture.rs
```diff
@@ -14,71 +14,73 @@
...
-            }
-            self.metrics.packets_dropped_total.inc();
-            return None;
-        }
-        let token = self
-            .capture
-            .capture(&mut ctx.contents, self.size, self.remove);
+        let capture = self.capture.capture(&mut ctx.contents, &self.metrics);
+        let mut capture_key = (*self.metadata_key).clone();
+        capture_key += "/is_present";
```
iffyio, line 73: the clone+concat looks expensive, we should be able to create the string once on instantiation since its not changing?

Can you elaborate what `/is_present` is for I didn't understand from the PR description and it wasnt clear looking at the code?

src/filters/capture/regex.rs
```diff
@@ -0,0 +1,33 @@
...
+            .regex
+            .find_iter(contents)
+            .map(|mat| Value::Bytes(bytes::Bytes::copy_from_slice(mat.as_bytes())))
+            .collect::<Vec<_>>();
+
+        if matches.len() > 1 {
+            Some(Value::List(matches))
+        } else {
+            matches.into_iter().next()
+        }
```
iffyio, line 25: this looks like regex can potentially capture a list or single value? if so we should document the behavior (or maybe just return a singleton list in the latter so its consistent?)

src/filters/capture.rs
XAMPPRocky: `/is_present` is more useful with the inclusion of the `Match` filter and other conditional flow filters. I wrote an [example](https://github.com/googleforgames/quilkin/discussions/459#discussion-3764877) that uses it, but in essence, it allows you to check for whether a capture succeeded without checking the specific contents.

This is more useful with the inclusion of regex because something like `\d{8}$` would capture the last eight digits, and if you wanted to check if that worked you'd need to match every possible combination of those digits, where as   with `/is_present` we can check it with a couple lines. E.g.

```yaml
  filters:
  - name: quilkin.extensions.filters.capture.v1alpha1.Capture
    config:
      strategy:
        kind: regex
        regex: \d{8}$
  - name: quilkin.extensions.filters.matches.v1alpha1.Matches
    config:
        metadataKey: quilkin.dev/capture_bytes/is_present
        fallthrough: pass
        branches:
        # If present, drop the packet.
        - value: true
          filter: quilkin.extensions.filters.drop.v1alpha1.Drop
```

src/filters/capture.rs
```diff
@@ -14,78 +14,80 @@
...
 
-impl CaptureBytes {
+impl Capture {
     fn new(config: Config, metrics: Metrics) -> Self {
-        CaptureBytes {
-            capture: config.strategy.as_capture(),
+        Self {
+            capture: config.strategy.into_capture(),
             metrics,
+            is_present_key: Arc::new(config.metadata_key.clone() + "/is_present"),
```
markmandel, line 64: Just noticed that this key is not documented, so we should add that to `capture.md`.

src/filters/capture.rs
```diff
@@ -97,8 +99,8 @@ impl FilterFactory for CaptureBytesFactory {
     fn create_filter(&self, args: CreateFilterArgs) -> Result<FilterInstance, Error> {
         let (config_json, config) = self
             .require_config(args.config)?
-            .deserialize::<Config, ProtoConfig>(self.name())?;
-        let filter = CaptureBytes::new(config, Metrics::new(&args.metrics_registry)?);
+            .deserialize::<Config, proto::Capture>(self.name())?;
```
markmandel, line 102: I like `proto::Capture` better here 👍🏻 Just reads nicer.

src/filters/capture.rs
```diff
@@ -219,39 +238,52 @@ mod tests {
...
-        assert_eq!(b"abc".to_vec(), result);
+        let result = beg.capture(&mut contents, &metrics);
+        assert_eq!(Some(Value::Bytes(b"abc".to_vec().into())), result);
         assert_eq!(b"abchello".to_vec(), contents);
 
-        let result = beg.capture(&mut contents, 3, true);
-        assert_eq!(b"abc".to_vec(), result);
+        beg.remove = true;
+
+        let result = beg.capture(&mut contents, &metrics);
```
markmandel, line 285: I'd like to see a `regex_capture()` test like `end_capture()` and `beginning_capture()`, so we can have a faster interation loop without have to rely on an e2e test to determine if the regex capture is working as expected.

tests/capture.rs
```diff
@@ -0,0 +1,88 @@
...
+/// This test covers both token_router and capture filters,
+/// since they work in concert together.
+#[tokio::test]
+async fn token_router() {
+    let mut t = TestHelper::default();
+    let echo = t.run_echo_server().await;
+
+    let capture_yaml = r#"
+regex:
+    pattern: .{3}$
```
markmandel, line 38: I actually didn't realise that you could do neat stuff like this! 😮 (I always relate regex with doing character matching) -- this is very, very cool, I hadn't thought through the implications of this addition.

I think it would be awesome to show in capture.md under "Configuration Examples" that you could do first 3 values capturing using `prefix`, but also do the same thing using `regex` -- two different ways to do the same thing, just to give people ideas of what is possible.

Very nice.

docs/src/filters/capture.md
```diff
@@ -8,9 +8,25 @@ This is often used as a way of retrieving authentication tokens from a packet, a
...
+#### Suffix
+Captures bytes from the end of the packet.
+
+#### Prefix
+Captures bytes from the start of the packet.
+
+#### Regex
+Captures bytes using a regular expression. Unlike other capture strategies,
+the regular expression can return one or many values if there are
+multiple matches.
```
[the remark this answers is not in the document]
XAMPPRocky: > Also wondering if we need to be explicit that a Regex filter that returns more than one value will be dropped by the Token Router.

Well that's why it currently returns a single value, so that it continues to work with token_router. I wonder if we enforce regex to always be a list, that we should also change token_router to check if anything in the list matches?

src/filters/capture.rs
```diff
@@ -14,78 +14,80 @@
...
         ctx.metadata
-            .insert(self.metadata_key.clone(), Value::Bytes(token.into()));
+            .insert(self.is_present_key.clone(), Value::Bool(capture.is_some()));
 
-        Some(ctx.into())
+        if let Some(value) = capture {
+            ctx.metadata.insert(self.metadata_key.clone(), value);
+            Some(ctx.into())
+        } else {
+            None
```
markmandel, line 81: I just noticed this change in behaviour - this would mean that if there is no match, we drop the whole packet.

I don't think we want to change the behaiviour here. The job of this filter is purely to capture data - it's not to make decisions about it.

🤔 Unless we want to add an option to "drop on failing to match" - which I think would be a nice addition (we could do this in a subsequent PR/track in an issue)?

docs/src/filters/capture.md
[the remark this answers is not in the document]
markmandel: Except `Regex` can return a `Value::List` or a single value?

https://github.com/googleforgames/quilkin/blob/2828b8678d6ab3440ddc61eac27e6264f5528bed/src/filters/capture/regex.rs#L20-L26

Or am I reading `matches.into_iter().next()` wrong?

> that we should also change token_router to check if anything in the list matches?

There's probably an interesting discussion of having an option on `token_router` to allow matching against lists of values - shall we create an issue for it? (seems to me like it should be opt in - optimise on the extra security/validation first).

src/filters/capture.rs
XAMPPRocky: This is not a change in behaviour. See here https://github.com/googleforgames/quilkin/blob/a439516e16199857bda8a2d2a8c3c1bfcf47f37a/src/filters/capture_bytes.rs#L64-L73 the packets are dropped when suffix or prefix don't have enough bytes.

Now whether this behaviour is desirable is a different story. I agree that this filter should probably only be able capturing, and make it configurable as to whether it drops on failing captures. I can add that change to this PR if you agree.

markmandel: ooooh yeah, I see your point. 👍🏻 Oh that's load of fun. Prefix and Suffix are checking for a match just by the nature of checking if there is enough data to actually grab. Excellent point, I didn't think of it that way.

> Now whether this behaviour is desirable is a different story. I agree that this filter should probably only be capturing, and make it configurable as to whether it drops on failing captures. I can add that change to this PR if you agree.

I would absolutely be happy for this addition. I'll leave it up to you if you want to do it in this PR, or we can create an issue and track it as a separate unit of work.

docs/src/filters/capture.md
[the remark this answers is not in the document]
XAMPPRocky: > Except Regex can return a Value::List or a single value?

Yes, currently but @iffyio mentioned the alternative being always have it be a list even when it's a single value. 

> (seems to me like it should be opt in - optimise on the extra security/validation first)

This is kinda why I think I'd lean more towards returning a single value over a list with a single value, so that it can be compatible with token router, since most of the you probably only want a single capture group anyway.

I'll update token router's documentation to say that it requires a single value.

src/filters/capture.rs
XAMPPRocky: This is now documented in the config struct, once #478 is merged it will be documented in `capture.md`.

markmandel (approved):

PR #462, opened 2022-01-10
xds: delete snapshot for disconnected proxies

iffyio:
When proxies disconnect this clears their snapshot from
the cache. It periodically checks for snapshots that have
open watches (i.e has a proxy waiting for an update) and
if a snapshot seems idle after a while (the proxy isn't likely
to reconnect) it removes it from the cache.


markmandel (approved):
LGTM. Nice stuff!

XAMPPRocky (approved):

PR #469, opened 2022-01-13
management-server: update metrics and doc

iffyio:
- Adds some new metrics to the management server
- Add metrics to the documentation
- Update documentation to show the kubernetes docs first
  since that's the primary binary. Followed by the file
  implementation.

### DIFF
--- a/xds/README.md
+++ b/xds/README.md
@@ -1,4 +1,4 @@
-#### XDS Management Server
+# XDS Management Server
 
 This project contains an [XDS] management server implementation that configures
 Quilkin proxies.
@@ -7,11 +7,21 @@ upstream endpoints and filter configuration from some external source, and then 
 that information to generate cluster and filter updates that is pushed to any connected
 Quilkin proxy.
 
+## Opinionated Agones Implementation
+
 The project has a binary **cmd/controller.go** with a server implementation that runs in [Kubernetes].
    
 1. Cluster information is retrieved from [Agones] - the server watches for `Allocated`
-   [Agones GameServers] and exposes their IP address and Port as upstream endpoints to
+   [Agones GameServers] and exposes their IP address and Port as [upstream endpoints][upstream-endpoint] to
    any connected Quilkin proxies.
+   The set of tokens for the associated endpoint can be set by adding a comma separated standard base64 encoded strings.
+   This must be added under an annotation `quilkin.dev/tokens` in the [GameServer][Agones GameServers]'s spec.
+   For example:
+   ```yaml
+   annotations:
+     Sets two tokens for the corresponding endpoint with values 1x7ijy6 and 8gj3v2i respectively.
+     quilkin.dev/tokens: MXg3aWp5Ng==,OGdqM3YyaQ==
+   ```
 
    > Since an Agones GameServer can have multiple ports exposed, if multiple ports are in
    > use, the server looks for the port named `default` and picks that as the endpoint's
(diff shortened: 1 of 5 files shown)


XAMPPRocky (approved):

xds/README.md
```diff
@@ -7,115 +7,153 @@ upstream endpoints and filter configuration from some external source, and then
...
-   - `/metrics`: Exposes Prometheus metrics.
+1. Cluster information is retrieved from [Agones] - the server watches for `Allocated`
+   [Agones GameServers] and exposes their IP address and Port as upstream endpoints to
+   any connected Quilkin proxies.
+
+   > Since an Agones GameServer can have multiple ports exposed, if multiple ports are in
+   > use, the server looks for the port named `default` and picks that as the endpoint's
+   > port (otherwise it picks the first port in the port list).
+
+1. Filter chain is configurable on a per-proxy basis. By default an empty filter chain is
```
markmandel, line 20: Since we're in here, shall we also document the usage of the annotation of `quilkin.dev/tokens` ?

https://github.com/googleforgames/quilkin/blob/b32e48af083795bb3e02c8915b34e3d8c5949623/xds/pkg/cluster/agones/agones.go#L138

xds/README.md
```diff
@@ -7,115 +7,153 @@ upstream endpoints and filter configuration from some external source, and then
...
-     typed_config:
-       '@type': quilkin.extensions.filters.debug.v1alpha1.Debug
-       id: hello
-   ```
-
-   > The file.go binary is primarily an example and mostly suitable for demo purposes.
-   > As a result, some configuration options and features might be missing.
-
-1. **cmd/controller.go**: A server implementation that runs in [Kubernetes].
+The project has a binary **cmd/controller.go** with a server implementation that runs in [Kubernetes].
```
markmandel, line 10: Want to add some big headers, finding it hard to separate the sections when looking over all the text (also will give us a nice TOC).

```suggestion
## Opinionated Agones Implementation

The project has a binary **cmd/controller.go** with a server implementation that runs in [Kubernetes].
```

xds/README.md
```diff
@@ -7,115 +7,153 @@ upstream endpoints and filter configuration from some external source, and then
...
+   to a proxy update - a proxy only gets the latest snapshot so it might miss intermediate
+   snapshots if it lags behind.
+- `quilkin_management_server_snapshots_cache_size` (Gauge)
+
+   The current number of snapshots in the in-memory snapshot cache. This corresponds 1-1 to
+   proxies that connect to the server. However the number may be slightly higher than the number
+   of connected proxies since snapshots for disconnected proxies are only periodically cleared
+   from the cache.
+
+##### File Server
```
markmandel, line 90: To match size and semantic level as the other section.

```suggestion
## Example: File Server Implementation
```

(Not sure if I'm sold on the heading content change, but something like this to make it clear on scan through that it is an example, vs the Agones one which is aimed at being used in production)

xds/pkg/cluster/cluster.go
```diff
@@ -18,6 +18,20 @@ package cluster
 
 import (
 	"context"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"quilkin.dev/xds-management-server/pkg/metrics"
+
```
markmandel, line 24: Super tiny nit -- I'd _love_ to get rid of spaces between the non-go native imports.

Non blocking though, it's my personal irk. So if I get really annoyed I'll go in at a later date and do it myself 😄

xds/README.md
iffyio: Added!

xds/README.md
iffyio: Added!

markmandel (approved):

markmandel:
Failure not related to the PR - looks like the Google Snappy package removed their github pages. Filing a PR shortly.

PR #472, opened 2022-01-24
Add Pass & Drop filters, Refactor `Match` config

### LINKED ISSUE #454: Rename `Matches` to `Match`.
`Match` could be a better name for the filter? think I remember a while ago @markmandel and I had discussed having a convention for filters and to use verbs like compress/encrypt where possible (not sure how well we've kept to it ^^)

_Originally posted by @iffyio in https://github.com/googleforgames/quilkin/pull/447#discussion_r759947759_

XAMPPRocky:
This adds simple "pass" and "drop" filters, that either always allow a packet to pass, or always drop a packet. These filters aren't useful on their own, but become useful when used in combination with conditional flow filters like `Match` (#447).

closes #454


markmandel:
Only question I have: Did we want to include the documentation with these Filters in this PR, or should it come as a separate? (if separate, let's track that in an issue, so we don't forget).

Otherwise, LGTM!

XAMPPRocky:
I've added documentation, and I've also refactored a small bit of `Matches`, I've now renamed it to `Match`, and I've updated it from having special cased `Pass` and `Drop` branches in its configuration, if you want to drop or pass a packet, you should use the `Pass` and `Drop` filters. This simplifies some of the configuration.

@markmandel Would nice to get this in the next release, so that it's not technically a breaking change.

src/filters.rs
```diff
@@ -30,11 +30,13 @@ pub mod capture;
 pub mod compress;
 pub mod concatenate_bytes;
 pub mod debug;
+pub mod drop;
 pub mod firewall;
 pub mod load_balancer;
 pub mod local_rate_limit;
-pub mod matches;
+pub mod r#match;
```
markmandel, line 37: _(non blocking):_ I assume the `r#` is because `match` is a keyword?

markmandel (approved):

PR #473, opened 2022-01-24
Update Rust to 1.58.1

XAMPPRocky:
This PR updates Quilkin to use 1.58.1, moving some format strings to use the new syntax (`format!("{var}")` instead of `format!("{}", var)`), and removes some more unneeded string parsing for IP addresses that were found while moving to the new syntax.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -470,7 +470,7 @@ dynamic:
         for config in configs {
             let result = Config::from_reader(config.as_bytes());
             let error = result.unwrap_err();
-            assert!(format!("{:?}", error).contains("unknown field"));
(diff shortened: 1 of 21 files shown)


src/config.rs
```diff
@@ -470,7 +470,7 @@ dynamic:
         for config in configs {
             let result = Config::from_reader(config.as_bytes());
             let error = result.unwrap_err();
-            assert!(format!("{:?}", error).contains("unknown field"));
+            assert!(format!("{error:?}").contains("unknown field"));
```
markmandel, line 473: Oh that is super nice!

markmandel (approved):

PR #478, opened 2022-01-31
Add auto-generation of JSON schema.

XAMPPRocky:
This commit adds support for Quilkin to generate JSON schema for any of the included filters, and adds some glue to make it so that these schemas are automatically included in the mdBook documentation.

Additionally to work with mdbook, the `-q/--quiet` flag was added to run Quilkin in silent mode.

There's still some improvements to be made for some of the special items, but this is already much better IMO than having to maintain these configurations by hand.

### Breaking Changes
(these are mostly changes to better accept `config`  now with the change to a derive macro for the CLI, which makes it easier to accept and validate types like `PathBuf`)

- `runner::run_with_config` has been removed and `runner::run` now accepts a `Config` argument.
- `Config::find` has been removed and functionality has been moved into the binary directly. This is to make it easier to move the environment handling to `clap`.
- `--config` now comes before the command.
- Added new required method `config_schema` to `FilterFactory`.


markmandel:
```
error[E0046]: not all trait items implemented, missing: `config_schema`
   --> src/filters/matches.rs:161:1
    |
161 | impl FilterFactory for MatchesFactory {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `config_schema` in implementation
    |
   ::: src/filters/factory.rs:62:5
    |
62  |     fn config_schema(&self) -> schemars::schema::RootSchema;
    |     -------------------------------------------------------- `config_schema` from trait

For more information about this error, try `rustc --explain E0046`.
error: could not compile `quilkin` due to previous error
```

Ooops, new filter.

docs/src/quickstart-agones-xonotic.md
```diff
@@ -112,7 +112,7 @@ replace the `${GAMESERVER_IP}` and `${GAMESERVER_PORT}` values in your copy of `
 Run this configuration locally as:
 
 ```shell
-quilkin run -c ./client-compress.yaml`
+quilkin -c ./client-compress.yaml run`
```
markmandel, line 115: The reordering of this feels counter intuitive to me (thinking `docker run` vs `docker build` determining which args are able to be passed in).

I assume this is something we can't get away from?

@luna-duclos what are your thoughts?

luna-duclos: I am quite familiar to this in various tools, so its fine by me at least to have some top level configs

markmandel: Fair enough - I'm out voted 😄 I rescind my objections!

markmandel (approved):

XAMPPRocky:
@markmandel Can't figure out the CI failure here. That (`Makefile:120`) is the `archive_dependencies.sh` script FWIW.

```
Step 3/13 : WORKDIR /
make: *** [Makefile:120: build-image] Terminated
```

markmandel:
> 

Yeah, that's a weird one. Lemme take a peek, see what's up.

markmandel:
Ah found it - if we look at the whole log, we see as:

```
TIMEOUT
ERROR: context deadline exceeded
```

Which is hidden at the step level.

Cloud Build is set to 30m build time (at some point we should dig into why each step keeps re-downloading and compiling - thought that volume was shared between builds).

I'll send a PR to bump it to 2h.

Issue #410, opened 2021-10-04
Replace listen distributor task with multithreaded `SO_REUSEPORT` task.


XAMPPRocky:
Currently all of our traffic goes through a distributor task which distributes all messages in the UDP buffer amongst all workers. Under heavy workloads this is likely to be one of the main bottlenecks in the program. While reading [this blog post](https://medium.com/@fujita.tomonori/scalable-server-design-in-rust-with-tokio-4c81a5f350a3) it introduced me to the `SO_REUSEPORT`, which is designed specifically to address this bottleneck in network applications. 

Using `SO_REUSEPORT` and `SO_REUSEADDRESS` we can eliminate the distributor task entirely, and have each worker entirely responsible for their socket, this has the potential to have serious performance improvements as seen in the blog post where the reused port server continues to scale linearly past 300,000 as the number clients increased, while the listen distributor server struggles reach that.

iffyio:
This sounds worth exploring indeed! I guess the idea would be to give each worker its own socket with this enabled?

XAMPPRocky:
Yeah, the [exact code they used is here](https://github.com/fujita/tokio-reuseport/blob/master/greeter-reuseport/src/main.rs). I included inline the socket stuff for convenience. One thing to figure out would be if there's an equivalent option for Windows. MacOS does have it, though it's behaviour is slightly different, in that Linux behaves specially it seems mostly in relation its TCP implementation though, which isn't relevant for us.

- https://programmer.group/the-evolution-of-reuseport-in-the-linux-kernel.html
- https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/setsockopt.2.html

```rust
    let sock = socket2::Socket::new(
        match addr {
            SocketAddr::V4(_) => socket2::Domain::IPV4,
            SocketAddr::V6(_) => socket2::Domain::IPV6,
        },
        socket2::Type::STREAM,
        None,
    )
    .unwrap();


    sock.set_reuse_address(true).unwrap();
    sock.set_reuse_port(true).unwrap();
    sock.set_nonblocking(true).unwrap();
    sock.bind(&addr.into()).unwrap();
    sock.listen(8192).unwrap();


    let incoming =
        tokio_stream::wrappers::TcpListenerStream::new(TcpListener::from_std(sock.into()).unwrap());
```

markmandel:
Very cool! Excited to see the results!

[...]

PR #479, opened 2022-02-01
Benchmark comparing read and write throughput

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Wanted to be able to highlight if we had bottlenecks in performance on read vs write operations on the proxy.

This adds an extra benchmark to throughput.rs called "readwrite" and follows a similar pattern as the overall throughput benchmark, with both direct and proxies traffic utilised as extra comparison values.

Work on #410

This is my first benchmark in Rust - feedback appreciated.

Would be interested if we think the new benchmark should be in a `readwrite.rs` ? Since it reused some machinery and was still essentially about throughput I figured it would make sense to stay in the same file, but wasn't 100% on it.

### DIFF
--- a/benches/throughput.rs
+++ b/benches/throughput.rs
@@ -1,7 +1,12 @@
-use std::net::UdpSocket;
+use std::net::{Ipv4Addr, SocketAddr, UdpSocket};
+use std::sync::{atomic, mpsc, Arc};
+use std::thread::sleep;
+use std::time;
 
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use once_cell::sync::Lazy;
+
+use quilkin::config::Admin;
 
 const MESSAGE_SIZE: usize = 0xffff;
 const DEFAULT_MESSAGE: [u8; 0xffff] = [0xff; 0xffff];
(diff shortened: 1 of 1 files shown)


markmandel:
Noting some interesting results from the benchmark. I've attached one of the `readwrite` runs as a PDF.
[readwrite Summary - Criterion.rs.pdf](https://github.com/googleforgames/quilkin/files/7982420/readwrite.Summary.-.Criterion.rs.pdf)

On average, throughput looks about the same:

image

Were I see some differences (and it's not 100% consistent, so we should triple check the results, and I should run it without other processes on the machine, for sure - but it does align with what I've seen previously on Xonotic+grafana graphs), is that the outliers on `read` tend to be much larger than that of write.

I'm cherry picking here, but to illustrate the general trend I'm seeing here is `read/508 bytes`:
image

And the corresponding `write/508 bytes`:
image

This can fluctuate between runs though, so I should run this in a cleaner environment and do multiple runs before definitely drawing conclusions.

The other thought I had too was it might be worth doing benchmarks with multiple udp clients (maybe more than current cpu on the machine?) and multiple servers, and look at the same things.

Thoughts?

XAMPPRocky (approved):

XAMPPRocky:
> Would be interested if we think the new benchmark should be in a readwrite.rs ?

I think it's fine for now, and maybe split it up once we have even more benchmarks.

> The other thought I had too was it might be worth doing benchmarks with multiple udp clients (maybe more than current cpu on the machine?) and multiple servers, and look at the same things.

I think this would be a good variable to benchmark, another variable worth benchmarking in my opinion is the amount of filters in a given configuration. For example, how does it look with 5 filters vs 10? We'd probably just want to use the `Pass` filters for that because we really just want to benchmark the internal machinary rather the performance of an individual filter (though I could also adding benchmarks for specific filters as well [routing, compression, etc]).

PR #485, opened 2022-02-07
Add basic alpha version of the UE4 plugin

XAMPPRocky:
Adds the basic initial version of the UE4 client, right now it just attaches a routing token and version to the end of the packet.


markmandel (commented):
Apache licences on everything please.

I know nadda about UE4. @luna-duclos would yourself, or someone else at Embarl be good to review this? (unless it already has?)

The only other thought I have, is that we should reference this in our [book](https://googleforgames.github.io/quilkin/main/book/) somewhere. Maybe a new "SDKs" section?

markmandel:
Another thought on having someone with UE4 knowledge give a LGTM - just to double check everything is included, and where it should be. Not something I can do.

sdks/ue4/README.md
```diff
@@ -0,0 +1,9 @@
+# Quilkin Unreal Engine Plugin
```
markmandel, line 1: I would recommend moving this documentation into the [Guide](https://googleforgames.github.io/quilkin/main/book/), as well as providing some description of how to integrate. (Does it need to be put somewhere specific? How do I integrate it? Is there a UI I should use? etc).

bombsimon (approved):
I can't say I'm an UE4 expert but this code looks good to me, properly structured as UE4 plugin and just wraps the `FUniqueSocket` for all calls.

LGTM!

markmandel:
> I can't say I'm an UE4 expert but this code looks good to me, properly structured as UE4 plugin and just wraps the `FUniqueSocket` for all calls.
> 
> LGTM!

Awesome. Then the only thing blocking for me is getting some documentation into the Guide!

docs/src/sdks.md
```diff
@@ -0,0 +1,3 @@
+# SDKs
+
+You will learn here about Quilkin SDKs that can be used as features in your UDP proxy.
```
XAMPPRocky, line 3:
```suggestion
You will learn here about Quilkin SDKs that can be used as clients for interacting with your UDP proxy.
```

PR #489, opened 2022-02-21
Refactor FilterRegistry into private static

XAMPPRocky:
This PR refactors the state of `FilterRegistry` to being internal to `FilterRegistry` and is now a global append-only and always safe to load (e.g. not blocking) registry of filters. This simplifies some of the state in our code, removing the need for a `FilterRegistry` to be passed down at every level, you can get access to new instances with `FilterRegistry::get` which is available anywhere.


markmandel (approved):
Super nice change! 👍🏻

PR #491, opened 2022-02-22
Refactor FilterManager into SharedFilterChain

XAMPPRocky:
This PR refactors `FilterManager` into `SharedFilterChain`, which instead of using message passing, uses `ArcSwap` to provide a updatable handle to a filter chain. This simplifies a lot of the communication infrastructure as we can just pass `SharedFilterChain` directly around instead, update it as needed, and it will be updated on subsequent loads. This also makes it easy to inspect the current state of the filter chain.


src/filters/chain.rs
```diff
@@ -121,15 +125,30 @@ impl FilterChain {
...
+
+    /// Returns an iterator over the current filters' configs.
+    pub(crate) fn get_configs(&self) -> impl Iterator<Item = (&str, Arc<serde_json::Value>)> {
+        self.filters
+            .iter()
+            .map(|(config_json, config)| (config_json.as_str(), config.config.clone()))
+    }
+}
+
+impl TryFrom<&[FilterConfig]> for FilterChain {
```
markmandel, line 140: Nice refactor 👍🏻

src/filters/chain/shared.rs
```diff
@@ -0,0 +1,138 @@
+/*
+ * Copyright 2021 Google LLC
```
markmandel, line 2:
```suggestion
 * Copyright 2022 Google LLC
```

markmandel (approved):
One minor nit on the copyright, but approved so you can merge whenever it's fixed.

PR #493, opened 2022-02-23
Refactor ClusterManager into SharedCluster

XAMPPRocky:
Similar to #490 except that `SharedCluster` isn't static. This PR refactors `SharedCluster` to use `ArcSwap` as its backing storage, and removes the need to use message passing for updates.


src/endpoint.rs
```diff
@@ -361,27 +381,27 @@ mod tests {
 
     #[test]
     fn new_endpoints() {
-        assert!(Endpoints::new(vec![]).is_none());
-        assert!(Endpoints::new(vec![ep(1)]).is_some());
+        assert!(Endpoints::try_new(vec![]).is_none());
+        assert!(Endpoints::try_new(vec![ep(1)]).is_some());
```
bombsimon, line 385: Since this PR also introduces the `new` constructor that could panic, would it make sense to add an explicit test for it so it's clear it's intentional and nothing temporary or bad practice? Or is it just too verbose and obvious?

```suggestion
        assert!(Endpoints::try_new(vec![]).is_none());
        assert!(Endpoints::try_new(vec![ep(1)]).is_some());
        assert_eq!(Endpoints::new(vec![ep(1)]).0.len(), 1);
    }
    
    #[test]
    #[should_panic]
    fn new_endpoints_panic() {
        Endpoints::new(vec![]);
```

src/filters/chain.rs
```diff
@@ -121,15 +125,30 @@ impl FilterChain {
...
+
+    /// Returns an iterator over the current filters' configs.
+    pub(crate) fn get_configs(&self) -> impl Iterator<Item = (&str, Arc<serde_json::Value>)> {
+        self.filters
+            .iter()
+            .map(|(config_json, config)| (config_json.as_str(), config.config.clone()))
+    }
+}
+
+impl TryFrom<&[FilterConfig]> for FilterChain {
```
bombsimon, line 140: 👍

bombsimon (approved):
LGTM!

src/endpoint.rs
XAMPPRocky: I've now added documentation about the panic, and a test for the behaviour.

src/cluster/shared.rs
```diff
@@ -0,0 +1,270 @@
+/*
+ * Copyright 2020 Google LLC
```
markmandel, line 2:
```suggestion
 * Copyright 2022 Google LLC
```

markmandel (approved):
Found one tiny nit in the year on the Apache licence, but other than that LGTM 👍🏻 

Merge when ready.

PR #494, opened 2022-02-23
Quilkin Socket API

XAMPPRocky:


XAMPPRocky:
Going to close this, as a lot of this was implemented in #552, and it will be easier to write the other parts from scratch once that's merged.

PR #506, opened 2022-03-24
Rust xDS server

XAMPPRocky:
The following is a minimal implementation of a State of The World (SoTW) Aggegated Discovery Service (aDS) server. This allows us to re-use types from Quilkin inside our xDS server without duplication, easily use our actual xDS server in tests, and simplify our deployments by allowing running the xDS server from the same binary as the client.


markmandel:
Curious - you thinking to replace the Go implementation, or is this just an extra way of doing a similar thing?

XAMPPRocky:
> you thinking to replace the Go implementation, or is this just an extra way of doing a similar thing?

The goal is to replace the Go implementation for a few reasons.

- **Consistency & code re-use** — Most of the Go xDS server is currently recreating a lot of the concepts that exist in Quilkin, having the management server in Rust allows us to reuse a lot of functionality without any changes. This is particularly important for configuration consistency, for example as it stands the configuration for filters in the management server is slightly different from Quilkin because it relies entirely on the Protobuf definitions.

  Where as in Rust we can re-use the exact same configuration, this will also simplify a lot of user stories, as this will mean that users who have a static configuration, and want to move to using agones, can move that configuration into a configmap and have it work out of the box, rather than needing to migrate the configuration. We would also then need to have documentation teaching users how to write their configuration for xDS, or we could invest a bunch of time making it so that the configuration matches between the management server and Quilkin, but that's a lot of work and maintenance just to be consistent, our time would be better spend building new features IMO.

- **Testing** Writing integration tests across different languages (particularly Go & Rust) is hard, because you're getting into distributed systems level behaviour of spawning two different servers and then needing a way to introspect their state at a given moment. We'd be able to test a lot more specific behaviour with a lot less overhead to writing tests if both of these services were in Rust, as we could easily run the request-response lifecycle procedurally, and we'd have access to their internals so we wouldn't need to add extra work to just to inspect state. 

- **Better Integration** I also think that having it in Quilkin provides a much better integration story for both us and users. Right now I've written the code so that integrates into the existing quilkin binary, and this provides a much better user experience, as now spawning a management server in agones is just `quilkin manage agones`, and spawning a management server that is watching a file and pushing updates is `quilkin manage file ./path/to/config.yaml`. This also simplifies a good bit of DevOps as it allows us to reuse one image deployment for both quilkin and the management server, so we can know that if you're using a given image version the management server and quilkin are guaranteed to be compatible, which today they are not.

  Again, we could invest a bunch of time and code into building a bazel/make build system that ensures that those things remain consistent, but that is investing a lot of work and maintenence just to maintain consistency, where I think that time would be better spent working on new features and improving Quilkin. This code already does most of what the current management server does and is a tenth of the size.

markmandel:
This 100% sounds awesome. I believe the original intent to utilise Go was to take advantage of https://github.com/envoyproxy/go-control-plane - but if that's less useful here (you're digging far deeper into things than I have), that seems very reasonable.

Also - from first pass, the Rust K8s client looks really nice.

I think the whole idea of `quilkin manage agones` is straight awesome (this does assume they are in the same cluster?) - my only thought - I think it would be worth doing some design work in a ticket up front here, just so we can all reach consensus on the command surface and generally how it will work, and hopefully make sure no wasteful development work occurs in a direction we ultimately decide isn't going to work (or are we just mimicking what exists in Go right now?)

`quilkin manage file ./path/to/config.yaml` is brilliant as well. 🔥 

I'm all in. Love it.

XAMPPRocky:
>  I think it would be worth doing some design work in a ticket up front here, just so we can all reach consensus on the command surface and generally how it will work, and hopefully make sure no wasteful development work occurs in a direction we ultimately decide isn't going to work (or are we just mimicking what exists in Go right now?)

Yes, the current goal is only to replicate the existing behaviour of the Go xDS server so there should be nothing surprising in this implementation. Definitely open to more design discussions as we discuss more advanced features, and move past what is currently implemented in the Go xDS server, which is a very straightforward State Of The World aggregated xDS.

XAMPPRocky:
This code is now ready for review. It implements the initial version of xDS with support for Agones to be able to run Quilkin as a cluster. There is still some more work to be done, but to not make this PR too big, we're going to do them as follow-ups, I'll mention some of the work below, which I'll turn into issues to track their progess.

- **File provider** — A simpler provider that allows you manage Quilkin clusters from a single configuration file. (currently being worked on by @rezvaneh)
- **Metrics & Admin** — Currently this management server doesn't include metrics, ideally we'd like to have some, but it's not essential to running it, relatedly, there's no admin server because it wouldn't be doing very much without the metrics.
- **Resource specific responses** Currently the management server assumes you're always asking for wildcard or "all" resources (this is also the case in the current go management server and the Quilkin xDS client.), but we should probably all you to have multiple clusters and request resources for that specific cluster or listener.

---


@markmandel The CI is currently failing because it cannot find OpenSSL, but AFAICT the build-image does install `libssl-dev`, so I'm not sure why it's not finding it.

```
Step #7 - "build":   run pkg_config fail: "`\"pkg-config\" \"--libs\" \"--cflags\" \"openssl\"` did not exit successfully: exit status: 1\nerror: could not find system library 'openssl' required by the 'openssl-sys' crate\n\n--- stderr\nPackage openssl was not found in the pkg-config search path.\nPerhaps you should add the directory containing `openssl.pc'\nto the PKG_CONFIG_PATH environment variable\nNo package 'openssl' found\n"
Step #7 - "build": 
Step #7 - "build":   --- stderr
Step #7 - "build":   thread 'main' panicked at '
Step #7 - "build": 
Step #7 - "build":   Could not find directory of OpenSSL installation, and this `-sys` crate cannot
Step #7 - "build":   proceed without this knowledge. If OpenSSL is installed and this crate had
Step #7 - "build":   trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the
Step #7 - "build":   compilation process.
Step #7 - "build": 
Step #7 - "build":   Make sure you also have the development packages of openssl installed.
Step #7 - "build":   For example, `libssl-dev` on Ubuntu or `openssl-devel` on Fedora.
Step #7 - "build": 
Step #7 - "build":   If you're in a situation where you think the directory *should* be found
Step #7 - "build":   automatically, please open a bug at https://github.com/sfackler/rust-openssl
Step #7 - "build":   and include information about your system as well as this message.
Step #7 - "build": 
Step #7 - "build":   $HOST = x86_64-unknown-linux-gnu
Step #7 - "build":   $TARGET = x86_64-unknown-linux-gnu
Step #7 - "build":   openssl-sys = 0.9.72
Step #7 - "build": 
Step #7 - "build":   ', /cargo/registry/src/github.com-1ecc6299db9ec823/openssl-sys-0.9.72/build/find_normal.rs:180:5
Step #7 - "build":   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Step #7 - "build": warning: build failed, waiting for other jobs to finish...
Step #7 - "build": error: build failed
Step #7 - "build": make: *** [Makefile:94: build-linux-binary] Error 101
Finished Step #7 - "build"
ERROR
```

markmandel:
The CI failure is a hilarious amount of fun.

TL:DR - cross deliberately removed OpenSSL:
* https://github.com/cross-rs/cross/issues/229

You can reproduce by using `make build-linux-binary`.

The mac build with `rust-linux-darwin-builder` also fails, but I think that might just be pointing to the wrong OpenSSL library.

I'll dig in, see what I can work out.

markmandel:
Just testing a solution with the `openssl = { version = "0.10", features = ["vendored"] }` feature, and so far, it's working for the linux build under cross. Attempting the rest now.

markmandel:
Okay, that worked with linux and windows - gotta work out what's going on with macos, that's still not working.

markmandel:
TIme to for me to finish up for the evening, but I _think_ I've got all the platforms compiling 😄 . I'll do some extra testing tomorrow and tidy it up.

If you want to take it for a spin, you can see it all here: https://github.com/markmandel/quilkin/tree/pr/rf/xds-server-rust

src/xds/provider/agones.rs
```diff
@@ -0,0 +1,398 @@
...
+}
+
+#[derive(kube::CustomResource, Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[kube(
+    group = "agones.dev",
+    version = "v1",
+    kind = "GameServer",
+    status = "GameServerStatus",
+    namespaced
+)]
```
bombsimon, line 231: A lot of these types and information is also implemented in [`rymder`](https://github.com/EmbarkStudios/rymder/blob/50ee2876d71f4dd4a82245e54023edc5b46fa5be/src/sdk/gameserver.rs). Is it a bad idea to consider patching `rymder` to make it more mature if needed and then depend on that crate here instead of duplicating everything? Or if not maybe the [agones Rust SDK](https://github.com/googleforgames/agones/tree/main/sdks/rust)?

bombsimon (commented):
LGTM! Mostly curious of how we want to work with the Agones SDK and common types. Would be nice to not have to duplicate if we don't have to, even if it's not that much code.

src/xds/provider/agones.rs
XAMPPRocky: I think that would be up to @markmandel. My understanding from going through the Rust SDK and the other language SDKs. Is that some of those types are available, but that is for interacting with Agones through its GRPC API, rather than through k8s which is what we're doing.

I've no problem pushing those types elsewhere, but we'd want to ensure that they're compatible with `kube` which is also why I didn't use `rymder`/Rust SDK.

src/xds.rs
```diff
@@ -107,15 +107,46 @@ mod google {
...
+pub const EXTENSION_CONFIG_TYPE: &str =
+    "type.googleapis.com/envoy.config.core.v3.TypedExtensionConfig";
+pub const LISTENER_TYPE: &str = "type.googleapis.com/envoy.config.listener.v3.Listener";
+pub const ROUTE_TYPE: &str = "type.googleapis.com/envoy.config.route.v3.RouteConfiguration";
+pub const RUNTIME_TYPE: &str = "type.googleapis.com/envoy.service.runtime.v3.Runtime";
+pub const SCOPED_ROUTE_TYPE: &str =
+    "type.googleapis.com/envoy.config.route.v3.ScopedRouteConfiguration";
+pub const SECRET_TYPE: &str =
+    "type.googleapis.com/envoy.extensions.transport_sockets.tls.v3.Secret";
+pub const VIRTUAL_HOST_TYPE: &str = "type.googleapis.com/envoy.config.route.v3.VirtualHost";
```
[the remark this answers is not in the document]
XAMPPRocky: It depends on the case, I think with this one adding `skip` wouldn't add much here, because it would still be having two lines for each one. If we want to make this clearer, I could add this macro here instead.

```rust
macro_rules! type_urls {
    ($($base_url:literal : {$($const_name:ident = $type_url:literal),+ $(,)?})+) => {
        $(
            $(
                pub const $const_name : &str = concat!($base_url, "/", $type_url);
            )+
        )+
    }
}

type_urls! {
    "type.googleapis.com": {
        CLUSTER_TYPE = "envoy.config.cluster.v3.Cluster",
        ENDPOINT_TYPE = "envoy.config.endpoint.v3.ClusterLoadAssignment",
        EXTENSION_CONFIG_TYPE = "envoy.config.core.v3.TypedExtensionConfig",
        LISTENER_TYPE = "envoy.config.listener.v3.Listener",
        ROUTE_TYPE = "envoy.config.route.v3.RouteConfiguration",
        RUNTIME_TYPE = "envoy.service.runtime.v3.Runtime",
        SCOPED_ROUTE_TYPE = "envoy.config.route.v3.ScopedRouteConfiguration",
        SECRET_TYPE = "envoy.extensions.transport_sockets.tls.v3.Secret",
        VIRTUAL_HOST_TYPE = "envoy.config.route.v3.VirtualHost",
    }
}
```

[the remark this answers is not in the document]
XAMPPRocky: >  Thought it would make rustfmt not split lines if they're joined to a single one per url.

It would, but there would now also have to be a `#[rustfmt::skip]` on top of each one, which IMO would it harder to read than just having the lines split.

Also FWIW there's no real overhead for a macro like that, it's just very simple expansion.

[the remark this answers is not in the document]
bombsimon: Oh yeah my bad, didn't test this properly, thought it would affect the whole group.

I mostly meant syntactically and to know what's going on without having to look what the macro does. I think it's just as good to have a regular `pub const` with the `&str` value and live with the line breaks but I don't have a very strong opinion if you want to go the macro way. Do what you think is nicest 😃

src/xds/provider/agones.rs
markmandel: +1 to what @XAMPPRocky said - the `rymder` and Agones SDK definitely has similar structures (because they are based on the same underlying data structures), but are aimed at different use cases. I think (at least for now), it's better to keep them split, so they don't have dependencies on each other - also, interacting with a K8s `GameServer` resource doesn't need the gRPC SDK integration for Agones, so it feels to me like something we should keep seperate.

src/main.rs
```diff
@@ -59,6 +59,44 @@ enum Commands {
...
+            short,
+            long,
+            default_value = "quilkin",
+            help = "Namespace under which the proxies run."
+        )]
+        config_namespace: String,
+        #[clap(
+            short,
+            long,
+            default_value = "gameservers",
```
markmandel, line 95: I recommend we set these both to `default ` for both. That namespace is the default, in a Kubernetes cluster - so if people are testing this out, it probably makes the most sense to point it there, since I expect people will dump everything in there sometimes anyway (or at least on first attempt).

src/main.rs
```diff
@@ -93,6 +131,27 @@ async fn main() -> quilkin::Result<()> {
...
+            admin_port,
+            provider,
+        } => {
+            let provider = match provider {
+                ProviderCommands::Agones {
+                    gameservers_namespace,
+                    config_namespace,
+                } => std::sync::Arc::from(
+                    quilkin::xds::provider::AgonesProvider::new(
+                        gameservers_namespace,
```
markmandel, line 145: Not for this PR (and I'm sure you've thought the same), but I expect long term, we'll want to support multiple namespaces, particularly for Agones GameServers (in dev, a namespace per developer is not uncommon).

src/xds/server.rs
```diff
@@ -0,0 +1,119 @@
...
+        Self { provider }
+    }
+
+    /// Creates a new server for a [DiscoveryServiceProvider] with a cache.
+    pub fn with_cache<P: DiscoveryServiceProvider + 'static>(provider: P) -> Self {
+        Self::new(Cache::new(provider))
+    }
+}
+
+#[tonic::async_trait]
```
markmandel, line 38: Does it make sense to pull in `async_trait` crate into Config.toml and use it directly? 

Looking at [the docs](https://docs.rs/tonic/latest/tonic/attr.async_trait.html), it says "A re-export of async-trait for use with codegen." -- small nit, but feels a bit weird to being using it in our own code?? 🤔

src/xds/provider/agones.rs
```diff
@@ -0,0 +1,438 @@
...
+        })
+    }
+}
+
+#[derive(kube::CustomResource, Clone, Debug, Deserialize, Serialize, JsonSchema)]
+#[kube(
+    group = "agones.dev",
+    version = "v1",
+    kind = "GameServer",
+    status = "GameServerStatus",
```
markmandel, line 229: Technically we don't have an actual k8s/status value for `GameServer` (for our use case it reduces hits to the API server) - but since we're not updating a `GameServer`, it likely doesn't matter for our use case (I'm guessing it might be required for this build toolchain? 🤔)

src/xds/provider/agones.rs
```diff
@@ -0,0 +1,438 @@
...
+    failure_threshold: i32,
+    /// Initial delay before checking health
+    #[serde(
+        rename = "initialDelaySeconds",
+        default = "default_initial_delay_seconds"
+    )]
+    initial_delay_seconds: i32,
+}
+
+fn default_period_seconds() -> i32 {
```
markmandel, line 268: I feel like you shouldn't need all of this default value logic (or keep them maintained in case any of them change). The default values would be set at the Agones controller/k8s layer, so you could potentially remove all of these and cut down on some long term maintenance costs.

src/xds/provider/agones.rs
```diff
@@ -0,0 +1,438 @@
...
+
+const QUILKIN_CONFIGMAP: &str = "quilkin-config";
+const DEFAULT_CLUSTER_NAME: &str = "default-quilkin-cluster";
+const QUILKIN_URL: &str = "quilkin.dev";
+const QUILKIN_TOKEN_KEY: &str = "tokens";
+const QUILKIN_TOKEN_LABEL: &str = "quilkin.dev/tokens";
+
+pub struct AgonesProvider {
+    config: kube::Api<k8s_openapi::api::core::v1::ConfigMap>,
+    gameservers: kube::Api<GameServer>,
```
markmandel, line 36: Rather than using `kube::Api`, we should use [Reflectors](https://github.com/kube-rs/kube-rs#reflectors) instead, such that we have a local cache that is updated on changes, that we can poll for free with no performance hit to the K8s API control plane (In client-go we use Informers and Listeners, but the concept is the same).

Constant polling against the K8s API server is unfortunately a great way to make the API control plane (and etcd) really upset and will cause the performance of the entire cluster to plummet.

src/main.rs
XAMPPRocky: That's fine with me, these are just what's in the current xDS implementation.

src/xds/provider/agones.rs
XAMPPRocky: I thought so too, but when I tried connecting to our clusters we hit some of these fields being absent (they are marked with `omitempty` in Go), and thus I need to provide some default value for them.

src/xds/provider/agones.rs
XAMPPRocky: Are you sure, because there is a `GameServerStatus` type that is defined here. https://github.com/googleforgames/agones/blob/9fa432c24d5aefbfcd6cf7eef61fcd4b35034663/pkg/apis/agones/v1/gameserver.go#L221

src/xds/provider/agones.rs
markmandel: Aaaah. That makes sense. It's not like the defaults are going to change on the Agones side, so it is going to be implemented once (🤞🏻)

src/xds/provider/agones.rs
markmandel: Yes, his is correct, we have a status value - but it's not registered a a `/status` subresource on `GameServer`, because reasons.

Details: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#subresources

(Basically it's a nice way of updating only a section of a CRD - but the way we treat `GameServers` in Agones, it's better if we update GameServer's entire value as a single transaction, so we don't do this for this resource)

If it's working for this, it doesn't actually matter - it might have been an issue if we were trying to update `/status` values on the `GameSever` as it might use the wrong API.

Mostly this is just a point of interest, and not a blocking issue in the PR.

XAMPPRocky:
Closing in favour of #527

PR #511, opened 2022-04-06
Metrics for Match Filter

### LINKED ISSUE #453: Add metrics to `Match` filter.
Let's increment counter metrics here for when we find a match vs not, for visibility

_Originally posted by @iffyio in https://github.com/googleforgames/quilkin/pull/447#discussion_r759939668_

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Implements metric counters for matched and fallthrough packets.

Closes #453

N/A

### DIFF
--- a/src/filters/match/metrics.rs
+++ b/src/filters/match/metrics.rs
@@ -30,13 +30,13 @@ impl Metrics {
     pub(super) fn new() -> prometheus::Result<Self> {
         Ok(Metrics {
             packets_matched_total: IntCounter::with_opts(filter_opts(
-                "packets_matched",
+                "packets_matched_total",
                 "Match",
                 "Total number of packets where the dynamic metadata matches a branch value.",
             ))?
             .register_if_not_exists()?,
             packets_fallthrough_total: IntCounter::with_opts(filter_opts(
-                "packets_fallthrough",
+                "packets_fallthrough_total",
                 "Match",
                 "Total number of packets that are processed by the fallthrough configuration",
             ))?
--- a/src/filters/match/metrics.rs
+++ b/src/filters/match/metrics.rs
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+use crate::metrics::{filter_opts, CollectorExt};
+use prometheus::{
+    core::{AtomicU64, GenericCounter},
+    IntCounter,
+};
+
+/// Register and manage metrics for this filter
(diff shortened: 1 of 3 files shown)


markmandel:
Oops, looks like I deleted my branch of htmltest when it was merged. Fix incoming.

src/filters/match/metrics.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+impl Metrics {
+    pub(super) fn new() -> prometheus::Result<Self> {
+        Ok(Metrics {
+            packets_matched_total: IntCounter::with_opts(filter_opts(
+                "packets_matched",
+                "Match",
+                "Total number of packets where the dynamic metadata matches a branch value.",
+            ))?
```
bombsimon, line 36: In the updated docs this is listed as `quilkin_filter_Match_packets_matched` but I fail to see where the `filter` part comes from. `filter_opts` just wraps the code below, where does the `_filter` part come from? Or will this be named `quilkin_Match_packets_match`?

```rust
Opts::new(name, description)
  .namespace("quilkin")
  .subsystem(subsystem)
```

To better follow the [prometheus naming convention](https://prometheus.io/docs/practices/naming/) that suggests a `total` suffix for counters, maybe an alternative could be like suggested change which would result in `quilkin_filter_packets_matchhed_total`?

```suggestion
            packets_matched_total: IntCounter::with_opts(filter_opts(
                "packets_matched_total",
                "filter",
                "Total number of packets where the dynamic metadata matches a branch value.",
            ))?
```

src/filters/match/metrics.rs
```diff
@@ -0,0 +1,46 @@
...
+            .register_if_not_exists()?,
+            packets_fallthrough_total: IntCounter::with_opts(filter_opts(
+                "packets_fallthrough",
+                "Match",
```
bombsimon, line 40: Suggesting the same change here which should result in `quilkin_filter_packets_fallthrough_total`:

```suggestion
                "packets_fallthrough_total",
                "filter",
```

bombsimon (commented):
Nice! Added a question and based on the reply a suggestion on alternative metric naming.

src/filters/match/metrics.rs
markmandel: > To better follow the prometheus naming convention that suggests a total suffix for counters

🤦🏻 forgot the `_total` at the end. Good catch! Thanks! Fix incoming.

> but I fail to see where the filter part comes from

https://github.com/googleforgames/quilkin/blob/8f26f76e16aeee1a0bbaa7d5d6663e76e5773765/src/metrics.rs#L61-L63

is where `filter_` comes from.

Re: replacing "Match" with "filter" - as we have in our other Filters, it's good to know which Filter the metric is coming from (hence it's in the name of the metric). - so the name of the Filter is an important part of the metric name. Also it prevents metric collisions.

So once the `_total` part is in the metric packet, this will match the metric convention that exists throughout the project:
* https://googleforgames.github.io/quilkin/main/book/filters/firewall.html#metrics
* https://googleforgames.github.io/quilkin/main/book/filters/local_rate_limit.html#metrics
* https://googleforgames.github.io/quilkin/main/book/filters/compress.html#metrics

(Although on review, I think some are missing a `_total` suffix).

Make sense?

XAMPPRocky (approved):

src/filters/match/metrics.rs
bombsimon: Totally, thanks for explaining that!

bombsimon (approved):

PR #514, opened 2022-04-12
Add conversion to protobuf for most core types

XAMPPRocky:
This splits out most of the type conversion code from #506 in order to make it easier to review and merge.``

### DIFF
--- a/src/filters/compress/config.rs
+++ b/src/filters/compress/config.rs
@@ -21,7 +21,8 @@ use serde::{Deserialize, Serialize};
 
 use super::compressor::{Compressor, Snappy};
 use super::quilkin::filters::compress::v1alpha1::{
-    compress::Action as ProtoAction, compress::Mode as ProtoMode, Compress as ProtoConfig,
+    compress::{Action as ProtoAction, ActionValue, Mode as ProtoMode, ModeValue},
+    Compress as ProtoConfig,
 };
 use crate::{filters::ConvertProtoConfigError, map_proto_enum};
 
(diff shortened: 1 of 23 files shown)


src/filters/load_balancer.rs
```diff
@@ -14,15 +14,17 @@
...
 mod endpoint_chooser;
 
 use crate::filters::{prelude::*, DynFilterFactory};
 
-use config::ProtoConfig;
+pub use config::{Config, Policy};
 use endpoint_chooser::EndpointChooser;
 
-pub use config::{Config, Policy};
+use self::quilkin::filters::load_balancer::v1alpha1 as proto;
```
markmandel, line 27: I like this pattern of using `proto` a lot 👍🏻

src/filters/compress/config.rs
```diff
@@ -21,7 +21,8 @@ use serde::{Deserialize, Serialize};
 
 use super::compressor::{Compressor, Snappy};
 use super::quilkin::filters::compress::v1alpha1::{
-    compress::Action as ProtoAction, compress::Mode as ProtoMode, Compress as ProtoConfig,
+    compress::{Action as ProtoAction, ActionValue, Mode as ProtoMode, ModeValue},
```
markmandel, line 24: Should we switch this one out to be `use super::quilkin::filters::compress::v1alpha1 as proto` and not alias each import individually, like has been refactored in the other Filters, for consistency?

markmandel (commented):
Thanks for breaking this out - made it much easier to review.

rezvaneh (approved):

src/filters/compress/config.rs
```diff
@@ -49,6 +50,22 @@ impl Default for Mode {
     }
 }
 
+impl From<Mode> for ProtoMode {
```
markmandel, line 53: Not sure if I was clear in the above comment which was resolved.

I was suggesting that we should follow the same pattern of using the `proto` import alias as we have elsewhere in this PR for this file as well for consistency (and I like the pattern better).

So this should become:

```suggestion
impl From<Mode> for proto::Mode {
```

and so on and so forth for the other items in this file as well.

XAMPPRocky: The problem with doing it for these types, is that they're embedded within the `compress` message namespace, so it would be `proto::compress::Mode::Snappy` for example rather than `proto::Mode::Snappy`. I could move the types outside the message in their Protobuf definition to fix that, but I wanted to make sure separately that that's what we want to do here separately.

markmandel (approved):

PR #515, opened 2022-04-13
Add `StaticFilter` trait

XAMPPRocky:
This PR is borne out of initially an effort to pull the `yaml_to_protobuf` changes to `FilterFactory` from #506 into a separate PR, and also add a `protobuf_to_yaml` method. However adding another new method was pretty tedious, and when thinking more holistically about the problem of converting configuration back and forth from YAML and Protobuf, I realised that we needed a more generic solution that would allow us to statically only define the types and have the same dynamic behaviour.

This thought process led me to this PR which adds a new `StaticFilter` trait which does exactly that. Now if you're writing a filter in Rust you no longer have to define a `FilterFactory` to construct it, you only implement `StaticFilter` and `Filter`, and then Quilkin will automatically *at compile-time* generate a virtual table safe version of `StaticFilter` (still called `FilterFactory` for the sake of not changing everything all at once).

This new trait design effectively eliminates all boilerplate needed when defining your own filter. For example; this is entire definition for the `Compress` as a filter (excluding the actual `Filter` impl). As shown, nearly everything is defined statically with types and constants, the only behaviour needed to be implemented is converting from an `Option<Self::Configuration>` into a `Self`.

```rust
impl StaticFilter for Compress {
    const NAME: &'static str = "quilkin.filters.compress.v1alpha1.Compress";
    type Configuration = Config;
    type BinaryConfiguration = proto::Compress;

    fn new(config: Option<Self::Configuration>) -> Result<Self, Error> {
        Ok(Compress::new(
            Self::ensure_config_exists(config)?,
            Metrics::new()?,
        ))
    }
}
```

It's hard to understate how much this single abstraction saves in terms of time and effort. The seperation provides two key things, the most obvious of which is mostly type-based API that is verified at compile-time for getting access to a filter and its configuration. The second benefit is that having a seperate trait allows us to split the dynamic dispatch `FilterFactory` functionality, this means that changes to `FilterFactory` have no affect on `StaticFilter` definitions. This is how this PR is able to add two safely add new `FilterFactory` methods which allow you to dynamically transcode the configuration format, and for it to still be net negative in terms of code added.

### Breaking Changes
- Filter modules no longer expose their name through a module level constant, instead preferring `StaticFilter::NAME`.
- Filter modules no longer expose a `factory` free function, use `StaticFilter::factory` to achieve the same effect.

### DIFF
--- a/docs/src/filters/writing_custom_filters.md
+++ b/docs/src/filters/writing_custom_filters.md
@@ -20,7 +20,7 @@ A [trait][Filter] representing an actual [Filter][built-in-filters] instance in 
 A [trait][FilterFactory] representing a type that knows how to create instances of a particular type of [Filter].
 
 - An implementation provides a `name` and `create_filter` method.
-- `create_filter` takes in [configuration][filter configuration] for the filter to create and returns a [FilterInstance] type containing a new instance of its filter type.  
+- `create_filter` takes in [configuration][filter configuration] for the filter to create and returns a [FilterInstance] type containing a new instance of its filter type.
   `name` returns the Filter name - a unique identifier of filters of the created type (e.g quilkin.filters.debug.v1alpha1.Debug).
 
 ### FilterRegistry
@@ -72,7 +72,7 @@ We start with the [Filter] implementation
 #
 // src/main.rs
 use quilkin::filters::prelude::*;
- 
+
 struct Greet;
 
 impl Filter for Greet {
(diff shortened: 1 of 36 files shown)


src/config/config_type.rs
```diff
@@ -48,19 +48,21 @@ impl ConfigType {
...
         self,
         filter_name: &str,
-    ) -> Result<(serde_json::Value, Static), Error>
+    ) -> Result<(serde_json::Value, TextConfiguration), Error>
     where
-        Dynamic: prost::Message + Default,
-        Static: serde::Serialize
-            + for<'de> serde::Deserialize<'de>
-            + TryFrom<Dynamic, Error = ConvertProtoConfigError>,
+        BinaryConfiguration: prost::Message + Default,
```
markmandel, line 56: That is a nice renaming 👍🏻

src/config/config_type.rs
```diff
@@ -48,19 +48,21 @@ impl ConfigType {
...
+        TextConfiguration:
```
markmandel, line 57: `StringConfiguration` ? (Rather than `TextConfiguration`)?

Just a possible suggestion, non-blocking.

src/filters.rs
```diff
@@ -43,21 +42,84 @@ pub mod token_router;
...
+    /// The binary configuration of the filter. **Must** be [`prost`] compatible,
+    /// and be convertible to and from [`Self::Configuration`].
+    type BinaryConfiguration: prost::Message
+        + Default
+        + TryFrom<Self::Configuration>
+        + Send
+        + Sync
+        + Sized;
+
+    /// Instaniates a new [`StaticFilter`] from the given configuration, if any.
```
markmandel, line 102:
```suggestion
    /// Instantiates a new [`StaticFilter`] from the given configuration, if any.
```

Nit: typo

docs/src/filters/writing_custom_filters.md
```diff
@@ -94,31 +94,70 @@ Next, we implement a [FilterFactory] for it and give it a name:
 # #![allow(unused)]
 # fn main() {
 #
+# #[derive(Default)]
```
markmandel, line 97: Line above this: 
> Next, we implement a [FilterFactory] for it and give it a name:

Should probably be something like:

> Next, we implement a [StaticFilter] for it and give it a name:

docs/src/filters/writing_custom_filters.md
```diff
@@ -20,7 +20,7 @@ A [trait][Filter] representing an actual [Filter][built-in-filters] instance in
 A [trait][FilterFactory] representing a type that knows how to create instances of a particular type of [Filter].
 
 - An implementation provides a `name` and `create_filter` method.
-- `create_filter` takes in [configuration][filter configuration] for the filter to create and returns a [FilterInstance] type containing a new instance of its filter type.  
+- `create_filter` takes in [configuration][filter configuration] for the filter to create and returns a [FilterInstance] type containing a new instance of its filter type.
```
markmandel, line 23: I think this section needs to be replaced as one about `StaticFilter`, which should explain how it exports a `FilterFactory`.

docs/src/filters/writing_custom_filters.md
```diff
@@ -130,7 +169,7 @@ impl FilterFactory for GreetFilterFactory {
 #### 3. Start the proxy
 
 We can run the proxy in the exact manner as the default Quilkin binary using the [run][runner::run] function, passing in our custom [FilterFactory].
-Let's add a main function that does that. Quilkin relies on the [Tokio] async runtime, so we need to import that 
+Let's add a main function that does that. Quilkin relies on the [Tokio] async runtime, so we need to import that
```
markmandel, line 172: Line above:

> We can run the proxy in the exact manner as the default Quilkin binary using the [run][runner::run] function, passing in our custom [FilterFactory] that is exported from our implemented [StaticFilter]

docs/src/filters/writing_custom_filters.md
```diff
@@ -220,41 +259,36 @@ First let's create the config for our static configuration:
 ```rust,no_run,noplayground
 // src/main.rs
 # use serde::{Deserialize, Serialize};
-# #[derive(Serialize, Deserialize, Debug)]
+# use quilkin::filters::prelude::*;
```
markmandel, line 262: Heading above:

> ###### 4. Finally, update `StaticFilter` to extract the greeting from the passed in configuration and forward it onto the `Greet` Filter.

docs/src/filters/writing_custom_filters.md
```diff
@@ -282,7 +316,7 @@ let config = match args.config.unwrap() {
 
 The [Dynamic][ConfigType::dynamic] contains the serialized [Protobuf] message received from the [management server] for the [Filter] to create.
 As a result, its contents are entirely opaque to Quilkin and it is represented with the [Prost Any][prost-any] type so the [FilterFactory]
-can interpret its contents however it wishes.  
+can interpret its contents however it wishes.
```
markmandel, line 319: Haven't got a better spot to put this, but all the documentation about Dynamic Config vs Static should be changed to align with Text/Binary configuration.

markmandel (commented):
This is a _really_ nice refactor!!!

The "writing a customer filter" documentation needs a once over to update to this new architecture though (you may well have been waiting to do that post initial review).

bombsimon (commented):
Looks like a nice way to implement filters!

docs/src/filters/writing_custom_filters.md
```diff
@@ -94,31 +94,70 @@ Next, we implement a [FilterFactory] for it and give it a name:
 # #![allow(unused)]
 # fn main() {
 #
+# #[derive(Default)]
 # struct Greet;
+# impl Greet {
```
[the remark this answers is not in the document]
XAMPPRocky: Yes in [mdbook](https://rust-lang.github.io/mdBook/format/mdbook.html#hiding-code-lines) and rustdoc you can hide code lines by prefixing them with `#`.

src/config/config_type.rs
XAMPPRocky: I chose `Text` over `String` because `String` is a specific type in Rust as opposed to just "human readable", and it's shorter.

bombsimon (approved):
Really nice documentation update, I think it's great to have the fully working code example in `examples` and making the implementation details clearer. Also nice with example on how to test the filter with `nc`!

PR #525, opened 2022-04-26
Refactor configuration and builder pattern

### LINKED ISSUE #520: Make `config::Config` compatible with xDS manager
Currently the xDS management server uses a slightly different file configuration than the quilkin client, because the Quilkin client requires some fields that aren't needed for the xDS manager. I think to simplify the code, that we re-use the same configuration type for both sides. To make this work we would just need to make a few changes to the schema to make it compatible with both agents. Mainly removing the dynamic/static distinction, and allowing all the current sub-properties to be optional in the root object.   This would allow both client and server to read the same properties.

XAMPPRocky:
This PR updates Quilkin's configuration to be more dynamic friendly, and to allow for the configuration to contain not just `static` or `dynamic`, but instead make `endpoints`, `filters`, and `management_servers` each optional fields which allows for us to reuse the same configuration on the xDS server side as we would on the Quilkin client side. The only change in behaviour here is that you can now set Quilkin's initial state before it connects to the management servers.

### Breaking Changes
- Removed `static` and `dynamic` properties in `Config`, instead replacing them with `endpoints`, `filters`, and `management_servers` directly.
- Removed `proxy::Builder`, instead constructing `Server` from `Config` directly.

closes #520


rezvaneh (approved):
Nice work!

bombsimon (approved):
LGTM

PR #528, opened 2022-04-29
Update xDS doc 

rezvaneh:
- It updates the doc for xDS server management with the new Rust implementation which we will have soon.
- It moves the doc to the [guide](https://googleforgames.github.io/quilkin/main/book/).


docs/src/xds.md
```diff
@@ -36,6 +29,113 @@ Since the range of resources configurable by the xDS API extends that of Quilkin
...
+     quilkin.dev/tokens: MXg3aWp5Ng==,OGdqM3YyaQ==
+   ```
+
+   > Since an Agones GameServer can have multiple ports exposed, if multiple ports are in
+   > use, the server looks for the port named `default` and picks that as the endpoint's
+   > port (otherwise it picks the first port in the port list).
+
+2. Filter chain is configurable on a per-proxy basis. By default an empty filter chain is used and from there the filter chain can configured using a configMap name `quilkin-config` on the proxy's pod.
+
+As an example, the following runs the server with subcommnad _agones_ against a cluster (using default kubeconfig configuration) where Quilkin pods run in the `quilkin` namespace and game-server pods run in the `gameservers` namespace:
```
XAMPPRocky, line 55:
```suggestion
As an example, the following runs the server with subcommand `manage agones` against a cluster (using default kubeconfig configuration) where Quilkin pods run in the `quilkin` namespace and game-server pods run in the `gameservers` namespace:
```

docs/src/xds.md
```diff
@@ -36,6 +29,113 @@ Since the range of resources configurable by the xDS API extends that of Quilkin
...
+quilkin manage --port 18000 agones --config-namespace quilkin --gameservers-namespace gameservers
+```
+
+> A proxy's pod must have a `quilkin.dev/role` key in `quilkin-config` configMap set to the value `proxy` in order for the management server to detect the pod as a proxy and push updates to it.
+
+> Note that currently, the server can only discover resources within a single cluster.
+
+### Filesystem
+
+The file server command is primarily an example and mostly suitable for demo purposes. As a result, some configuration options and features might be missing. This file implementation watches a configuration file on disk and sends updates to proxies whenever that file changes.
```
XAMPPRocky, line 67: I think the comment about it being demo-only is not needed, you totally could use it in production, there'd be nothing wrong with it for deployments that are mostly static and/or small.

```suggestion
The filesystem provider watches a configuration file on disk and sends updates to proxies whenever that file changes.
```

docs/src/xds.md
```diff
@@ -36,6 +29,113 @@ Since the range of resources configurable by the xDS API extends that of Quilkin
...
+  typed_config: {
+    # @type must be equivalent to name - the name of the filter.
+    # It is an extra, required field.
+    '@type': string
+    # ...
+    # The rest of the body contains filter specific configuration or
+    # is empty if the filter has no configuration.
+  }
+}]
+```
```
XAMPPRocky, line 112: I'd remove all of this and just point people at https://googleforgames.github.io/quilkin/main/book/proxy-configuration.html as this is the configuration the provider will be using.

```suggestion
```

docs/src/xds.md
```diff
@@ -36,6 +29,113 @@ Since the range of resources configurable by the xDS API extends that of Quilkin
...
+    port": 29
+    metadata:
+      'quilkin.dev':
+         tokens:
+         - "MXg3aWp5Ng=="
+filterchain:
+- name: quilkin.filters.debug.v1alpha1.Debug
+  typed_config:
+    '@type': quilkin.filters.debug.v1alpha1.Debug
+    id: hello
```
XAMPPRocky, line 128:
```suggestion
filters:
- name: quilkin.filters.debug.v1alpha1.Debug
  config:
    id: hello
```

docs/src/xds.md
```diff
@@ -36,6 +29,113 @@ Since the range of resources configurable by the xDS API extends that of Quilkin
...
+```
+
+## Admin server
+
+In addition the gRPC server, a http server (configurable via `--admin-port`is also started to serve administrative functionality.
+The following endpoints are provided:
+- `/ready`: Readiness probe that returns a 5xx if communication with the Kubernetes api is problematic.
+- `/live`: Liveness probe that always returns a 200 response.
+- `/metrics`: Exposes Prometheus metrics.
```
XAMPPRocky, line 137: I'd remove this and instead mention and link to the "Administration" chapter.

```suggestion
```

XAMPPRocky (approved):

PR #534, opened 2022-05-20
Api: Add uptream_address to proxy config

### LINKED ISSUE #345: Bind Quilkin's sender socket to a given address:port
This is a feature proposal issue
It would be nice to be able to bind Quilkin's sender socket to a given address and port pair. When Quilkin proxies the packets to an upstream host server, it chooses a random socket to use. This feature would remove the randomness and allow us to control which address and port pair to use.

Why do we need this? 
- The first reason behind this feature request is that our telco use-case uses RTPengine as a backend RTP proxy. RTPengine supports resilience using a Redis store, but due to its design, it stores the connection's port number as well. So when it comes to resilience the second RTPengine will read call the call record from the Redis store and proxy the packets to a port that was used by the first RTPengine... Port binding would solve this. 
- The second design issue is that RTPengine thinks that it's talking directly to a client and not to some proxy. In a call setup, there are initiating SDP messages that include the client's sender port. When the RTPengine gets these SDP messages it expects the media stream to come from that port number that was in the SDP control messages. There are workarounds but the best solution would be to bind the sender socket to that very port.

We are using a separate control and media plane, and dynamically (via an XDS server) create a media plane pipeline for every single RTP/RTCP stream. After the mentioned SDP messages we would know where to bind the sender socket.

Not sure if this should be a filter or not. Or in general how the configuration would look like?  What do you think about this?

nezdolik:
Signed-off-by: Kateryna Nezdolii <kateryna.nezdolii@gmail.com>

**What type of PR is this?**
 feature

Adds possibility to configure Quilkin with upstream address to bind to when proxying packets upstream.

Closes #345


src/proxy/sessions/session.rs
```diff
@@ -134,8 +136,11 @@ impl SessionArgs {
 impl Session {
     /// internal constructor for a Session from SessionArgs
     async fn new(args: SessionArgs) -> Result<Self> {
-        let addr = (std::net::Ipv4Addr::UNSPECIFIED, 0);
-        let socket = Arc::new(UdpSocket::bind(addr).await.map_err(Error::BindUdpSocket)?);
+        let socket = Arc::new(
```
markmandel, line 139: Won't this fail if you end up with more than one session (or endpoint)? Since each session will attempt to bind to the same port, and will then fail, and 

I'm not quite sure this will work 🤔 Maybe each endpoint you could be able to specify the port?

The proxy is built to be non-transparent, so Sessions do use unique sender and receiver information as it's storage for individual Sessions:

https://github.com/googleforgames/quilkin/blob/786d0d3561ad758aee7aa90966f9c106e4847f30/src/proxy/sessions/session_manager.rs#L26-L27

Breaking that contract could have some unintended consequences as well.

I would also advocate that there is some integration tests for this to ensure it works as expected.

markmandel: Just reviewing @XAMPPRocky 's [original comment ](https://github.com/googleforgames/quilkin/issues/345#issuecomment-887297745
)

Rather than bind a new port on each session, we probably need to do something more like having an `Option<Arc<UdpSocker>>` that gets passed into the Session, and if it's `None` then we bind to the ephemeral port.

Still not quite sure how/if that solves the unique key issue within `SessionManager` though 🤔 I'll write some thoughts on the original PR for some design discussion.

markmandel:
If you rebase against `upstream/main` that will fix the documentation 404 error.

src/proxy/sessions/session.rs
nezdolik: Thanks for review @markmandel, good catch. Tested on live Quilkin process but with one session. That should de definitely covered in integration test. Looking forward to design discussions, meanwhile will pause dev work on this.

markmandel:
Shall we close this PR, until the details of this issue are worked out in #345? (which is also a gentle bump to take another look at the suggestions in that ticket 😄 )

nezdolik:
@markmandel makes sense, closing.

Issue #510, opened 2022-04-05
e2e tests for Agones integration


markmandel:
**Is your feature request related to a problem? Please describe.**

Once we have the Agones integration from #506, we don't currently have a way to automatically test the integration end to end (which is not a good thing).

**Describe the solution you'd like**

So we need a way to run the e2e tests both locally and also in Cloud Build CI. We can probably break this down into a few steps/built in tools into our Makefile.

The only tricky part with local testing is that connectivity to something like Minikube on Agones can be tricky, because of the layers of virtualisation (see [guide](https://agones.dev/site/docs/installation/creating-cluster/minikube/#known-working-drivers))

I believe we can mimic what we did in Agones to do a similar thing in Rust as we did with Go (sacrificial draft):
* We'll need a `make` target to install the latest version of Agones on the current authorised cluster (it's a `helm install` - [pretty easy](https://agones.dev/site/docs/installation/install-agones/helm/)). (be good to have a terraform script that sets up the cluster too).
* All the [e2e tests in Agones are standard Go tests](https://github.com/googleforgames/agones/tree/main/test/e2e), that use the local `.kube` credentials to run. It looks like Rust's `kube` client will happily do the same. This has two benefits - you can run individual tests in your IDE, and also run the whole suite with standard Rust tooling.
* For CI Agones [has a custom build step that does the installation and running of tests](https://github.com/googleforgames/agones/tree/main/build/e2e-image) - we could do similar things. The fun part here is that we need to use a shared lock to make sure no two tests are running at the same time (unless there is a way to make it so that it doesn't matter 🤔). In Agones we use consul to do the distributed lock, although I have debated implementing something using [gcslock](https://github.com/mco-gh/gcslock/blob/master/gcslock.sh) or similar.

**Describe alternatives you've considered**

None. We'll need automated testing for this.

**Additional context**

I would probably suggest starting with getting tests running locally in a Rust context, and then build out the CI platform from there.

markmandel:
Starting work on this - starting with creating a simple terraform script for creating a agones powered GKE cluster, and creating a module that will create some kind of simple test we can expand on.

Will probably add some of the tooling (terraform, helm, gcloud etc) into the build image as well for convenience if they are not installed locally.

[...]

PR #549, opened 2022-07-21
Dev: Build, Push and Test Agones Integrations

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

This builds out the `build/Makefile` tooling such that you can build, push and test the Quilkin Agones integration against either Minikube or a hosted Kubernetes cluster (tested against GKE + Artifact Registry) in one operation.

Also includes:

* Refactoring such that variable naming across `cargo test` and `make` targets all lined to up to be consistent.
* If `IMAGE_TAG` is not present in `cargo test`, panic.
* Upgrade Agones version in Terraform script.
* Update release cloudbuild.yaml - include an update to changelog generator, as I found a bug in the version we were using.
* More aggressive .dockerignore. Got the build context down from >3GB to  ~900M, which was a big improvement in image creation speed.

Work on #510

Once this is merged, I can work on integrating it with our CI builds.

### DIFF
--- a/agones/README.md
+++ b/agones/README.md
@@ -7,17 +7,6 @@ This folder containers the integration tests for Quilkin and Agones integration.
 * A Kubernetes cluster with [Agones](https://agones.dev) installed.
 * Local authentication to the cluster via `kubectl`.
 
-## Running Tests
-
-To run the tests, run `cargo test` in this folder. This will run the e2e to tests with the default Quilkin image.
-
-When writing new tests for new features, you will want to specify a development image hosted on a container 
-registry to test against. This can be done through the `QUILKIN_IMAGE` environment variable like so:
-
-```shell
-QUILKIN_IMAGE=us-docker.pkg.dev/my-project-name/dev/quilkin:0.4.0-dev cargo test
-```
-
 ## Creating an Agones Minikube Cluster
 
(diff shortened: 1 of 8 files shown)


agones/README.md
```diff
@@ -55,6 +44,68 @@ terraform apply -var project="<YOUR_GCP_ProjectID>"
...
+### Build, Push and Test in one Go 💪
+
+The [`build/Makefile`](../build/Makefile) provides a targets that can be executed to build a development image, 
+push it a appropriate location, and run the set of Agones integration tests, depending on where you Kubernetes 
+cluster is set up. 
+
+#### Minikube
+
+This target assumes that you have a [working minikube cluster](#creating-an-agones-minikube-cluster),
+under the profile `quilkin`, with Agones installed, and the local `.kube` configuration is currently
```
markmandel, line 68: In the [Agones minikube documentation](https://agones.dev/site/docs/installation/creating-cluster/minikube/), it directs you to use a profile named `agones`. 

I went in the direction to push people to keep the Quilkin testing separate from other things, so picked a different default profile for these targets -- but what do you think, should it be `agones` ?

Can also change it later if it starts to annoy people.

XAMPPRocky (approved):
LGTM, I obviously haven't tested it locally but I think it's better to give that feedback as its own PR or issue.

PR #551, opened 2022-07-28
Run Agones tests in CI

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

This makes it such that on each PR, the Agones integration tests are run
against the GKE + Agones cluster specified in ./agones/main.tf

This includes:
* Fix for main.tf to align release branch with Agones version.
* Configure how long to wait before cleaning up past test namespaces.
* Cleanup and optimisation of cloudbuild.yaml.

Work on #510

### DIFF
--- a/agones/src/lib.rs
+++ b/agones/src/lib.rs
@@ -20,7 +20,9 @@ use std::{
     time::{SystemTime, UNIX_EPOCH},
 };
 
-use k8s_openapi::{api::core::v1::Namespace, apimachinery::pkg::apis::meta::v1::ObjectMeta};
+use k8s_openapi::{
+    api::core::v1::Namespace, apimachinery::pkg::apis::meta::v1::ObjectMeta, chrono,
+};
 use kube::{
     api::{DeleteParams, ListParams, PostParams},
     Api, ResourceExt,
@@ -33,6 +35,7 @@ mod pod;
 static CLIENT: OnceCell<Client> = OnceCell::const_new();
 #[allow(dead_code)]
 const IMAGE_TAG: &str = "IMAGE_TAG";
+const DELETE_DELAY_SECONDS: &str = "DELETE_DELAY_SECONDS";
 
 pub struct Client {
     /// The Kubernetes client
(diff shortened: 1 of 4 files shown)


cloudbuild.yaml
```diff
@@ -16,35 +16,34 @@ steps:
   - name: gcr.io/cloud-builders/git
     args: [ submodule, update, --init, --recursive ]
     id: fetch-git-submodules
+    waitFor:
+      - "-"
```
XAMPPRocky, line 20: Maybe add a comment here explaining this. Maybe this is obvious to people who use cloudbuild, but it's not obvious to me compared to other usages of waitFor.

agones/src/lib.rs
```diff
@@ -75,10 +78,23 @@ async fn setup_namespace(client: kube::Client) -> String {
     let lp = ListParams::default().labels("owner=quilkin-test");
     let nss = namespaces.list(&lp).await.unwrap();
     let dp = DeleteParams::default();
+
+    let delay = env::var_os(DELETE_DELAY_SECONDS)
+        .map(|value| chrono::Duration::seconds(value.into_string().unwrap().parse().unwrap()));
```
XAMPPRocky, line 83: Nit: You don't need to use `var_os` if you're converting to string anyway, and you can flatten the unwraps.

```suggestion
    let delay = env::var(DELETE_DELAY_SECONDS)
        .as_deref()
        .ok()
        .and_then(|s| i64::parse(s).ok())
        .map(chrono::Duration::seconds);
```

XAMPPRocky (approved):
LGTM, only nits

cloudbuild.yaml
markmandel: Excellent suggestion - will do.

markmandel:
GCP IAM permission on fail (which I expected but wasn't sure if we would hit).

agones/src/lib.rs
markmandel: I swear I never saw `env::var()` existing before. Totally missed that. Thank you!

markmandel:
That took an hour - I should look at why. I think some of my optimisations in cloudbuild.yaml caused a bunch of redundant work to be done.

markmandel:
Ah! Found it.

PR #555, opened 2022-08-08
Fix `Config::default` impl, and use default when no config is present

XAMPPRocky:
Previously the default from `serde` and the default from `Default::default` were slightly different (namely admin was null). This fixes that and removes the superfluous requirement to provide a config file at startup since Quilkin is now runtime updatable.


markmandel (commented):
Looks good!

Can we update documentation here please:
https://googleforgames.github.io/quilkin/main/book/using.html

Also, should we remove the https://github.com/googleforgames/quilkin/blob/main/image/quilkin.yaml from the Docker image as well?

docs/src/using.md
```diff
@@ -13,9 +13,14 @@ has debug level logging enabled.
...
-Quilkin needs to be run with an accompanying [configuration file](./proxy-configuration.md), like so:
+Quilkin can be run with an accompanying [configuration file](./proxy-configuration.md), like so:
 
-`quilkin --config="configuration.yaml" run`
+```
+quilkin --config=quilkin.yaml run
+```
+
+By default quilkin will check both the current directory and `/etc/quilkin`
+for `quilkin.yaml`.
```
markmandel, line 23:
```suggestion
By default Quilkin will check both the current directory and `/etc/quilkin`
for `quilkin.yaml`. If `quilkin.yaml` is not found, the system will start in a no-op state that receives incoming packets, but does not send them anywhere.
```

WDYT?

markmandel:
Just for my own edification (and we should probably document this too) - if you don't provide a config file, how do you tell the Quilkin Proxy its new configuration? Don't you at least need to tell it what xDS endpoint to point to?

Or is there some other way?

XAMPPRocky:
> Just for my own edification (and we should probably document this too) - if you don't provide a config file, how do you tell the Quilkin Proxy its new configuration? Don't you at least need to tell it what xDS endpoint to point to?

This bring up a good point that currently the `run` command doesn't have a good default state. This was written mostly with `manage` in mind which does have a good default, and the most important options are available via the command line. I think we should get `run` into the same state (as a seperate PR).

Using other proxies as prior art (looking at mostly Caddy for the CLI parts, but also Envoy and Traefik), I think we should have provide a few of the important options through the `run` command that would used if you haven't provided a configuration file. I think if you're invoking `run` without a config, you essentially want to run quilkin as a simple reverse proxy. So we should have a required `--to/--dest` flag that you have to provide a socket address to forward packets to. We'd also provide optional `--from` to change the listening port, and a optional `--management-server` flag to allow you to dynamically update it.

Also as an aside to improving the `run` command, I think we should align with proxies like envoy and traefik, and by default watch the config file that we read from if one is present so updating the file automatically updates the proxy without setup. This is already implemented in the code for the manager, we just need to also call it in run.

markmandel:
> > Just for my own edification (and we should probably document this too) - if you don't provide a config file, how do you tell the Quilkin Proxy its new configuration? Don't you at least need to tell it what xDS endpoint to point to?
> 
> This does bring up a good point that currently the `run` command doesn't have a good default state. This was written mostly with `manage` in mind which does have a good default, and the most important options are available via the command line. I think we should get `run` into the same state (as a seperate PR).
> 
> Using other proxies as prior art (looking at mostly Caddy for the CLI parts, but also Envoy and Traefik), I think we should have provide a few of the important options through the `run` command that would used if you haven't provided a configuration file. I think if you're invoking `run` without a config, you essentially want to run quilkin as a simple reverse proxy. So we should have a required `--to/--dest` flag that you have to provide a socket address to forward packets to. We'd also provide optional `--from` to change the listening port, and a optional `--management-server` flag to allow you to dynamically update it (This would also make `--to` optional). This would also allow us to change the current health code to state the proxies are unhealthy if they currently have no endpoints.
> 
> Also as an aside to improving the `run` command, I think we should align with proxies like envoy and traefik, and by default watch the config file that we read from if one is present so updating the file automatically updates the proxy without setup. This is already implemented in the code for the manager, we just need to also call it in run.

100% +1 on all of the above! Love it.

markmandel:
Created a couple of issues from the above commentary.

Otherwise, only thing blocking for me is the documentation tweaks, and making the tests pass 👍🏻

PR #561, opened 2022-08-09
Improve build-image incremental workflow

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Several improvements to workflows that flow through the build-image to improve build and iteration times.

This includes:

## Removal of debug images

We don't need this anymore, so no reason to build it for each architecture.

## Use `build-image` target directory

I realised that when mixing the `build-image` workflow with local rust tools, that going between them, they would invalidate each other's incremental cache, since it was fingerprinted on absolute directories - drastically expanding build times between iterations.

Moving to a specific target directory for `build-image` usage drastically improves both the incremental build times of images, and every other workflow through `build-image`, such as e2e testing. (`make test-agones` can be used iteratively now).

## More aggressive docker ignore

We only use a few files for the Docker image, so let's ignore everything else. Took the docker context down from 1GB+ to < 50MB.

On a linux machine all this brings incremental builds down from 10's of minutes to a minute or less, since only what has changed is compiled and pushed as we can now take advantage of rust incremental build and docker image caching.

Work on #553

I know what the impact is on a Linux environment, would be good to know how it improves things on a Mac and Windows environment where there is an extra layer of virtualisation.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -34,19 +34,12 @@
 /sdks
 /src
 /tests
-target/debug
-target/release
-target/**.d
-target/*/*/build
-target/*/*/deps
-target/*/*/incremental
-target/criterion
-target/doc
-target/tmp
+/target
 
 # windows files
 *.exe
 *.rlib
 
+!target/build-image/x86_64-unknown-linux-gnu/release/quilkin
 !dependencies-src.zip
 !image/quilkin.yaml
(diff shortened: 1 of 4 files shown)


markmandel:
CI failure should be fixed by #560

XAMPPRocky (approved):

markmandel:
Ah! This is actually a real issue:

```
Step #4 - "test-docs": book/filters/load_balancer.html
Step #4 - "test-docs":   target does not exist --- book/filters/load_balancer.html --> ../../api/quilkin/filters/load_balancer/struct.Config.html
```

I'll go fix.

PR #570, opened 2022-08-15
Refactor CLI organisation, and improve termination

XAMPPRocky:
As part of what we discussed in #555 we want to improve the CLI of `run` to be better, as part of that to make code for the commands more easily maintainable, I've first gone ahead and done a small refactor that moves each command into its own file and implementing them as their own types with a method to run them. This improves locality by a lot as now everything related to one command exists entirely in that one file.

The only actual changes in this PR are the following.

- In `src/main.rs` we now catch the result ourselves and ensure to emit it as a `tracing::error`, using the `Display` formatting before exiting with a non-zero error code. Previously it would use the `Debug` formatting and output to `stdout`.
- We now expose a `Cli` type from the library that can be used to drive the application, as opposed to exposing `run` and `manage` as free functions.


bombsimon (approved):
LGTM! 🧹

PR #571, opened 2022-08-22
Resubscribe to resources after disconnect, fix test flakiness

XAMPPRocky:
Previously we were reconnecting to the stream, but we weren't resubscribing to resources, so now the stream tracks what resources it's requested in the past and makes fresh discovery requests on reconnect.


bombsimon (approved):
LGTM!

PR #574, opened 2022-08-25
Add arguments to `quilkin run`

### LINKED ISSUE #572: Configure basic configuration values with command flags
> Just for my own edification (and we should probably document this too) - if you don't provide a config file, how do you tell the Quilkin Proxy its new configuration? Don't you at least need to tell it what xDS endpoint to point to?

This does bring up a good point that currently the `run` command doesn't have a good default state. This was written mostly with `manage` in mind which does have a good default, and the most important options are available via the command line. I think we should get `run` into the same state (as a seperate PR).

Using other proxies as prior art (looking at mostly Caddy for the CLI parts, but also Envoy and Traefik), I think we should have provide a few of the important options through the `run` command that would used if you haven't provided a configuration file. I think if you're invoking `run` without a config, you essentially want to run quilkin as a simple reverse proxy. So we should have a required `--to/--dest` flag that you have to provide a socket address to forward packets to. We'd also provide optional `--from` to change the listening port, and a optional `--management-server` flag to allow you to dynamically update it (This would also make `--to` optional). This would also allow us to change the current health code to state the proxies are unhealthy if they currently have no endpoints.

_Originally posted by @XAMPPRocky in https://github.com/googleforgames/quilkin/issues/555#issuecomment-1213493828_

XAMPPRocky:
Closes #555, closes #572

This adds the basic options to run quilkin as a reverse proxy, allowing you to set the listen port and the destination address. Additionally I added some `admin` related options for setting admin port and disabling admin entirely.


markmandel:
Can we get some docs, probably here https://googleforgames.github.io/quilkin/main/book/using.html please?

Re-running CI to see if that's a flaky test error or not.

XAMPPRocky:
It's not, I have the fix locally. The change is that the admin spawning is pulled out of the proxy to let the admin logic be shared amongst all commands, and these are tests that are testing the admin server.

image/Dockerfile
```diff
@@ -16,8 +16,7 @@ FROM gcr.io/distroless/cc:nonroot as base
 WORKDIR /
 COPY ./license.html .
 COPY ./dependencies-src.zip .
-COPY --chown=nonroot:nonroot ./image/quilkin.yaml /etc/quilkin/quilkin.yaml
-COPY ./target/image/quilkin .
+COPY ./target/image/quilkin /usr/local/bin
```
markmandel, line 19: I'm fairly sure that there is no $PATH in a distroless image, since there is no shell, hence the hardcoded path in the `ENTRYPOINT`.

https://github.com/GoogleContainerTools/distroless/blob/main/README.md#entrypoints

XAMPPRocky: Yeah I know but they're still Debian images, so I wanted to see if that would work.

markmandel: 😄 worth a shot!

cloudbuild.yaml
```diff
@@ -46,16 +46,6 @@ steps:
     args:
       - build
     id: build
-  # Run the built images for 5 seconds to make sure that the entrypoint and default config works out of the box
-  - name: gcr.io/cloud-builders/docker
-    dir: ./build
-    entrypoint: bash
-    args:
-      - '-c'
-      - 'timeout --signal=INT --preserve-status 5s docker run --rm ${_REPOSITORY}quilkin:$(make version)'
```
markmandel, line 55: Rather than deleting, could make it (something like)

`'timeout --signal=INT --preserve-status 5s docker run --rm ${_REPOSITORY}quilkin:$(make version)' run`

To make it run in `run` mode. 

Although one could argue the integration tests also test this (although this would provide a nicer error message if something basic was broken). I could go either way on this one.

XAMPPRocky:
@markmandel Rather than duplicating documentation, I figured it was easiest and best to expose the CLI in the rust documentation and just link to that. That way we're maintaining a single source of truth as our documentation now matchces what the CLI outputs in help.

I'm having trouble figuring out what these Agones tests want me to do, do you have any idea?

markmandel:
> @markmandel Rather than duplicating documentation, I figured it was easiest and best to expose the CLI in the rust documentation and just link to that. That way we're maintaining a single source of truth as our documentation now matches what the CLI outputs in help.

Makes sense.

> 
> I'm having trouble figuring out what these Agones tests want me to do, do you have any idea?

I'm at an event at the moment, but will take a look as soon as I can 👍🏻

markmandel:
Looks like a clippy failure:

```
error[E0308]: mismatched types
  --> src/cli/run.rs:56:61
   |
56 |                 clusters.default_cluster_mut().localities = vec![self.to.clone().into()];
   |                 -----------------------------------------   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `cluster::LocalitySet`, found struct `std::vec::Vec`
   |                 |
   |                 expected due to the type of this binding
   |
   = note: expected struct `cluster::LocalitySet`
              found struct `std::vec::Vec<_>`
   = note: this error originates in the macro `vec` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0308`.
```

markmandel:
Ah, on my mobile, so difficult to see past build output.

Have you been able to replicate the test failure by running the integration test locally? Either directly, or through the Makefile?

Then you could introspect the Pods and see what is failing.

If not, I'll be able to take the PR for a spin on Friday and see what is failing.

XAMPPRocky:
> Have you been able to replicate the test failure by running the integration test locally? Either directly, or through the Makefile?

I'm unable to use the Makefile, it just gets killed. I haven't tried directly yet.

```
#11 147.3 <jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
#11 147.3 <jemalloc>: (This is the expected behaviour if you are running under QEMU)
#11 147.4 <jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
#11 147.4 <jemalloc>: (This is the expected behaviour if you are running under QEMU)
#11 147.4 <jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
#11 147.4 <jemalloc>: (This is the expected behaviour if you are running under QEMU)
#11 147.5 <jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
#11 147.5 <jemalloc>: (This is the expected behaviour if you are running under QEMU)
#11 149.5    Compiling quote v1.0.21
#11 150.0 <jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
#11 150.0 <jemalloc>: (This is the expected behaviour if you are running under QEMU)
#11 152.6 Killed
------
executor failed running [/bin/sh -c wget --quiet https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init &&     chmod +x rustup-init &&     ./rustup-init -y --no-modify-path --default-toolchain $RUST_TOOLCHAIN &&     rm rustup-init &&     chmod -R a+w $RUSTUP_HOME $CARGO_HOME &&     rustup component add rustfmt clippy &&     rustup target add x86_64-pc-windows-gnu &&     cargo install cargo-watch mdbook &&     cargo install cargo-about &&     cargo install --locked cargo-deny &&     rustup --version &&     cargo --version &&     rustc --version]: exit code: 137
make: *** [ensure-build-image] Error 1
```

markmandel:
🤔 are you on a M1 mac?

markmandel:
Aaah, that old chestnut. Trying to install the x86 version of rust on a aarch64 chip is going to fall over every time.

We should put some OS checks in the makefile to manage that. I don' t have access to a M1 mac, so will be tricky for me to test, but I can steal some stuff from the Agones project, because we do that there.

markmandel:
> 
> I'm not too familiar, but when I was reading through it, all that should be required is having a step that runs `uname -m` and then matches on the string (because `uname -m` returns `arm64` while it needs to be `aarch64`)

That sound exactly right 👍🏻, then we need to pass that through as a Docker arg, and use that in appropriate places in the build Docker file. 

I'm sure other fun bits will come up, but that is the general gist.

markmandel:
The other option, as arch issues can get tricky fast, is to force Docker to Linux amd64 for all things:

`export DOCKER_DEFAULT_PLATFORM=linux/amd64`

Might actually be the easiest option (we could also bake this into the Makefile in some capacity).

XAMPPRocky:
TIL, I’ll try it with that tomorrow

markmandel:
Just tried to take it for a spin, but got stuck at the compilation issue I referenced above 😢 

I was going to note as well, if you have a local `kubectl` config that has access to an Agones cluster (although if you have occasionally access issues, try running a `kubectl` command to refresh the token), you should be able to run the e2e test directly, of the `IMAGE_TAG` env var is set.

XAMPPRocky:
Decided to just to revert removing the `quilkin.yaml` and changing the entrypoint for now, as I'd like to get these changes in and we can make those changes separately.

> The other option, as arch issues can get tricky fast, is to force Docker to Linux amd64 for all things:

This doesn't work because the `build-image` is hardcoded to use x86_64 in multiple places, so the build-image needs to be reworked before it's usable on ARM machines.

markmandel:
> Decided to just to revert removing the `quilkin.yaml` and changing the entrypoint for now, as I'd like to get these changes in and we can make those changes separately.

That makes sense 👍🏻

> 
> > The other option, as arch issues can get tricky fast, is to force Docker to Linux amd64 for all things:
> 
> This doesn't work because the `build-image` is hardcoded to use x86_64 in multiple places, so the build-image needs to be reworked before it's usable on ARM machines.

Huh. That's strange, because if you set `DOCKER_DEFAULT_PLATFORM` is affects both build and run operations, so the image itself will be x86, so having it install x86 binaries is what should work in that environment. 🤔 Prune the local image cache maybe, if there are old versions lying around?

I almost wish I had a ARM machine so I could test this out. If you are keen, could also try a GVC sometime. It's entirely possible I'm missing something, since i'm shooting in the dark a bit.

In the meantime, I'll take on finding what has broken in the integration tests for this PR, and update you with my results (probably send a PR to your PR.)

Also I'll prioritise #593 such that you can use the same image from CI to locally test an integration test.

src/cli.rs
```diff
@@ -18,38 +18,50 @@ mod generate_config_schema;
...
-    config: PathBuf,
-    #[clap(
-        short,
-        long,
-        env,
-        help = "Whether Quilkin will report any results to stdout/stderr."
-    )]
-    quiet: bool,
+    /// The path to the configuration file for the Quilkin instance.
+    #[clap(short, long, env = "QUILKIN_CONFIG", default_value = "quilkin.yaml")]
```
markmandel, line 44: I was actually thinking of coming back around and fleshing out the help for the rest of the commands for a nice command line reference.

Curious if you didn't want a command line reference here?

markmandel:
Digging into what's failing in the integration tests - not sure what's causing it, but looks like there is something trying to bind to 9091 more than once. I'm seeing this in a few of the tests.

This is in agones management mode:

```
─➤  kubectl logs -n 1663639419 quilkin-manage-agones-5fb9bf9df7-6rqm9
{"timestamp":"2022-09-20T02:09:39.317682Z","level":"INFO","fields":{"message":"Starting Quilkin","version":"0.4.0-dev","commit":"f45d971388ce356229cc922c36376f267d1e6d40"},"target":"quilkin::cli"}
{"timestamp":"2022-09-20T02:09:39.317882Z","level":"DEBUG","fields":{"message":"provided path not found","path":"quilkin.yaml"},"target":"quilkin::cli"}
{"timestamp":"2022-09-20T02:09:39.318230Z","level":"INFO","fields":{"message":"Starting admin endpoint","address":"0.0.0.0:9091"},"target":"quilkin::admin"}
{"timestamp":"2022-09-20T02:09:39.318298Z","level":"TRACE","fields":{"message":"registering event source with poller: token=Token(1), interests=READABLE | WRITABLE","log.target":"mio::poll","log.module_path":"mio::poll","log.file":"/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.8.4/src/poll.rs","log.line":521},"target":"mio::poll"}
{"timestamp":"2022-09-20T02:09:39.319482Z","level":"INFO","fields":{"message":"Starting admin endpoint","address":"0.0.0.0:9091"},"target":"quilkin::admin","span":{"name":"spawn"},"spans":[{"name":"spawn"}]}
thread 'main' panicked at 'error binding to 0.0.0.0:9091: error creating server listener: Address already in use (os error 98)', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.20/src/server/server.rs:77:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
{"timestamp":"2022-09-20T02:09:39.319629Z","level":"ERROR","fields":{"message":"Panic has occurred. Moving to Unhealthy","panic_info":"panicked at 'error binding to 0.0.0.0:9091: error creating server listener: Address already in use (os error 98)', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.20/src/server/server.rs:77:13"},"target":"quilkin::admin::health","span":{"name":"spawn"},"spans":[{"name":"spawn"}]}
{"timestamp":"2022-09-20T02:09:39.319646Z","level":"ERROR","fields":{"message":"Panic has occurred. Moving to Unhealthy","panic_info":"panicked at 'error binding to 0.0.0.0:9091: error creating server listener: Address already in use (os error 98)', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.20/src/server/server.rs:77:13"},"target":"quilkin::admin::health","span":{"name":"spawn"},"spans":[{"name":"spawn"}]}
{"timestamp":"2022-09-20T02:09:39.319634Z","level":"TRACE","fields":{"message":"no local config found, falling back to local in-cluster config","error":"failed to read kubeconfig from '\"/home/nonroot/.kube/config\"': No such file or directory (os error 2)"},"target":"kube_client::config"}
{"timestamp":"2022-09-20T02:09:39.320006Z","level":"TRACE","fields":{"message":"deregistering event source from poller","log.target":"mio::poll","log.module_path":"mio::poll","log.file":"/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.8.4/src/poll.rs","log.line":652},"target":"mio::poll"}
{"timestamp":"2022-09-20T02:09:39.320187Z","level":"DEBUG","fields":{"message":"buffer closing; waking pending tasks"},"target":"tower::buffer::worker"}
{"timestamp":"2022-09-20T02:09:39.320399Z","level":"TRACE","fields":{"message":"deregistering event source from poller","log.target":"mio::poll","log.module_path":"mio::poll","log.file":"/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/mio-0.8.4/src/poll.rs","log.line":652},"target":"mio::poll"}
```

(Side note - looks like the check in `create_quilkin_pod` shouldn't just check if it's running, because the pod is in crashloop backoff in the integration test).

Looks like all the tests are failing for this reason.

markmandel:
Aha! I can replicate. Running a local binary with a config file reproduces the error:

```
➜  agones-xonotic git:(pr/ep/run-args) ✗ ../../target/build-image/x86_64-unknown-linux-gnu/release/quilkin -c ./client-compress.yaml run
{"timestamp":"2022-09-20T02:26:50.508501Z","level":"INFO","fields":{"message":"Starting Quilkin","version":"0.4.0-dev","commit":"f45d971388ce356229cc922c36376f267d1e6d40"},"target":"quilkin::cli"}
{"timestamp":"2022-09-20T02:26:50.508851Z","level":"INFO","fields":{"message":"Starting admin endpoint","address":"0.0.0.0:9091"},"target":"quilkin::admin"}
{"timestamp":"2022-09-20T02:26:50.508956Z","level":"INFO","fields":{"message":"Starting","port":7000,"proxy_id":"markmandel"},"target":"quilkin::proxy"}
{"timestamp":"2022-09-20T02:26:50.508973Z","level":"INFO","fields":{"message":"Starting admin endpoint","address":"0.0.0.0:9091"},"target":"quilkin::admin"}
{"timestamp":"2022-09-20T02:26:50.509022Z","level":"ERROR","fields":{"message":"Panic has occurred. Moving to Unhealthy","panic_info":"panicked at 'error binding to 0.0.0.0:9091: error creating server listener: Address already in use (os error 98)', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.20/src/server/server.rs:77:13"},"target":"quilkin::admin::health"}
{"timestamp":"2022-09-20T02:26:50.509039Z","level":"ERROR","fields":{"message":"Panic has occurred. Moving to Unhealthy","panic_info":"panicked at 'error binding to 0.0.0.0:9091: error creating server listener: Address already in use (os error 98)', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.20/src/server/server.rs:77:13"},"target":"quilkin::admin::health"}
thread 'main' panicked at 'error binding to 0.0.0.0:9091: error creating server listener: Address already in use (os error 98)', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.20/src/server/server.rs:77:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtr
```

Looks like it's trying to recreate the admin server twice.

cloudbuild.yaml
```diff
@@ -46,16 +46,6 @@ steps:
     args:
       - build
     id: build
-  # Run the built images for 5 seconds to make sure that the entrypoint and default config works out of the box
```
markmandel, line 49: I replicated the issue as well with running the container from this PR. I'd recommend leaving in this test, as it also will replicate the breaking issue with the Agones integration tests, but output the logs within cloud build, and make it much easier to debug if there is a problem with the base image.

We may want to consider running the container in a few different common configuration options might also be useful in a few tests like the above, just to prove it all still runs even without any traffic flowing through.

src/cli.rs
markmandel: Oh wait - when I run the command, I see help everywhere, 

Wait.... does clap take the help from the comment above it? If so, that's a super neat trick.

markmandel:
Re-running, so that the new CI failure messages can give us a build image to debug against.

src/cli.rs
```diff
@@ -78,25 +90,68 @@ impl Cli {
...
+        let _admin_task = if self.no_admin {
+            config.admin.remove();
+            None
+        } else {
+            if let Some(address) = self.admin_address {
+                config
+                    .admin
+                    .store(Arc::new(crate::config::Admin { address }));
+            }
+            Some(tokio::spawn(crate::admin::server(config.clone())))
```
markmandel, line 103: Pretty sure this is the offending line causing the double creation of the admin server, since both `Proxy::run` and `ControlPlane::from_arc` both create admin instances and run them.

markmandel: Doing some debugging ([code](https://github.com/markmandel/quilkin/commit/09de12a7f906e9ea0451689d54b7ff0a40fdd8fc)), it looks like the `Config` that comes from this line:

https://github.com/markmandel/quilkin/commit/09de12a7f906e9ea0451689d54b7ff0a40fdd8fc#diff-b2812f19576dd53d0c35b107a322f58a00fce6977f4b5976e1961853982af3ccR93

```rust
let config = Arc::new(Self::read_config(self.config)?);
```

Has a `None` admin field, but looking at the code I'm not sure why it's `None` when it should have the default admin port 🤔 

You can see it in my debug logs:

```log
{"timestamp":"2022-09-22T04:41:50.829061Z","level":"DEBUG","fields":{"message":"Main Config","no_admin":false,"config":"Config { admin: None, clusters: None, filters: None, management_servers: None, proxy: None, version: None, metrics: Metrics { active_clusters: GenericGauge { v: Value { desc: Desc { fq_name: \"cluster_active\", help: \"Number of currently active clusters.\", const_label_pairs: [], variable_labels: [], id: 6550223101314893, dim_hash: 9568000379080501117 }, val: AtomicI64 { inner: 0 }, val_type: Gauge, label_pairs: [] } }, active_endpoints: GenericGauge { v: Value { desc: Desc { fq_name: \"cluster_active_endpoints\", help: \"Number of currently active endpoints.\", const_label_pairs: [], variable_labels: [], id: 9175583387471564798, dim_hash: 12641526325508385022 }, val: AtomicI64 { inner: 0 }, val_type: Gauge, label_pairs: [] } } } }"},"target":"quilkin::cli"}
```

I'll keep digging into this tomorrow unless you get to it first.

markmandel: I think I've found the issue. I'll send a PR to your PR today.

markmandel (approved):

Issue #546, opened 2022-07-14
Release 0.4.0 Requirements


markmandel:
Writing a list of issues, PR's and tasks to complete that are blocking a 0.4.0 release:

### Required

* [x] Merge of #527 (or what new PR it becomes) - closing #233
* [x] Do a review of the guide and make tickets on what has changed (logging, new xDS interface are a few I can think of), add any issues created from that review here as necessary.
* [x] #559
* [x] #605
* [x] #589
* [x] #609

### Potential / Nice to have

* [x] #541 (depending on how logging changes come about in above's xDS update)
* [x] #510 (or at least some work started here).
* [x] #608

Drop a comment below if you think we're missing anything else?

markmandel:
Working on a big review of the docs. Still WIP, but if you want to follow along, you can see a preview here: https://markmandel.github.io/quilkin/pr/docs/big-review/book/

markmandel:
Almost done with documentation review -- biggest task than I had anticipated. Hopefully complete today, time forgiving.

markmandel:
SO! 🥳 

We have everything we wanted to get done for the 0.4.0 release! Any objections to cutting a release?

Do we want to wait for #639 ? Or is that a forcing function to get another release done in December (and get more releases done in shorter periods) 😄

XAMPPRocky:
> Do we want to wait for https://github.com/googleforgames/quilkin/pull/639 ?

Let’s not wait. We can always release it later, and gives us more time to play with metadata references, for example, I’d like to make it consistent so any filter config you can insert a literal value into you can use a reference, and that’s not done yet.

markmandel:
Awesome, I'll get moving on getting a release done. Perfect timing, since I'm demo'ing this stuff to a bunch of people next week 😄

PR #576, opened 2022-08-25
Update configuration reference docs

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup

/kind documentation

> /kind feature
> /kind hotfix

Update the configuration reference docs to include `clusters` and `localities`.

Also cleans up a few smaller issues in the current reference that I found.

Work on #546

At some point we should find a better way to represent this reference, but this will do for now.


XAMPPRocky (approved):
LGTM, though we should probably make this auto-generated

PR #578, opened 2022-08-26
Move proxy::Server to Proxy and sessions::session to sessions

XAMPPRocky:
This just shuffles some of the proxy organisation around to remove a level a nesting and to use some clearer terms in the API. No actual behaviour changes, and the only public facing change is renaming `Server` to `Proxy`. I feel that's more clear because the component is doing a lot more than being a server and the project markets itself as a proxy so having a type called `Proxy` as the main unit makes sense to me.


bombsimon (approved):
This makes a lot of sense to me as well, LGTM!

PR #581, opened 2022-08-29
Relax trait bounds for Default, Deserialize, and Serialize for Slot<T>

XAMPPRocky:
This replaces the derived implementations with manual ones to relax the trait bounds on `Slot<T>`. This allows for example `Slot<T>` to implement `Default` even if `T` does not.


bombsimon (approved):
👍

PR #591, opened 2022-09-13
Add ready probe endpoint

### LINKED ISSUE #590: Readiness probe based on number of endpoints

Since we are putting Quilkin behind UDP Load Balancers in K8s, we should have a readiness endpoint, so that Services know if the proxy is ready to receive traffic or not.

This I feel is particularly true when scaling up a proxy that is connected to an xDS resource, but has yet to receive endpoint instructions yet.

I think the easiest option is to have a `/ready` admin http probe that returns 200 when the proxy has 1 or more endpoints configured.

If the proxy has no endpoints, it will return `StatusCode::INTERNAL_SERVER_ERROR`.

That way new Proxies that have yet to receive their endpoints from the xDS server won't get sent data that won't go anywhere.

Also if no proxies have endpoints, there's no point for the Service to send data to a proxy anyway, since it's not going anywhere.

We could leave things as is, and rely on the eventual consistency, but I think that will potentially put unnecessary delays in UDP network streams / connections.

**Additional context**

https://github.com/googleforgames/quilkin/blob/25d93603c7c83603def8e79d26a390d5dc83755f/src/admin.rs#L47-L71

XAMPPRocky:
Adds a k8s ready probe endpoint to check when a Quilkin instance is ready to receive traffic, which currently maps to whether it has any endpoints available to forward traffic to.

Closes #590

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -14,13 +14,16 @@
  *  limitations under the License.
  */
 
+mod health;
+
 use std::convert::Infallible;
 use std::sync::Arc;
 
 use hyper::service::{make_service_fn, service_fn};
 use hyper::{Body, Method, Request, Response, Server as HyperServer, StatusCode};
 
-use crate::{config::Config, proxy::Health};
+use self::health::Health;
+use crate::config::Config;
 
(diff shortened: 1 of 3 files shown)


bombsimon (approved):
LGTM! 🧑‍⚕️

PR #598, opened 2022-09-20
Allow gameservers to not have port

XAMPPRocky:
When gameservers are deleted, their port is removed. This fixes it so that the gameserver will have a port of zero instead of returning an error.

### DIFF
--- a/src/config/watch/agones/crd.rs
+++ b/src/config/watch/agones/crd.rs
@@ -270,7 +270,7 @@ impl TryFrom<GameServer> for Endpoint {
             .ports
             .as_ref()
             .and_then(|ports| ports.first().map(|status| status.port))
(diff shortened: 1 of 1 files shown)


bombsimon (approved):

markmandel:
🤔 on deletion I don't think we remove anything from the status part of the GameServer CRD (there's no reason to) in Agones. I should go double check though.

I wonder what is causing Quilkin to see this.

XAMPPRocky:
@markmandel Interesting, it doesn't happen all the time, it happens some of the time, I'll log the output and post here.

markmandel:
Sounds good.  Let's merge this though!

markmandel (approved):

markmandel:
Figure it would be good to have a fix in, and then we can also look to a potentially better solution over time.

Issue #586, opened 2022-09-07
Optionally specify the session key through a Filter


markmandel:
**Is your feature request related to a problem? Please describe.**

This is driven by the fact that legacy game servers utilise the connection address + port as information of a user connection, and will likely get confused by packets coming from multiple Quilkin proxies, since they will be seen as different users. Which makes it difficult to setup a horizontally scalable set of proxies in front of legacy game servers.

What I think we need is a way to be able to proxy incoming connections from a variety of sources, and specify how we want that to be turned into a Session Key, such that if there is identifiable information in the packet, it can be utilised as a Session Key, regardless of where it comes from.

**Describe the solution you'd like**

I visualise this like so:

```
                            Dynamic Load Balancer

                              ┌─────────┐
                              │         │
                              │         │
                      ┌───────► Quilkin │
┌─────────┐           │       │ Proxy   ├───────────┐      Converts incoming packets
│         │           │       │         │           │      into singular connections
│ Client  ├───────────┤       └─────────┘           │      per routing token.
│         │           │                             │      ┌───────────┐            ┌────────────┐
└─────────┘           │       ┌─────────┐           │      │           ├────────────►            │
                      │       │         │           │      │  Quilkin  │            │  DGS       │
                      │       │         │       ┌───┴──────►  Sidecar  │            │            │
                      ├──────►│ Quilkin ├───────┤          │           ├────────────►            │
┌─────────┐           │       │ Proxy   │       │          └───────────┘            └────────────┘
│         │           │       │         │       │
│ Client  ├───────────┤       └─────────┘       │                        Two discreet connections
│         │           │                         │                        Since there are two clients
└─────────┘           │       ┌─────────┐       │
                      │       │         │       │
                      │       │         │       │
                      └──────►│ Quilkin ├───────┘
                              │ Proxy   │
                              │         │
                              └─────────┘
```

So the new part would be the sidecar to pull back all the routed tokens from the scalable proxies into the expected set of connections.

To do this, I would recommend adding a `session_key: Option<String>` to `ReadResponse`
https://github.com/googleforgames/quilkin/blob/4800db48b5da64df227e4368ee485b4d6a89ad85/src/filters/read.rs#L81-L88

If set, that would be the session key utilised when creating and looking up the session. If not set, we use the default we use now.

We could then write a `SessionKey` filter (or some better name 😄 ) that could use the token stored in dynamic metadata to provide the session key.

My theory being, that an implementation of the above sidecar would be the following config:

[...]

PR #604, opened 2022-09-26
Add ASN maxmind database integration

### LINKED ISSUE #450: More in-depth network metrics about clients (IPv4 and IPv6)
**Introduction**

Quilkin will do many things at Embark for our games. One of the things we want to use Quilkin for is re-routing players to specific proxies or black-list specific proxies for specific players when we notice that there is broken routing, or just poor performance to a specific proxy. 

In order to make good decisions you will need some more metrics to base these decisions on instead of just having IP addresses to base them on. In our case we would like to have **AS-number**, **AS-name**, **AS-location**, **IPv(4/6)-prefix**, **Prefix Owner** for each IP address that establishes a game session with us.

With this information we can then build tools and heatmaps which will show where players are having poor connection performance or quality to our infrastrastructure. 

**Proposed solution**

I myself recently came across [IPNetDB](https://ipnetdb.com) which gave fire to this idea I have been thinking about for some time. Essentially I would like Quilkin to have support for doing a query for each connection when established, getting specific information about the IP address and then have this sent with other metrics downstream. 

IPNetDB in this case is based on MaxMind Data Base (`mmdb`) but support should probably also include GeoIP2 and GeoLite2 databases. There are librarys in rust for this already, for instance [oschwald/maxminddb-rust](https://github.com/oschwald/maxminddb-rust). 

Below you can see output from a query of my home IP address. The information we would need is all in the query. 

```
{'allocation': '37.247.0.0/19',
 'allocation_cc': 'SE',
 'allocation_registry': 'ripe',
 'allocation_status': 'allocated',
 'as': 33885,
 'as_cc': 'SE',
 'as_entity': 'Ownit Broadband AB',
 'as_name': 'OWNIT',
 'as_private': False,
 'as_registry': 'ripe',
 'prefix': '37.247.0.0/19',
 'prefix_asset': [],
 'prefix_assignment': 'allocated pa',
 'prefix_bogon': False,
 'prefix_entity': 'Ownit Broadband AB',
 'prefix_name': 'SE-OWNIT-20120417',
 'prefix_origins': [33885],
 'prefix_registry': 'ripe'}
```

**Problems?**
Since these databases are file based, we would need to figure out a solution where Quilkin uses and updates the file on disk in order to have the most up to date version of the database.

XAMPPRocky:
This adds a Maxmind DB integration so that if you provide quilkin with the source of a maxmind database that matches the structure of IPNetDB, it will output info about the ASN of a new session source.

I also refactored the admin server to be in `cli` to reduce duplication between run and manage.

closes #450

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -25,7 +25,7 @@ use std::{
 
 use tokio::{signal, sync::watch};
 
-use crate::Config;
+use crate::{admin::Mode, Config};
 
 pub use self::{
     generate_config_schema::GenerateConfigSchema,
@@ -40,15 +40,15 @@ const ETC_CONFIG_PATH: &str = "/etc/quilkin/quilkin.yaml";
 #[derive(clap::Parser)]
 #[non_exhaustive]
 pub struct Cli {
+    /// Whether to spawn the admin server or not.
+    #[clap(default_value = "true", env, long)]
+    pub admin: bool,
     /// The path to the configuration file for the Quilkin instance.
     #[clap(short, long, env = "QUILKIN_CONFIG", default_value = "quilkin.yaml")]
     pub config: PathBuf,
     /// The port to bind for the admin server
     #[clap(long, env = "QUILKIN_ADMIN_ADDRESS")]
     pub admin_address: Option<std::net::SocketAddr>,
-    /// Whether to spawn the admin server or not.
-    #[clap(long, env = "QUILKIN_NO_ADMIN")]
-    pub no_admin: bool,
     /// Whether Quilkin will report any results to stdout/stderr.
     #[clap(short, long, env)]
     pub quiet: bool,
@@ -64,9 +64,20 @@ pub enum Commands {
     Manage(Manage),
 }
 
+impl Commands {
+    pub fn admin_mode(&self) -> Option<Mode> {
+        match self {
+            Self::Run(_) => Some(Mode::Proxy),
+            Self::Manage(_) => Some(Mode::Xds),
+            Self::GenerateConfigSchema(_) => None,
+        }
+    }
+}
+
 impl Cli {
     /// Drives the main quilkin application lifecycle using the command line
     /// arguments.
+    #[tracing::instrument(skip_all)]
     pub async fn drive(self) -> crate::Result<()> {
         let version: std::borrow::Cow<'static, str> = if cfg!(debug_assertions) {
             format!("{VERSION}+debug").into()
@@ -80,6 +91,7 @@ impl Cli {
                 .from_env_lossy();
             tracing_subscriber::fmt()
                 .json()
+                .with_file(true)
                 .with_env_filter(env_filter)
                 .init();
         }
@@ -91,12 +103,16 @@ impl Cli {
         );
 
         let config = Arc::new(Self::read_config(self.config)?);
-        if self.no_admin {
+        let _admin_task = if let Some(mode) = self.command.admin_mode().filter(|_| self.admin) {
+            if let Some(address) = self.admin_address {
+                config
+                    .admin
+                    .store(Arc::new(crate::config::Admin { address }));
+            }
+            Some(tokio::spawn(crate::admin::server(mode, config.clone())))
+        } else {
             config.admin.remove();
-        } else if let Some(address) = self.admin_address {
-            config
-                .admin
-                .store(Arc::new(crate::config::Admin { address }));
+            None
         };
 
         let (shutdown_tx, shutdown_rx) = watch::channel::<()>(());
--- a/src/config.rs
+++ b/src/config.rs
@@ -49,9 +49,12 @@ base64_serde_type!(Base64Standard, base64::STANDARD);
 // For some log messages on the hot path (potentially per-packet), we log 1 out
 // of every `LOG_SAMPLING_RATE` occurrences to avoid spamming the logs.
 pub(crate) const LOG_SAMPLING_RATE: u64 = 1000;
+pub(crate) const BACKOFF_INITIAL_DELAY_MILLISECONDS: u64 = 500;
+pub(crate) const BACKOFF_MAX_DELAY_SECONDS: u64 = 30;
+pub(crate) const BACKOFF_MAX_JITTER_MILLISECONDS: u64 = 2000;
 
 /// Config is the configuration of a proxy
-#[derive(Clone, Debug, Default, Deserialize, Serialize, JsonSchema)]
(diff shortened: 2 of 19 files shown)


markmandel:
Looks like it's unhappy about the MPL, if you add in the exception here:

https://github.com/googleforgames/quilkin/blob/6a42df8f7a20d6a5547ccd23834989d584ce72e5/deny.toml#L30-L33

Please also set `dependecies` in  

https://github.com/googleforgames/quilkin/blob/6a42df8f7a20d6a5547ccd23834989d584ce72e5/image/archive_dependencies.sh#L29

to: `(webpki-roots)` so we capture the source in the image.

src/cli.rs
```diff
@@ -91,12 +103,16 @@ impl Cli {
         );
 
         let config = Arc::new(Self::read_config(self.config)?);
-        if self.no_admin {
+        let _admin_task = if let Some(mode) = self.command.admin_mode().filter(|_| self.admin) {
```
markmandel, line 106: Nice centralisation 👍🏻

src/config.rs
```diff
@@ -195,6 +198,21 @@ impl Config {
...
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            admin: Slot::with_default(),
+            clusters: <_>::default(),
+            filters: <_>::default(),
+            management_servers: <_>::default(),
+            proxy: <_>::default(),
+            version: <_>::default(),
+            maxmind_db: Slot::empty(),
```
markmandel, line 210: Now I'm curious, is there a publicly accessible maxmind db somewhere we could reference by default? 🤔  or licencing / security nightmare?

src/proxy/sessions.rs
```diff
@@ -127,14 +127,33 @@ impl Session {
...
-            source: args.source,
+            source: args.source.clone(),
             dest: args.dest,
             created_at: Instant::now(),
             expiration,
             shutdown_tx,
         };
         tracing::debug!(source = %s.source, dest = ?s.dest, "Session created");
 
+        if let Some(mmdb) = &*crate::MaxmindDb::instance() {
```
markmandel, line 138: Looks good - just so I'm super clear, we're only tracing this information on Session creation? (Which means this can't be a filter either, which is fair enough. i did wonder at one point if #586 should maybe have a "Session creation Filter/Plugin" - which this could have leaned into, but not sure if that's a good idea).

Do you think we'll want to also attach this information to any other traces that may come through this particular session?

Only other thought: We need some docs please 😃

src/proxy/sessions/manager.rs
```diff
@@ -98,6 +97,7 @@ impl SessionManager {
             .count();
 
         if expired_keys != 0 {
+            tracing::debug!("pruning expired sessions");
```
markmandel, line 100: Nice cleanup.

src/config.rs
XAMPPRocky: There might be, but I haven't one. The only one I know of is IPNetDB, which is $300/a year.

markmandel (approved):

src/config.rs
meeb: Hi, I stumbled over this discussion with an alert. I'm a founder of the company that releases IPNetDB. If helpful, as this is an open source project you're welcome to build against, test and develop with IPNetDB databases without paying anything and you can redistribute it with an open source project just with attribution. The paid licence is only required if you launched a for-profit service with an IPNetDB database embedded in it.  You're welcome to ask if you want clarification on any usage. The licence is not particularly restrictive.

We also have a mock mmdb database we use for testing internally which I can send over if you would prefer something to test with that has absolutely no licence but the same encoding format.

XAMPPRocky: @meeb Thank you for your response! To clarify, would a paid licence be required if this open source project embedded with IPNetDB communicated with a for-profit service over the network? This project is a UDP proxy primarily for games, which are often commercial projects. So long as that use case can make use of IPNetDB without the paid licence, I wouldn’t have a problem integrating with it directly (though it also needs @markmandel approval). I just wouldn’t want the primary use-case of the project to have surprising licencing for our users.

meeb: As it currently stands with the current bare-bones licence options, maybe if the end result was a paid-for SaaS or PaaS. As Quilkin sits between client applications and game servers (a similar comparable service architecture would be HAProxy but for UDP game traffic?) would I be correct in saying what IPNetDB would do in for this project is add additional metadata to connections as they pass through the proxy?

Having said that, we release IPNetDB for people to use so I'm happy to have a discussion about your specific use case. We have organised redistribution licences, as well as granted specific use licences to projects in the past.

There are other considerations as well such as the database updates weekly which given the fluid nature of the internet some updates will likely be required. If auto-updating your proxy would download from our CDN which creates logs with client IPs in etc. so there might be some further legalese to approve if you do embed IPNetDB with an auto-updating feature.

Given this is likely going somewhat off topic for the purpose of this PR you're welcome to email me at licence@ipnetdb.com with a mention you're XAMPPRocky (or markmandel) from GitHub and I'll pick it up. I can also send you a small test mmdb file you can embed for your test suite as well if you like.

markmandel: @meeb thanks so much for the offer!

A good first step for us would be to get our Open Source Programs Office to review the licences that are publicly available and get their approval. @meeb - is this the [correct link](https://ipnetdb.com/#:~:text=database%20creation%20libraries.-,Licencing,-Free%20for%20open), or is there something more comprehensive we should be looking at?

meeb: @markmandel yes those are the minimal licences. They're specifically brief to make it easy to use the database, however I would suspect you might want a slightly more robust licence if you wanted to redistribute the embedded database. Happy to chat to your OSPO. Given we allow free download of the databases the licence was always more trust based so we didn't bother with anything extensive to agree to. Assuming the use remains in an open source project I would suspect a simple attribution (and probably letting your users know their client may download data from our CDN periodically if the feature is enabled) would be sufficient. Feel free to pass on the above email address to anyone relevant internally to discuss it.

markmandel: Sweet - thanks - will pass this along and swing back around when I get a resolution.

src/proxy/sessions.rs
XAMPPRocky: Yeah for now I don't see much point in having it as a filter, as it's something that you want to only happen to successful connections, which is easy to guarentee here. We can always change it later.

> Do you think we'll want to also attach this information to any other traces that may come through this particular session?

I think we can if we find it useful. Right now I'm just starting with Quilkin outputting this info and then trying to use it, and seeing what it needs from there.

> Only other thought: We need some docs please 😃

Added

PR #613, opened 2022-09-30
Refactor session metrics and errors

XAMPPRocky:
Refactors session metrics to use static variables, this saves us 80 bytes per session (plus some more from being cloned down from proxy). The one downside is that I had to remove one test, because using static variables for metrics doesn't work well the with multithreaded tests. But the test was simple enough that I think it's worth losing that small bit of coverage.

It also refactors session's errors to be more helpful, and include more information when logged. I did this through a through a trait so that later we can make it easier to log other errors and types without duplication. I also refactored the branches in the packet send code to ensure that `PACKETS_DROPPED` is always being incremented when we don't send packets.


tests/metrics.rs
```diff
@@ -70,5 +70,5 @@ async fn metrics_server() {
 
     let response = String::from_utf8(resp.to_vec()).unwrap();
     dbg!(&response);
-    assert!(response.contains("quilkin_session_tx_packets_total 1"));
+    assert!(response.contains(r#"quilkin_packets_total{event="read"} 2"#));
```
XAMPPRocky, line 73: This is now 2 because this test uses a two Quilkin proxies.

markmandel (approved):
LGTM. Agreed it's a shame that we lose the coverage, but it's much cleaner, and will save us on passing through the date each time.

Not from this PR, but just making a note for later: `packets_dropped` should probably be `packets_dropped_total`, same with `packets_size` - just to have consistent naming (and I'm fairly sure that's the nomenclature for metrics).

Issue #605, opened 2022-09-26
`quilkin_session_rx_bytes_total` metric seems to not be working


XAMPPRocky:
I've noticed on the dashboard that this metric seems to not be working at all at them moment while `quilkin_session_tx_bytes_total` seems to be working just fine. Should investigate why, and add a test to prevent it from regressing.

markmandel:
Ah yep - I saw this on the dashboard as well! I did wonder if was a dashboard issue or something else, but I hadn't gone back to have a look since #595 landed.

I wonder if I accidentally dropped it during #543

PR #619, opened 2022-10-11
Remove duplicate metric

XAMPPRocky:
`PACKETS_SIZE` was the same as `BYTES_TOTAL`, but only for one direction. This fixes it to always be `BYTES_TOTAL` for both directions.


markmandel (approved):
Good catch!

markmandel:
Does this fix #605 ?

PR #621, opened 2022-10-12
Added ASN and Prefix labels to active sessions

XAMPPRocky:
This PR adds ASN and IP prefixes as labels to active sessions so that you can group active connections by ASN entity.

### DIFF
--- a/src/proxy/sessions.rs
+++ b/src/proxy/sessions.rs
@@ -58,6 +58,8 @@ pub struct Session {
     expiration: Arc<AtomicU64>,
     /// a channel to broadcast on if we are shutting down this Session
     shutdown_tx: watch::Sender<()>,
+    /// The ASN information.
+    asn_info: Option<crate::maxmind_db::IpNetEntry>,
 }
 
 // A (source, destination) address pair that uniquely identifies a session.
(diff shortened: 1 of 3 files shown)


src/proxy/sessions.rs
```diff
@@ -223,6 +210,23 @@ impl Session {
         }
     }
 
+    fn active_session_metric(&self) -> prometheus::IntGauge {
+        let labels = self
+            .asn_info
+            .as_ref()
+            .map(|asn| [asn.r#as.to_string(), asn.prefix.clone()])
```
markmandel, line 217: Is there a known set/size of prefixes and ASNs could be? I went to look it up, but couldn't find anything specific.

Trying to avoid a [cardinality explosion](https://www.robustperception.io/cardinality-is-key/)

XAMPPRocky: There is a theorhetical very high cardinality, as the max number of ASN's is `u16::MAX`, the maximum number of IPv4 prefixes is `u32::MAX`, and the IPv6 addresses is `u128::MAX`. Practically though I think the cardinality would be low (though it would definitely still be the highest cardinality metric). Because the prefixes aren't single IP addresses, they're blocks of IP addresses.

markmandel: Makes sense - and also (at least for now) it's likely to get traffic only from players in a limited set of regions, so that limits things as well.

markmandel (approved):

PR #623, opened 2022-10-13
Flatten proxy config section into root

XAMPPRocky:
This makes it clear that `port` and `id` can refer a management server's port and id, not just a proxy's. Also added management server's `port` flag.

### DIFF
--- a/agones/src/xds.rs
+++ b/agones/src/xds.rs
@@ -322,6 +322,8 @@ filters:
         let pp = PostParams::default();
         let config = r#"
 version: v1alpha1
+admin:
+    address: 0.0.0.0:9092
 management_servers:
   - address: http://quilkin-manage-agones:80
 "#;
--- a/agones/src/xds.rs
+++ b/agones/src/xds.rs
@@ -67,6 +67,7 @@ mod tests {
 
         let config = r#"
 version: v1alpha1
+admin: null
 filters:
   - name: quilkin.filters.capture.v1alpha1.Capture # Capture and remove the authentication token
     config:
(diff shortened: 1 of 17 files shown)


docs/src/proxy-configuration.md
```diff
@@ -18,21 +18,16 @@ properties:
...
           An identifier for the proxy instance.
-        default: On linux, the machine hostname is used as default. On all other platforms a UUID is generated for the proxy.
-      port:
-        type: integer
-        description: |
+      default: On linux, the machine hostname is used as default. On all other platforms a UUID is generated for the proxy.
+  port:
+      type: integer
+      description: |
           The listening port for the proxy.
```
markmandel, line 29:
```suggestion
          The listening port. In "proxy" mode, the port for traffic to be sent to. In "manage" mode, the port to connect to the xDS API.
```

Something like that, since it's both?

markmandel:
I got curious, so I took the PR for a spin to see what the issue was - I got back the following log from a failing Deployment of the management server:

```
➜  build git:(main) kubectl logs -n 1665779019 quilkin-manage-agones-8589cf77db-k7pcj
error: The following required arguments were not provided:
    --port <PORT>

USAGE:
    quilkin manage --port <PORT> <SUBCOMMAND>

For more information try --help
```

Looks like there's a bug where `--port` is now required for the management server?

The Args passed to quilkin in the Deployment that fails are:

```
    Args:
      manage
      agones
      --config-namespace
      1665779019
      --gameservers-namespace
      1665779019
```

Hope that helps!

agones/src/xds.rs
```diff
@@ -67,6 +67,7 @@ mod tests {
 
         let config = r#"
 version: v1alpha1
+admin: null
```
markmandel, line 70: `admin: null` shouldn't be required. Are you seeing it fail without this line?

markmandel (approved):

PR #624, opened 2022-10-14
Fix not receiving stream pushes

XAMPPRocky:
This is a smaller and better fix for the streams not always being established.

It's still not a full fix which know what issue is (the requests are sent before the stream is ready), and am working on PR for it, but haven't been able to make it pass tests yet.

### DIFF
--- a/src/xds/server.rs
+++ b/src/xds/server.rs
@@ -55,13 +55,16 @@ pub struct ControlPlane {
 
 struct Watchers {
     sender: tokio::sync::watch::Sender<()>,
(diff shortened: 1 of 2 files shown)


markmandel (approved):

src/xds/server.rs
```diff
@@ -173,8 +176,10 @@ impl ControlPlane {
             yield response;
 
             loop {
+                tracing::trace!("stream loop");
```
markmandel, line 179: Just double checking that you wanted to leave these `trace!` statements in here before merging.

Otherwise, merge when ready 👍🏻

Issue #456, opened 2021-12-16
how to use this proxy for call of duty games ?


isslam95:
hello,

can anyone please tell me how to use it on call of duty that uses port 3074 udp , I' m using ubuntu as router

XAMPPRocky:
Thank you for your issue! However this a reverse proxy for running your own game servers, not game clients and it seems like that’s what you’re trying to do. So this project wouldn’t work for you.

PR #625, opened 2022-10-18
Fixes for example Grafana dashboard

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

With all the renaming of the metrics, the example dashboard was broken.

This updates the queries such that it works again!

Work for #546

Noting that as part of #456 we'll need to update the metrics docs.


XAMPPRocky (approved):

XAMPPRocky:
LGTM, I've been meaning to upstream the dashboard we have internally since it has panels that use the ASN information.

PR #627, opened 2022-10-20
Add timestamp filter

XAMPPRocky:
This adds a new timestamp filter that reads in a metadata value as a unix timestamp, and then observes the duration between then and now in a histogram metric that's keyed by the direction and metadata key. This is mainly useful in combination with things like `Capture` which then allows you to parse timestamps from packets.

### DIFF
--- a/src/filters/timestamp.rs
+++ b/src/filters/timestamp.rs
@@ -1,0 +1,234 @@
+/*
+ * Copyright 2022 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


markmandel (commented):
This is awesome 😃 only thing missing is some docs please 👍🏻

src/filters/timestamp.rs
```diff
@@ -0,0 +1,234 @@
...
+                return;
+            }
+        };
+
+        // Create a normal DateTime from the NaiveDateTime
+        let datetime: DateTime<Utc> = DateTime::from_utc(naive, Utc);
+
+        let now = Utc::now();
+        let seconds = now.signed_duration_since(datetime).num_seconds();
+        self.metric(direction_label).observe(seconds as f64);
```
markmandel, line 96: (non blocking question): Does it make sense to combine this with maxmind information somehow? Either in the label or in tracing information?

Just thinking about tracking ping round trip times using this filter, and wanting some granularity on where geographically people are at?

markmandel (approved):

src/filters/timestamp.rs
XAMPPRocky: > Does it make sense to combine this with maxmind information somehow? Either in the label or in tracing information?

Yes it does, I'm going to try to find a way to do purely through grafana before trying to integrate it directly in code.

markmandel: Make sense! 👍🏻

PR #630, opened 2022-10-25
Refactor filter model from Context->Response into ref mut Context

XAMPPRocky:
This moves the filter model from converting `Context`s into `Response` to modifying the context in place. This removes a lot of conversion code in the binary and anecdotally I noticed minor improvements in performance just from using it like this.


markmandel (approved):
Love it. Everything is much cleaner 👍🏻

PR #633, opened 2022-10-26
Refactor metrics statics into free fns

XAMPPRocky:
This wraps a function around the metric static variables, this is to ensure at compile-time that all metrics always have the correct number of label values, previously there were one or two cases where one label was missing.

### DIFF
--- a/src/filters/chain.rs
+++ b/src/filters/chain.rs
@@ -258,9 +258,7 @@ impl Filter for FilterChain {
                     }
                     None => {
                         tracing::trace!(%id, "read dropping packet");
-                        crate::metrics::PACKETS_DROPPED
-                            .with_label_values(&[crate::metrics::READ_DIRECTION_LABEL, id])
-                            .inc();
+                        crate::metrics::packets_dropped(crate::metrics::READ, id).inc();
                         return None;
                     }
                 }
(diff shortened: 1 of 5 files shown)


src/metrics.rs
```diff
@@ -55,66 +59,103 @@ pub(crate) const BUCKET_FACTOR: f64 = 2.0;
...
+            registry(),
+        }
+        .unwrap()
+    });
+
+    PROCESSING_TIME.with_label_values(&[direction.label()])
+}
+
+pub fn bytes_total(direction: Direction) -> IntCounter {
+    pub(crate) static BYTES_TOTAL: Lazy<IntCounterVec> = Lazy::new(|| {
```
markmandel, line 97: Curious why these are `pub(crate)` - is that just leftover from the previous declarations?

src/filters/chain.rs
```diff
@@ -258,9 +258,7 @@ impl Filter for FilterChain {
                     }
                     None => {
                         tracing::trace!(%id, "read dropping packet");
-                        crate::metrics::PACKETS_DROPPED
-                            .with_label_values(&[crate::metrics::READ_DIRECTION_LABEL, id])
-                            .inc();
+                        crate::metrics::packets_dropped(crate::metrics::READ, id).inc();
```
markmandel, line 261: Love this. Much better 👍🏻

markmandel (approved):

PR #634, opened 2022-10-26
Try to recover when the main process fails

### LINKED ISSUE #626: Fatal error from too old resource version
Spotted earlier today. This should probably not be a fatal error, and instead be made recoverable.

```
error returned by apiserver during watch: too old resource version: 26934269 (26934894)
```

XAMPPRocky:
Fixes spurious errors like #626 

fixes #626

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -57,7 +57,7 @@ pub struct Cli {
 }
 
 /// The various Quilkin commands.
-#[derive(clap::Subcommand)]
+#[derive(Clone, clap::Subcommand)]
 pub enum Commands {
     Run(Run),
     GenerateConfigSchema(GenerateConfigSchema),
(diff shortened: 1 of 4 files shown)


src/cli.rs
```diff
@@ -137,16 +137,26 @@ impl Cli {
...
 
-        let fut = match self.command {
-            Commands::Run(runner) => tokio::spawn({
-                let shutdown_rx = shutdown_rx.clone();
-                async move { runner.run(config, shutdown_rx).await }
-            }),
-            Commands::Manage(manager) => tokio::spawn(async move { manager.manage(config).await }),
-            Commands::GenerateConfigSchema(generator) => {
-                tokio::spawn(std::future::ready(generator.generate_config_schema()))
+        let fut = tryhard::retry_fn({
```
markmandel, line 140: Shall we warn a log on retry?

markmandel (approved):

PR #638, opened 2022-10-31
Implement string interning and symbol resolution for metadata keys

XAMPPRocky:
This PR adds string interning and symbol resolution for metadata keys, this lowers memory usage for strings, and moves from needing `Arc` to just having a 32 bit id. This also adds support for resolving values to a metadata value in configuration. This allows you to use dynamic metadata values in configuration in places where a static literal is currently allowed.

### DIFF
--- a/src/metadata.rs
+++ b/src/metadata.rs
@@ -14,17 +14,21 @@
  * limitations under the License.
  */
 
-use std::{collections::HashMap, convert::TryFrom, sync::Arc};
-
-use crate::xds::config::core::v3::Metadata as ProtoMetadata;
+pub(crate) mod symbol;
 
 #[doc(hidden)]
 pub mod build {
     include!(concat!(env!("OUT_DIR"), "/built.rs"));
 }
 
+use std::{collections::HashMap, convert::TryFrom};
+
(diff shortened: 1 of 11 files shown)


src/metadata.rs
```diff
@@ -14,17 +14,21 @@
...
 #[doc(hidden)]
 pub mod build {
     include!(concat!(env!("OUT_DIR"), "/built.rs"));
 }
 
+use std::{collections::HashMap, convert::TryFrom};
+
+use crate::xds::config::core::v3::Metadata as ProtoMetadata;
+
+pub use symbol::{Key, Reference, Symbol};
```
markmandel, line 28: Just checking if I'm missing something, but are Reference/Symbol used anywhere yet?

XAMPPRocky: No they aren't, they will be used in a PR I'll put up today that updates `ConcatencateBytes` to accept `Symbol`s instead of just raw bytes, letting you accept bytes from metadata.

markmandel: Gotcha - that makes sense. Was going to ask for docs, but if it's not actually being used yet, we can hold off on that 👍🏻

PR #639, opened 2022-11-02
Refactor concatenate to accept metadata

XAMPPRocky:
This builds on the previous PR and allows `Concatenate` to accept references metadata to insert at execution time in the pipeline. This also adds computed properties for `Metadata` with the current and only one being `timestamp/now` which returns the current time in nanoseconds.


markmandel:
Something went weird with the CLA bot yesterday, but I think it's fixed now.

markmandel:
Ooh, this is slick. Did you want me to take a look at the failing Agones test? (Not sure if you have that up and running locally or not).

docs/src/filters.md
```diff
@@ -139,6 +139,20 @@ properties:
...
+anywhere you can supply a literal value you can also supply a reference to a
+piece of metadata. To do so prefix the id of the metadata with `$`
+(e.g. `$example.com/foo`). This will insert the value of the metadata from at
+that point in the pipeline for the configuration.
+
+There are also computed symbols under `quilkin.dev/computed` that will return the result of an operation when referenced (e.g. `$quilkin.dev/computed/timestamp/now`).
+
+| id | description |
+|----|-------------|
+| `/timestamp/now` | Returns the current time in nanoseconds |
```
markmandel, line 154: I'm assuming this likely used with the `Timestamp` filter, probably Capture / Concat as well - it would be good for onboarding to provide some examples of using this as a filter.

Otherwise we don't have an example of how to use $reference with a Filter along with a use case.

markmandel (commented):
This is cool! Code looks good (conflicts need resolving), just would love some examples with a few Filters' documentation (which we could link to from "Filter Metadata Data Model"), so that users can see a working example.

XAMPPRocky:
>  just would love some examples with a few Filters' documentation (which we could link to from "Filter Metadata Data Model"), so that users can see a working example.

Will do, currently testing it out in production first.

XAMPPRocky:
Going to close this since it's so out of date, if we still want it we can add it back as a new PR.

PR #650, opened 2022-11-18
Remove map_proto_enum and simplify proto enum conversions

XAMPPRocky:
We had quite a bit of unneeded code for converting from proto enums, this PR removes that.


src/filters/compress/config.rs
```diff
@@ -93,6 +98,16 @@ impl From<Action> for ProtoAction {
     }
 }
 
+impl From<ProtoAction> for Action {
+    fn from(action: ProtoAction) -> Self {
+        match action {
+            ProtoAction::DoNothing => Self::DoNothing,
+            ProtoAction::Compress => Self::Compress,
+            ProtoAction::Decompress => Self::Decompress,
```
markmandel, line 106: Well that's just so much easier, isn't it 😄 Nice.

markmandel (approved):

PR #651, opened 2022-11-18
Update dependencies

XAMPPRocky:


XAMPPRocky:
@markmandel Would you be able to help me out with this? No matter what I try, I can't get a protobuf compiler visible for both the main build and example project, and there seem to be new problems even building from source.

markmandel:
@XAMPPRocky heya! Back from time off - lemme take a look 👍🏻

markmandel:
Digging in a little - looks like `test-examples` is working (can confirm locally as well) - so you definitely got past hurdle No. 1.

The next thing that's failing is that it's not cross compiling to mac for us. Something is failing in the compilation of protoc on the `rust-linux-darwin-builder` image. I'm digging into it.

I'm seeing what I can find, but have also filed https://github.com/joseluisq/rust-linux-darwin-builder/issues/20 to see if anyone else has any ideas 🤔

markmandel:
....I have an idea 😄 

Adding a feature so that if you have a local protoc, it won't vendor. I think this will work.

markmandel:
Submitted #653 - I think this will solve the issue! (wil check if CI passes when I wake up tomorrow).

markmandel (approved):

PR #653, opened 2022-11-30
Fix for #651 vendor-protoc feature

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Implement `vendor-protoc` feature for if protoc is installed locally.

For `build-macos-binary` rather than vendor, since it fails to compile, install protoc from release.

N/A

Update to https://github.com/googleforgames/quilkin/pull/651 which should fix the issues with tests. Would love thoughts on the approach.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,7 +106,9 @@ pretty_assertions = "1.3.0"
 tonic-build = { version = "0.8.2", default_features = false, features = ["transport", "prost"] }
 prost-build = "0.11.2"
 built = { version = "0.5.1", features = ["git2"] }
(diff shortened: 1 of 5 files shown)


markmandel:
Doh. Looks like there's still some things to fix in `test-agones`. Will look tomorrow, unless @XAMPPRocky you want to take it on 🙂

XAMPPRocky (approved):

markmandel:
Got it. Out of sync `kube` dependency between the top level Cargo.toml and the agones Cargo.toml.

PR #657, opened 2022-12-05
Add Pprof profiling endpoint

XAMPPRocky:
This adds a proof profiling endpoint for Quilkin, allowing Quilkin to be continuously profiled using tools like Grafana Phlare.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,43 +53,47 @@ arc-swap = { version = "1.6.0", features = ["serde"] }
 async-stream = "0.3.5"
 base64.workspace = true
 base64-serde = "0.7.0"
-bytes = { version = "1.4.0", features = ["serde"] }
-cached = "0.45.0"
-chrono = "0.4.28"
-clap = { version = "4.4.2", features = ["cargo", "derive", "env"] }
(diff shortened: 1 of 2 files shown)


markmandel (approved):

markmandel:
Looks like a formatting issue, but otherwise should be fine.

Also thinking because of this PR, we should get another release out this month.

XAMPPRocky:
Yeah, though I need to make some changes as with the current implementation, we run out of memory pretty quickly right now, so I need to add some limits to stop that.

markmandel:
Oops, that's no good 😄 

I did have couple more thoughts this morning:

1. Should we have some kind of admin option to turn this off?
2. We should also document this somewhere 😁

markmandel:
Just watching an amazing talk about continuous profiling at GDC -- rather than baking in profiling, instead we could look at a eBPF profiler such as https://pyroscope.io/ to run low overhead continuous profiling on both GameServers and Quilkin itself.

I believe the only requirements are to have debug symbols (in some capacity) and frame pointers (I have not looked into details, just writing notes post talk).

XAMPPRocky:
Closed in favour of #875

PR #661, opened 2022-12-08
Rename run command to proxy

XAMPPRocky:
This PR changes the name of the run command to `proxy`. I think this makes it more clear what's being *run* than just run itself.

PR #662, opened 2022-12-10
Move xDS task out of Proxy, add Agones as another discovery task

XAMPPRocky:
This PR removes the requirement of using management server when in a single cluster. Now Quilkin can now watch for Agones changes in the proxy instance directly. This also moves the xDS client from being inside the proxy, to be run alongside it as a configuration discovery provider along with agones. This layout also opens the pathway in the future to support other configuration discovery providers such as Docker or Consul, if desired.

With #661 you'd now write the proxy as the following for agones and xDS

```
quilkin proxy agones
```

```
quilkin proxy xds http://quilkin-management-server
```

---

I'll update the documentation once CI passes.

### DIFF
--- a/agones/src/xds.rs
+++ b/agones/src/xds.rs
@@ -345,9 +345,14 @@ filters:
         let mut container = quilkin_container(
             client,
             Some(vec![
-                "proxy".into(),
-                "--management-server=http://quilkin-manage-agones:7800".into(),
-            ]),
+                            "proxy".into(),
+            <<<<<<< HEAD
+                            "--management-server=http://quilkin-manage-agones:7800".into(),
+            =======
+                            "xds".into(),
+                            "http://quilkin-manage-agones:80".into(),
+            >>>>>>> 76dd00a3 (Move xDS task out of Proxy, add Agones as another discovery task)
+                        ]),
             None,
         );
 
--- a/examples/agones-xonotic-xds/proxy-pool.yaml
+++ b/examples/agones-xonotic-xds/proxy-pool.yaml
@@ -43,7 +43,7 @@ spec:
       containers:
         - name: quilkin
           image: us-docker.pkg.dev/quilkin/release/quilkin:0.5.0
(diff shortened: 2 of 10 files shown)


markmandel:
Oh that makes sense! Very neat!

I'm in the middle of writing an Agones + xDS quickstart, but that's still useful if you want to have a full xDS API (say if you want to run your proxies outside of your cluster), and then we can add another one that's just Agones + `quilkin proxy agones` (may be able to include from one into the other too if we want).

One question I did have, which I couldn't quite see from the code - with `quilkin proxy agones` you still need to provide the `ConfigMap` with the `Filter` configuration? Or is there some kind of file mount instead, like you would for a static config?

XAMPPRocky:
> xDS API (say if you want to run your proxies outside of your cluster), and then we can add another one that's just Agones + quilkin proxy agones (may be able to include from one into the other too if we want).

Exactly, this makes xDS' role here more clear to me. When you have everything in a single cluster a proxy should be able to work just on it's own, and xDS is for when you want to share configuration across multiple clusters.

agones/src/xds.rs
```diff
@@ -358,8 +358,11 @@ management_servers:
             .await
             .unwrap();
         let mount_name = "config";
-        let mut container =
-            quilkin_container(client, Some(vec!["proxy".into()]), Some(mount_name.into()));
+        let mut container = quilkin_container(
```
markmandel, line 361: So I think think you'll want/need to change this test.

We will still need a test for the xDS connection, which this will break in this test.

That being said, we'll want a test for direct Agones integration as well. Can probably reuse the RBAC setup functions for this as well.

markmandel: Edited above - I **don't** think think you'll want/need to change this test 🤦🏻 (I can't type)

XAMPPRocky:
Closing because it's so old, if we still want it, we can open a new PR.

PR #664, opened 2022-12-14
Timeout for xDS gRPPC Client connection

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

I can't seem to find any default timeout for connection from a tonic gRPC connection - which leads me to think it will block forever.

This might be the cause of why there is only a single retry in #660, but at the very least, we should have a timeout anyway, just in case a connection ends up deadlocking for whatever reason.

Work on #660

### DIFF
--- a/src/xds/client.rs
+++ b/src/xds/client.rs
@@ -19,7 +19,7 @@ use std::{collections::HashSet, sync::Arc, time::Duration};
 use futures::StreamExt;
 use rand::Rng;
 use tokio::sync::{broadcast, Mutex};
-use tonic::transport::{channel::Channel as TonicChannel, Error as TonicError};
+use tonic::transport::{channel::Channel as TonicChannel, Endpoint, Error as TonicError};
 use tracing::Instrument;
 use tryhard::{
     backoff_strategies::{BackoffStrategy, ExponentialBackoff},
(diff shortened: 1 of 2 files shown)


src/xds/client.rs
```diff
@@ -116,12 +116,18 @@ impl Client {
...
                     ))),
-                    Some(endpoint) => AggregatedDiscoveryServiceClient::connect(endpoint)
-                        .instrument(tracing::debug_span!(
-                            "AggregatedDiscoveryServiceClient::connect"
-                        ))
-                        .await
-                        .map_err(RpcSessionError::InitialConnect),
+                    Some(endpoint) => {
+                        let endpoint = Endpoint::from_shared(endpoint)
+                            .map_err(RpcSessionError::InitialConnect)?
```
markmandel, line 121: One thought I had here - if we can't create an endpoint, we could return a specific error that ends up with   - rather than the string check of "invalid url" - but wanted to check before implementing. I think we end up in the same place, since if an Endpoint can't be created, we can't ever connect. WDYT?

XAMPPRocky: That makes sense to me.

XAMPPRocky (approved):

Issue #530, opened 2022-04-29
Cant build with `cross`


bombsimon:
**What happened**:

When running `make build-image` (or more exactly `build-linux-binary`) I get

```sh
make: *** [Makefile:94: build-linux-binary] Error 127
```

This also happens when running the failing step without `make`:

```sh
docker run --rm \
  -w /workspace \
  -v /path/to/quilkin:/workspace \
  -v ~/.cargo/registry:/usr/local/cargo/registry \
  -v /var/run/docker.sock:/var/run/docker.sock \
   --entrypoint=cross quilkin-build build --target x86_64-unknown-linux-gnu ; echo $?
127
```

**What you expected to happen**:

Quilkin to build successfully with `cross`.

**How to reproduce it (as minimally and precisely as possible)**:

```sh
docker run --rm \
  -w /workspace \
  -v /path/to/quilkin:/workspace \
  -v ~/.cargo/registry:/usr/local/cargo/registry \
  -v /var/run/docker.sock:/var/run/docker.sock \
   --entrypoint=cross quilkin-build build --target x86_64-unknown-linux-gnu
```

Just changing the `entrypoint` to `cargo` makes it work.

**Anything else we need to know?**:

I don't get any output from the `Makefile` or when running Docker manually but if I attach to the image and try to run `cross` manually I get the following:

```sh
docker run --rm -it \
  -v /path/to/quilkin:/workspace \
  -v ~/.cargo/registry:/usr/local/cargo/registry \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -w /workspace \
  --entrypoint=/bin/bash quilkin-build 
```

```sh
root@6f37b81e4dcc:/workspace#  echo $PATH
/usr/local/cargo/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin

root@6f37b81e4dcc:/workspace# echo $CARGO_HOME
/usr/local/cargo

root@6f37b81e4dcc:/workspace# cross build --target x86_64-unknown-linux-gnu
info: syncing channel updates for '1.59.0-x86_64-unknown-linux-gnu'

  1.59.0-x86_64-unknown-linux-gnu unchanged - rustc 1.59.0 (9d1b2106e 2022-02-23)

info: checking for self-updates
info: downloading component 'rust-src'
info: installing component 'rust-src'
sh: 1: cargo: not found
```

**Environment**:
- Quilkin version: 0.4.0
[...]

PR #670, opened 2023-01-02
Update dependencies

XAMPPRocky:


XAMPPRocky:
@markmandel It seems that using workspace dependencies breaks some assumptions in the Makefile. It would be nice if we could add support for it, as it simplifies dependency management greatly.

agones/Cargo.toml
```diff
@@ -24,8 +24,8 @@ description = "End to end integration tests to be run against a Kubernetes clust
 readme = "README.md"
 
 [dependencies]
-base64 = "0.13.0"
-k8s-openapi = { version = "0.16.0", features = ["v1_22"] }
-kube = { version = "0.76.0", features = ["runtime", "derive"] }
+base64.workspace = true
```
markmandel, line 27: Oh this is dope. This makes things far less brittle.

markmandel:
> @markmandel It seems that using workspace dependencies breaks some assumptions in the Makefile. It would be nice if we could add support for it, as it simplifies dependency management greatly.

Oh yeah, it really does. Lemme take a look, I'm sure I can fix it.

markmandel (approved):

agones/Cargo.toml
```diff
@@ -24,8 +24,8 @@ description = "End to end integration tests to be run against a Kubernetes clust
...
+k8s-openapi.workspace = true
+kube.workspace = true
```
markmandel, line 29: I realised this morning - this may well break our agones e2e integration tests.

Reason being, `"rustls-tls"` doesn't work to authenticate with local `~/.kube` configuration and connect to an IP (https://github.com/kube-rs/kube/issues/153), although it works fine in cluster.

Something like:
```suggestion
kube.workspace = { workspace = true, features = ["runtime", "derive"] }
```

Will probably need to be utilised.

agones/Cargo.toml
```diff
@@ -24,8 +24,8 @@ description = "End to end integration tests to be run against a Kubernetes clust
 readme = "README.md"
 
 [dependencies]
-base64 = "0.13.0"
-k8s-openapi = { version = "0.16.0", features = ["v1_22"] }
-kube = { version = "0.76.0", features = ["runtime", "derive"] }
+base64.workspace = true
+k8s-openapi.workspace = true
+kube = { workspace = true, features = ["runtime", "derive"] }
```
markmandel, line 29: Took some finagling, but iterated until I found the right combo:

```suggestion
kube = { workspace = true, features = ["openssl-tls", "client", "derive", "runtime"] }
```

Cargo.toml
```diff
@@ -17,6 +17,12 @@
 [workspace]
 members = [".", "./macros", "./agones"]
 
+[workspace.dependencies]
+kube = { version = "0.77.0", features = ["derive", "runtime", "rustls-tls", "client"], default-features = false }
+k8s-openapi = { version = "0.16.0", features = ["v1_22", "schemars"] }
+tokio = { version = "1.23.0", features = ["rt-multi-thread", "fs", "signal", "test-util", "parking_lot", "tracing"] }
```
markmandel, line 23:
```suggestion
tokio = { version = "1.24.0", features = ["rt-multi-thread", "fs", "signal", "test-util", "parking_lot", "tracing"] }
```

_Just_ came out.

> The highlight of this release is the reduction of lock contention for all I/O operations ([#530](https://github.com/tokio-rs/tokio/pull/530)). We have received reports of up to a 20% improvement in CPU utilization and increased throughput for real-world I/O heavy applications.

😄  seems like that might be us.

markmandel: Now it's 1.24.1 😄

markmandel:
Just gently bumping this - the two items above should make this good to merge.

Once this is done, I'll cut a 0.5.0 release 👍🏻

XAMPPRocky:
@markmandel Sorry, got lost in the shuffle, but also for changes like that, that are a definite improvement to the PR, you're more than welcome to commit the edits directly without needing my approval. 🙂

PR #674, opened 2023-01-03
Replace SessionManager with TtlMap, refactor downstream packet processing to track more errors in metrics

XAMPPRocky:
I realised that we essentially implemented the same thing with `SessionManager` and `TtlMap`, so I removed session manager and replaced it with `TtlMap`, while also changing this code, I updated the downstream processing code, because there were cases were errors can happen but wouldn't be tracked in metrics, so now the code is written to ensure that all paths lead to the metrics being counted.


markmandel (approved):
Nice change 👍🏻

PR #675, opened 2023-01-04
Make: package_version less brittle

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

There is no easy / cross platform way to get at the local crate's version, so we are doing some grepping, since we can assume grep is installed on local machines.

This change makes the grep pipeline less brittle by looking for the `name="quilkin"` block first before grepping for the version details.

Unblocks #670

N/A

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -29,10 +29,10 @@ rust_toolchain := $(shell grep channel $(project_path)/rust-toolchain.toml | awk
 
 # if this is a release, don't put the sha, otherwise, leave it off.
 ifdef QUILKIN_RELEASE
-	package_version := $(shell grep version -m 1 $(project_path)/Cargo.toml | awk '{print $$3}')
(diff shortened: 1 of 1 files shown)


markmandel:
> If we're willing to depend on `jq` locally (I believe we already depend on it in CI). The following command works and is less brittle than grep, because it pulls the version out of the JSON metadata.

I thought about this, but I wasn't feeling very comfortable assuming people had jq installed, and making it a hard dependency for building Quilkin artifacts. Whereas I feel way more comfortable with grep/awk being locally installed.

I tried to also run `jq` through the build image, but it got really messy really fast (also quite slow), so I figured stick with the greppy solution instead.

XAMPPRocky (approved):

PR #676, opened 2023-01-05
Add initial implementation of QCMP

### LINKED ISSUE #671: Add Control Packet: Ping
To allow the client to measure its latency between a proxy or multiple proxies, we need to add a way to allow the client to talk the proxy. The approach I'd like to propose for how to do this is to add a "Quilkin Control Message Protocol" (name derived from IP Control Message Protocol), which allows for clients who know they're communicating with a proxy to query it for things such as latency.

The packet would look something like the following; we have a magic number to ensure that the packet is distinguishable and unique from regular UDP traffic, a byte to discriminate what kind of command is being sent (currently just one, ping), and then followed by command specific data. 

```
| magic number | command | command specific data |
```

XAMPPRocky:
This PR adds an initial implementation for the Quilkin Control Message Protocol, which allows a client to communicate with the proxy for the purposes of diagnostics such as latency measurement. Currently the only supported command is sending a ping to the proxy, with the proxy returning a ping reply that allows you to calculate the round-trip delay between a client and proxy.

closes #671


markmandel (commented):
Implementation definitely makes sense 👍🏻 

For this initial implementation not a blocker, but it does make me wonder if we should revisit [How do Filters send packets? (or should they?)](https://github.com/googleforgames/quilkin/issues/309) , as then this could be a Filter that end users could choose to drop into their chain, rather than something built into Quilkin itself.

Otherwise - I'd love an integration test for this functionality, and some docs.

XAMPPRocky:
> For this initial implementation not a blocker, but it does make me wonder if we should revisit How do Filters send packets? (or should they?) , as then this could be a Filter that end users could choose to drop into their chain, rather than something built into Quilkin itself.

Even if we did pass a socket to the filters so that they could send packets, at least right now I don't think QCMP should be a filter. It has a divergent control flow that is unrelated to the rest of the filters, and I don't think it should be opt-in, if we have an option for it, it should be opt-out like our admin endpoints. QCMP is basically an admin API, so I think it should be treated closer to that.

> Otherwise - I'd love an integration test for this functionality, and some docs.

Yeah I'll add those over the next week or so, after testing the integration in more production environments, I also realised need to add a sequence number to the ping packet in order to be able to measure packet loss, so I'll also be adding that.

markmandel:
Sounds good on all fronts 👍🏻

XAMPPRocky:
@markmandel This is ready for a new review 🙂

markmandel (approved):

src/protocol.rs
```diff
@@ -0,0 +1,388 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//! Logic for parsing and generating Quilkin Control Message Protocol (QCMP) messages.
+
+use nom::bytes::complete;
+// Magic number to distinguish control packets from regular traffic.
+const MAGIC_NUMBER: &[u8] = b"QLKN";
```
markmandel, line 21: Just for curiosity's sake - did you generate this from the compiler toolchain, or did you write this by hand?

I was playing around with the yaml in https://ide.kaitai.io/# just for fun, so was curious.

XAMPPRocky: @markmandel I wrote it by hand because as far as I could tell the Rust runtime for the kaitai compiler isn't maintained much currently. I think some people are reviving the effort but it's not quite there yet where we can rely on it. Eventually we can replace it with a generated implementation once it is more stable and reliable.

markmandel: @XAMPPRocky  I did notice that about the Rust runtime when I went and looked, but wasn't sure if you had found some magic 👍🏻  thanks!

PR #677, opened 2023-01-05
Make: Even better package_version

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

#675 was better than the original, but still potentially brittle if the format of the Cargo.toml changed.

This is better than #675, since gets the value from a `cargo` command, so it no longer matters if the Cargo.toml changes -- everything should still work as always, since it's a format.

We are also now only relying on `cargo` and `awk` as dev dependencies, which is actually one less dependency than before (grep).

N/A

N/A

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -27,12 +27,11 @@ CARGO_HOME ?= $(build_path)/.cargo
 BUILD_IMAGE_TAG ?= quilkin-build
 rust_toolchain := $(shell grep channel $(project_path)/rust-toolchain.toml | awk '{ print $$3 }')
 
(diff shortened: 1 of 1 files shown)


XAMPPRocky (approved):

markmandel:
Blerg, I just realised why this is failing - CI (and really the `/build` toolchain) only assumes Make + Docker are the two installed dependencies... so `cargo` isn't' installed, so there is no value for the version installed.

Okay, going to close this and stick with grep + awk as it currently stands. If we run into issues again, I'll come up with some other clever ideas.

Issue #600, opened 2022-09-22
Multi-Cluster Support


XAMPPRocky:
Now that we have Quilkin running as a xds management server, we want to extend that capability to multiple clusters, as currently it needs to run in the same cluster as where the gameservers are hosted where as we want to be able to place Quilkin at the edge in more regions than just the ones that contain gameservers.

To do this I'd like to propose a very simple addition to our current architecture. Which is the introduction of a kind of "relay" service that sits on-top of the management servers. Each management server would send a message to the relay to give its IP address, and then the relay server would initiate an xDS discovery stream with the management server, and then load balancers would connect to the relay service as a management server which would serve the aggregation of all the management servers.

Having the management servers initiate to relay to start the communication is important, because that allows the relay to serve N management servers dynamically as they pop up and shut down, the rest is just using the same xDS logic that we already have.

markmandel:
I haven't gone deep on this, but could we use  https://github.com/envoyproxy/xds-relay ?

PR #688, opened 2023-01-18
Refactor documentation layout and define default ports

XAMPPRocky:
This PR defines a set of default ports for each service (currently just proxy and manage). Additionally as part of that, I wanted to restructure our current documentation to have a more clear flow through table of contents, summarising some of the important parts of the services to make it more clear for administrators, and better handle documenting new services like #600

The choice for the port numbers were as follows, instead of starting at a high number and incrementing by one for each service, I thought it would be better to have service ports incremented by 100, this makes it easier to read in logs, and means that we could use a range of ports for multiple listeners related to a service while making more obvious just from the port what service is being run. I went with `7777` as the new default port as on the [wikipedia list of TCP and UDP numbers](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers) there are more gameservers that use `7777` than any other one, and since the choice of a port number is mostly purely arbitrary, why not pick a number that is already somewhat common in gameservers in the wild, making it potentially one less thing you need to set to get things to work. 

| services | ports | Protocol |
|----------|-------|-----------|
| Proxy | 7777 | UDP (IPv4) |
| xDS | 7800 | gRPC(IPv4) |
| Relay (TBD) | 7900 | gRPC(IPv4) |
| Administration | 8000 | HTTP (IPv4 OR IPv6) |

### New Table Of Contents
Screenshot 2023-01-18 at 17 01 16


markmandel:
Lotsa broken links, but the TOC looks really nice 👍🏻

XAMPPRocky:
> Lotsa broken links, but the TOC looks really nice 

Yeah, I wanted to show what it would look like before polishing it up, I've fixed all the links now, but it's still failing, I've noticed that the make command deletes `print.html` before the htmltest, but I don't know why it fails for just a few of the pages and not all of them.

markmandel:
print.html has broken links because the relative links no longer work (and who needs it for a printed page) - so we delete em so that the link checker doesn't complain.

I can take a look, see what the deal is there. May just need to add some more cleanup of print.html

XAMPPRocky:
I've written a simple fix, we just insert an empty html document after deleting it, so it will pass for htmltest.

docs/src/SUMMARY.md
```diff
@@ -1,38 +1,42 @@
...
+
+# SDKs
+- [Unreal Engine](./sdks/unreal-engine.md)
+
+# Deployment
+- [Administration](./deployment/admin.md)
+- [Architecture Examples](./deployment/examples.md)
+- [Configuration](./deployment/configuration.md)
+
+# Quickstart Guides
```
markmandel, line 39: I'd be tempted to to move the quickstarts to before **Services**.

If we look at Google Cloud doc's as a reference, Quickstarts always come towards the top, as they are the easy path to getting started with the product.

For example:
https://cloud.google.com/kubernetes-engine/docs/deploy-app-cluster

image

build/Makefile
```diff
@@ -108,6 +108,7 @@ test-docs:
 			cargo doc --workspace --no-deps && cd docs && mdbook build --dest-dir /tmp/docs/book && \
 			cp -r "$(CARGO_TARGET_DIR)/doc" /tmp/docs/api && \
 			rm /tmp/docs/book/print.html && \
+			echo "<!doctype html>" > /tmp/docs/book/print.html && \
```
markmandel, line 111: 👍🏻

docs/src/deployment/examples.md
```diff
@@ -143,11 +143,11 @@ while also providing the most redundancy and security for your dedicated game se
 ## What Next?
 
 * Have a look at the [example configurations](./examples.md) for configuration and usage examples.
```
markmandel, line 145: With this rename, this now points to itself, we should either remove this line, replace it with something else or put back in a page for listing examples?

docs/src/examples.md
```diff
@@ -1,17 +0,0 @@
-# Examples
-
-See the [examples](https://github.com/googleforgames/quilkin/tree/{{GITHUB_REF_NAME}}/examples) folder on Github for configuration and
```
markmandel, line 3: I couldn't find this content anywhere - maybe add it back in under **Deployment** as the last item?

Or maybe it's better with a table that actually lists all the examples?

markmandel (commented):
This is a really nice update!

markmandel:
Also, sorry for the delay on review -- it's been a week 😞

docs/src/SUMMARY.md
```diff
@@ -1,38 +1,42 @@
...
+        - [Load Balancer](./services/proxy/filters/load_balancer.md)
+        - [Local Rate Limit](./services/proxy/filters/local_rate_limit.md)
+        - [Match](./services/proxy/filters/match.md)
+        - [Pass](./services/proxy/filters/pass.md)
+        - [Timestamp](./services/proxy/filters/timestamp.md)
+        - [Token Router](./services/proxy/filters/token_router.md)
+        - [Writing Custom Filters](./services/proxy/filters/writing_custom_filters.md)
+    - [Metrics](./services/proxy/metrics.md)
+
+---
```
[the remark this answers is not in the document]
XAMPPRocky: Well the intention was to add more breaks as we add more services to give them a bit more spacing since they can have a lot of parts to them. Here's how it would look like with a third service chapter. I don't feel super strongly about it one way or another.

## Without Break

Screenshot 2023-01-25 at 11 24 47


## With Break

Screenshot 2023-01-25 at 11 35 42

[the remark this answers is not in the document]
markmandel: Ah that makes perfect sense. So don't mind if you leave it in, or remove it and put it back later -- whatever works 👍🏻

docs/src/examples.md
XAMPPRocky: I've now included this information in deployment/examples

markmandel (approved):
LGTM!

PR #690, opened 2023-01-20
Also bind the UDP socket to IPv6

### LINKED ISSUE #666: Listening on IPv4 and IPv6 addresses

We have UDP based services that are able to run on IPv4 and IPv6 this helps users that have a "weak" IPv4 connectivity due to, e.g., their ISP providers lacking the IPv4 bandwidth (packet loss at the ISP's backbones as far as we have analyzed in most situations that we encountered).
For that we would like to run service accessible over IPv4 and IPv6.

It would be great if quilkin had an option to listen on IPv4 and IPv6 addresses so users with a better IPv6 connectivity due to, e.g., ISP "uplink" limitations.  Especially as quilkin seems to be capable to sending IPv4 traffic to IPv4 and IPv6 endpoints already from what I have read in the documentation and examples.

We can go IPv4-only but in the past even that has shown to cause some issues for "less optimized" services for users with ISPs that

XAMPPRocky:
fixes #666

### DIFF
--- a/tests/firewall.rs
+++ b/tests/firewall.rs
@@ -34,7 +34,7 @@ async fn firewall_allow() {
     let yaml = "
 on_read:
   - action: ALLOW
-    source: 127.0.0.1/32
+    source: ::/32
     ports:
        - %1
 on_write:
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -19,7 +19,7 @@ mod sessions;
 pub use sessions::{Session, SessionArgs, SessionKey};
 
 use std::{
-    net::{Ipv4Addr, SocketAddrV4},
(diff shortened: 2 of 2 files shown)


tests/firewall.rs
```diff
@@ -34,7 +34,7 @@ async fn firewall_allow() {
     let yaml = "
 on_read:
   - action: ALLOW
-    source: 127.0.0.1/32
+    source: ::/32
```
suom1, line 37: Replace `::/32` with `::1/128` as that's the common address for ipv6 loopback :)

XAMPPRocky:
Closing as this has been implemented in #752

PR #694, opened 2023-01-30
Strip Config of all service specific configuration for now

XAMPPRocky:
More work on #600. This removes all service specific configuration from `Config` (well specifically non-shared service configuration, filters is still there because filters are shared between services even though only proxy uses them), as currently `Config` is a little too proxy focused, and with the introduction of a third service, I wanted to restrict `Config` to just what is currently shared between all services, and service specific configuration should be set using CLI or environment parameters (or using the `cli` from the lib). This also removes xds' dependence on `Config` in order to better support running multiple types of clients and different services handling resource updates differently.

We could always later reintroduce these properties into `Config` when we've set up a good way to actually handle updates to these properties as they're currently only set once (such as binding new sockets when the port changes), but I think we should do that after #600 is finished.


src/config.rs
```diff
@@ -59,34 +55,17 @@ pub(crate) const CONNECTION_TIMEOUT: u64 = 5;
 #[serde(deny_unknown_fields)]
 #[non_exhaustive]
 pub struct Config {
-    #[serde(default = "Slot::<Admin>::with_default")]
-    pub admin: Slot<Admin>,
     #[serde(default)]
     pub clusters: Slot<ClusterMap>,
```
markmandel, line 59: (non blocking) what are your thoughts on keeping static cluster config? (since it's proxy specific)

markmandel (approved):

src/config.rs
XAMPPRocky: Well clusters is one of the least proxy specific pieces of information because the control plane and relay create a cluster map for proxies to consume through xDS.

We do still have static configuration, the API just also handles being updated through other sources if provided, unless I'm missing the question?

markmandel: Ah yes - that is a good point - everyone needs clusters/endpoints, so it's not actually proxy specific 👍🏻

Question answered!

PR #698, opened 2023-02-08
Add initial implementation of relay service

### LINKED ISSUE #600: Multi-Cluster Support

XAMPPRocky:
This PR adds the initial implementation of the multi-cluster relay service. This service allows you to connect multiple control planes into a single aDS server for all proxy services.

fixes #600

### DIFF
--- a/build/build-image/Dockerfile
+++ b/build/build-image/Dockerfile
@@ -68,8 +68,7 @@ RUN set -eux; \
     chmod -R a+w $RUSTUP_HOME $CARGO_HOME && \
     rustup component add rustfmt clippy && \
     rustup target add x86_64-unknown-linux-gnu aarch64-unknown-linux-gnu x86_64-pc-windows-gnu && \
-    cargo install cargo-watch mdbook && \
-    cargo install --git https://gitlab.com/hydai/mdbook-variables.git --branch bump_toml && \
+     cargo install cargo-watch mdbook mdbook-variables && \
     cargo install cargo-about && \
     cargo install --locked cargo-deny && \
     rustup --version && \
--- a/build/build-image/Dockerfile
+++ b/build/build-image/Dockerfile
@@ -68,7 +68,7 @@ RUN set -eux; \
     chmod -R a+w $RUSTUP_HOME $CARGO_HOME && \
     rustup component add rustfmt clippy && \
     rustup target add x86_64-unknown-linux-gnu aarch64-unknown-linux-gnu x86_64-pc-windows-gnu && \
-     cargo install cargo-watch mdbook mdbook-variables && \
+    cargo install cargo-watch mdbook mdbook-variables && \
     cargo install cargo-about && \
     cargo install --locked cargo-deny && \
     rustup --version && \
--- a/build/build-image/Dockerfile
+++ b/build/build-image/Dockerfile
@@ -25,7 +25,7 @@ ENV RUSTUP_HOME=/usr/local/rustup \
 # Install packages
 RUN set -eux && \
     apt-get update && \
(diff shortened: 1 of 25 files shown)


build/build-image/Dockerfile
```diff
@@ -68,7 +68,8 @@ RUN set -eux; \
     chmod -R a+w $RUSTUP_HOME $CARGO_HOME && \
     rustup component add rustfmt clippy && \
     rustup target add x86_64-unknown-linux-gnu aarch64-unknown-linux-gnu x86_64-pc-windows-gnu && \
-    cargo install cargo-watch mdbook mdbook-variables && \
+    cargo install cargo-watch mdbook && \
+    cargo install --git https://gitlab.com/hydai/mdbook-variables.git --branch bump_toml && \
```
XAMPPRocky, line 72: This is just to work around incompatible toml dependencies between mdbook and the plugin.

docs/src/services/relay.md
```diff
@@ -0,0 +1,23 @@
...
+For multi-cluster integration, Quilkin provides a `relay` service, that can be
+used with a multiple [control plane](./xds.md) services in different clusters to
+provide a unified "Aggregated Discovery Service" endpoint for [proxy](./proxy.md)
+services.
+
+To connect to a control plane to a relay, add the `--relay` flag to your control
+plane with the address of the relay. Then to connect a proxy service to the
+relay's ADS endpoint, you use the same `--management-server` argument for
+connecting to control planes.
+
```
markmandel, line 17: I'd love to see a simple example here for clarify - something someone could run locally on their machine.

Reading the docs I'm mentally trying to picture which if it's `quilkin proxy` has the `--relay` or if it's the `quilkin relay` or something else? (I'm pretty sure I get it, but I had to squint on it). Also to show which port should be connected to where. It looks like it's easy to mess up.

I was thinking something like (please make any corrections as needed):

---

As a simple local example (we'll disable `--no-admin` to avoid port conflicts), you can have:

To run the relay:
```shell
$ quilkin relay --no-admin
```

Connect a file control plane to the relay

```shell
$ quilkin manager file ./examples/proxy.yaml --no-admin --relay http://localhost:7900
```

No we can connect the proxy to the relay:

```shell
$ quilkin proxy --management-server http://localhost:7800
```

docs/src/services/relay.md
```diff
@@ -0,0 +1,23 @@
+# Control Plane Relay
```
markmandel, line 1: Link to this in SUMMARY.md please.

markmandel (commented):
Looks good - just some tweaks to the docs 👍🏻

markmandel:
Looks like `cargo install --git https://gitlab.com/hydai/mdbook-variables.git --branch bump_toml` doesn't exist anymore.

markmandel (approved):
Love the Quickstart.

https://crates.io/crates/mdbook-variables/0.2.1 - had a release a few days ago, hopefully that will resolve the issue you were facing 👍🏻

Issue #700, opened 2023-02-13
Replace `Config` with configuration components


XAMPPRocky:
Currently we have the concept of a single "monolith" config file that's shared between all services from when there was only a proxy service. In its current form I don't find it to always to be clear when something is being shared and when something isn't, so I'd like to propose that we move away from the current configuration and instead create configuration files for specific components in Quilkin, this will make it easier to isolate and share specific parts of configuration without accidentally overriding another part of configuration, which also enables us to potentially have multiple configuration providers for different components to be running at once. It will also allow for the configuration to diverge from the schema of the current `Config` to be potentially more user-friendly.

A couple examples of component configuration that I think we should start with, namely ones for each service, and for `FilterChain` configuration. To be able to share code and make the configuration of each component more unified, I think we should adopt the same schema as Kubernetes CRDs, this would allow us to reuse more code for parsing, it's a format that people are already familiar with, and it allows for these components to potentially become CRD's down the line that can be discovered using a kubernetes configuration provider.

```yaml
version: quilkin.dev/v1alpha1
kind: Proxy
spec:
  filter_chain: acme.games/hello-world
  endpoints:
    - 192.168.0.1:8080
    - 192.168.0.1:8080
```

```yaml
version: quilkin.dev/v1alpha1
kind: ControlPlane
spec:
  provider:
    kind: Agones
    config_namespace: default
    gameserver_namespace: default

```

```yaml
version: quilkin.dev/v1alpha1
kind: Relay
spec:
  ads_port: 7800
  cpds_port: 7900
```

```yaml
version: quilkin.dev/v1alpha1
kind: FilterChain
metadata:
     id: acme.games/hello-world
spec:
  filters:
    - name: quilkin.filters.firewall.v1alpha1.Firewall
      config:
        on_read:
          - action: ALLOW
            source: 192.168.51.0/24
            ports:
              - 10
              - 1000-7000
        on_write: 
          - action: DENY
            source: 192.168.51.0/24
            ports:
              - 7000
```

XAMPPRocky:
I think these components would also allow for another service called `agent`, that would allow you to have all services in a config source provider, the agent could run multiple services inside a single process, and it would make the listeners dynamic. The agent could add, change and remove proxies, control planes, relays, filters, and endpoints dynamically at runtime.

Then the current subcommands would essentially just be the CLI representations of these components.

[...]

PR #711, opened 2023-02-22
Remove config being picked up in the CLI

XAMPPRocky:
Part of #700 this PR removes `Config` from being picked up in the CLI for now. So now the only way to provide configuration is through the file provider for a service.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -14,10 +14,7 @@
  * limitations under the License.
  */
 
-use std::{
-    path::{Path, PathBuf},
-    sync::Arc,
-};
+use std::sync::Arc;
 
 use clap::crate_version;
 use tokio::{signal, sync::watch};
@@ -45,7 +42,6 @@ pub mod manage;
 pub mod proxy;
 pub mod relay;
 
-const ETC_CONFIG_PATH: &str = "/etc/quilkin/quilkin.yaml";
 const PORT_ENV_VAR: &str = "QUILKIN_PORT";
 
 /// The Command-Line Interface for Quilkin.
@@ -56,9 +52,6 @@ pub struct Cli {
     /// Whether to spawn the admin server or not.
     #[clap(env, long)]
     pub no_admin: bool,
-    /// The path to the configuration file for the Quilkin instance.
-    #[clap(short, long, env = "QUILKIN_CONFIG", default_value = "quilkin.yaml")]
-    pub config: PathBuf,
     /// The port to bind for the admin server
     #[clap(long, env = "QUILKIN_ADMIN_ADDRESS")]
     pub admin_address: Option<std::net::SocketAddr>,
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
@markmandel This seems to be failing in a weird way that I can't tell what's happening.

```
Finished Step #6 - "test-quilkin-image-default-config-file"
ERROR
ERROR: build step 6 "gcr.io/cloud-builders/docker" failed: step exited with non-zero status: 255
Step #9 - "test-agones": make: *** [Makefile:209: push] Terminated
Step #4 - "test-docs": [0m
```

markmandel:
Did you delete your comment?

Looks like the change broke the test that attempts to run the default Docker proxy image:

```
Already have image (with digest): gcr.io/cloud-builders/docker
{"timestamp":"2023-07-18T12:21:12.198881Z","level":"INFO","fields":{"message":"Starting Quilkin","version":"0.7.0-dev","commit":"cae0736c50862e687407775846eda1d351155ca4"},"target":"quilkin::cli","filename":"src/cli.rs"}
{"timestamp":"2023-07-18T12:21:12.199228Z","level":"INFO","fields":{"message":"Starting admin endpoint","address":"[::]:8000"},"target":"quilkin::admin","filename":"src/admin.rs"}
{"timestamp":"2023-07-18T12:21:12.199669Z","level":"ERROR","fields":{"message":"fatal error","error":"`quilkin proxy` requires at least one `to` address or `management_server` endpoint.","error_debug":"`quilkin proxy` requires at least one `to` address or `management_server` endpoint."},"target":"quilkin","filename":"src/main.rs"}
```

https://github.com/googleforgames/quilkin/blob/ea5509f74c7c9b2048542ee2d8c6b1fc29112b4a/cloudbuild.yaml#L55-L63

Do we want to remove the Config search completely for this development path?

If we want to maintain a search path, and since we'll have multiple files we could do `/etc/quilkin/quilkin.yaml`  (maybe for a backward compatibility / not break tests), and then load all  yaml files in `/etc/quilkin/quilkin.d/` for all the different `Kind` of configurations.

That would align with Linux config and we could likely keep a backward compatibility, merge features towards #700, still do releases and continue with this work until complete.

WDYT?

XAMPPRocky:
Yeah I figured out what's wrong with this, I need to also add the file provider to the proxy service so that the examples work.

> Do we want to remove the Config search completely for this development path?

No no, I want move proxy to being provider based like the other services. So you'd write `quilkin proxy file` instead, because that would also allow for `quilkin proxy agones` where you have the discovery happening in a single service.

markmandel:
> No no, I want move proxy to being provider based like the other services. So you'd write `quilkin proxy file` instead, because that would also allow for `quilkin proxy agones` where you have the discovery happening in a single service.

Got it 👍🏻 

Yeah, that makes 100% sense then to remove it.

Or if you want to do this + `quilkin proxy file`, all in one PR, that could mean we could keep the cloudbuild test step?

Just thinking through removing and adding test steps is all.

XAMPPRocky:
Yeah exactly, I'll update this with adding that.

XAMPPRocky:
Needs more to be mergeable

Issue #695, opened 2023-01-31
Print pod logs when steps fail in kubernetes integration tests


XAMPPRocky:
Currently when the agones tests fail all you usually see is the `Elapsed` error. It would be nice if we added an error handler task to the agones integration, that whenever a task for a pod fails it outputs the logs from that pod, as well as the pod description, so that you can get a detailed overview of what went wrong in the test rather than just a "exceeded timeout".

markmandel:
🤔 can you catch the panic on a failing Rust Test? I'd always assumed not. But yes, this would be extremely handy. Although with a panic handler you would need to go find the pods for that given test.

The other option is to have a utility "printPodLog(name: string)` function, that we can put in appropriate places. Would need to do things like manually check for equality first before doing the `assert` check, but that's perfectly doable.

markmandel:
Another fun thought I just had. We could make a utility that puts a test specific unique label on each pod that a test creates, that way if a test fails, you could lookup all the pods that are being used for the test, and then grab all the logs and output them.

Oooh, I like this idea 😄

XAMPPRocky:
> can you catch the panic on a failing Rust Test? I'd always assumed not

You can catch a panic, but we don't need to catch a panic because that's an error in the current process, rather than an error in the nodes, we just need to replace the `.unwrap()` on the timeouts with a error handler path that prints the logs I think.

PR #712, opened 2023-02-22
Don't apply xDS resources when the version matches the previously applied version

XAMPPRocky:
Previously the client would always attempt to apply configuration changes, this changes it so we only apply changes when changes are present.

### DIFF
--- a/agones/src/xds.rs
+++ b/agones/src/xds.rs
@@ -56,6 +56,8 @@ mod tests {
     /// for this test, we should only have single Agones integration test in this suite, since they
     /// could easily collide with each other.
     async fn agones_token_router() {
+        quilkin::test_utils::enable_tracing_log();
+
         let client = Client::new().await;
 
         let deployments: Api<Deployment> = client.namespaced_api();
--- a/src/test_utils.rs
+++ b/src/test_utils.rs
@@ -34,8 +34,7 @@ use crate::{
 static ENABLE_LOG: Lazy<()> = Lazy::new(|| {
     tracing_subscriber::fmt()
         .pretty()
-        .with_env_filter("quilkin=trace")
-        .with_max_level(tracing::Level::TRACE)
+        .with_env_filter("info,quilkin=trace")
         .init()
 });
 
--- a/src/test_utils.rs
+++ b/src/test_utils.rs
@@ -34,7 +34,7 @@ use crate::{
 static ENABLE_LOG: Lazy<()> = Lazy::new(|| {
     tracing_subscriber::fmt()
         .pretty()
-        .with_max_level(tracing::Level::Debug)
+        .with_max_level(tracing::Level::DEBUG)
         .init()
 });
 
--- a/src/test_utils.rs
+++ b/src/test_utils.rs
@@ -34,6 +34,7 @@ use crate::{
 static ENABLE_LOG: Lazy<()> = Lazy::new(|| {
     tracing_subscriber::fmt()
         .pretty()
+        .with_env_filter("quilkin=trace")
         .with_max_level(tracing::Level::TRACE)
         .init()
 });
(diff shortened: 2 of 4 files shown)


markmandel:
Cleaning up PR's - I was going to dig into this and see why it was failing. Let me know if this PR is no longer relevant.

May end up working on some #695 to make this self serve.

XAMPPRocky:
This PR is still relevant, it would save some amount of compute especially as the amount of resources grows.

markmandel:
Sweet - will investigate!

markmandel:
Sorry about the rebuilding - using this PR to test out the new functions I have for debugging Agones' integrations in CI.

PR to this PR incoming shortly.

markmandel:
I'm fairly sure I know what the issue is, but not entirely sure why this is only failing now, and it wasn't failing before.

For context - the Agones test creates a deployment of Proxies, and waits for them to become `ready`. This works in `main`.

If we look at the code for a proxy's `/ready` handler, it's this:

https://github.com/googleforgames/quilkin/blob/af01921a1b8aed25eb3ba6c25c9c4fe0fe7c6d3b/src/admin.rs#L98-L106

So if there are no endpoints, it should return a `500` http code. Weirdly, this _passes_ on `main`, even though we don't even create a `GameServer`, let alone allocated one, until _after_ we have created the Deployment of Proxies.

But within this PR, it actually fails -- which is what it likely should do. If you have any thoughts on why this PR might change this, I'd be curious.

I can change the order in the test, but I am wondering why this passed in the first place. I might dig in and see what I can see.

markmandel:
Actually - ignore the above comment! (except for the reminder on what makes `/ready` return a HTTP 200.

So this is where the error happens:

https://github.com/googleforgames/quilkin/blob/c9ba50453343e284450b1fa92bf9fee34b00a6c5/agones/src/xds.rs#L140-L156

Which, as you can see, is _right after_ we allocate a GameServer. 

When I manually check the `/config` of the proxies, I get:

```json
{
  "clusters": {
    "default": {
      "localities": [
        {
          "locality": null,
          "endpoints": []
        }
      ]
    }
  },
  "filters": [
    {
      "name": "quilkin.filters.capture.v1alpha1.Capture",
      "label": null,
      "config": {
        "metadataKey": "quilkin.dev/capture",
        "suffix": {
          "size": 3,
          "remove": true
        }
      }
    },
    {
      "name": "quilkin.filters.token_router.v1alpha1.TokenRouter",
      "label": null,
      "config": null
    }
  ],
  "id": "quilkin-proxies-756545ff86-xczx5",
  "version": "v1alpha1"
}
```

So no endpoints.

Interestingly! If I check the xds management system, it _does_ have the endpoints!

```json
{
  "clusters": {
    "default": {
      "localities": [
        {
          "locality": null,
          "endpoints": [
            {
              "address": "34.168.219.69:7060",
              "metadata": {
                "quilkin.dev": {
                  "tokens": [
                    "NDU2"
                  ]
                },
                "name": "fleet-jb6zw-c9jcg-dw4wj"
              }
            }
          ]
        }
      ]
    }
  },
  "filters": [
    {
      "name": "quilkin.filters.capture.v1alpha1.Capture",
      "label": null,
      "config": {
        "metadataKey": "quilkin.dev/capture",
        "suffix": {
          "size": 3,
          "remove": true
        }
      }
    },
    {
      "name": "quilkin.filters.token_router.v1alpha1.TokenRouter",
      "label": null,
      "config": null
    }
  ],
  "id": "quilkin-manage-agones-7b89c44dff-np4nb",
  "version": "v1alpha1"
}

```

So - there is a bug in this code such the proxies are not accepting the configuration information from the xds provider 😢

XAMPPRocky:
Well it's great you found the bug 😄 I'll investigate this further when it's work hours.

XAMPPRocky:
Superseded by #873

PR #719, opened 2023-03-18
Create unreal CI

XAMPPRocky:
Trying out https://unrealcontainers.com to see if it can build and test the unreal plugin in CI.


XAMPPRocky:
@markmandel we need a personal access token in the secrets to be able to pull the image.

PR #720, opened 2023-03-21
Workaround for an issue in the quilkin proxy

andersroos:
The proxy makes dns lookups that slows down proxying.

This patch makes dns lookups in the manager instead, sending ips to the proxy instance so it does not have to do lookups.

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Closes #

### DIFF
--- a/src/config/providers/k8s/agones.rs
+++ b/src/config/providers/k8s/agones.rs
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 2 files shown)


andersroos:
Unintentional opening of PR, don't think the fix should be there. Sorry :-D

Issue #725, opened 2023-03-30
Release 0.6.0


markmandel:
Milestone: https://github.com/googleforgames/quilkin/milestone/4

Issue is for discussion on what should be in/out of the release.

Checklist of items blocking release at this stage:

Move documentation for relay to experimental?

* [x] #718 
   * [x] Move Filter trait to have async functions
   * [x] #721 
* [x] Update relay documentation to mark as "EXPERIMENTAL"
* [x] Filter metric updates
  * [x]  Updates to Grafana dashboards
  * [x]  Documentation
* [x] Update/check examples to make sure they still work.
* [x] Use github changelog to generate the changelog, and replace the tool we have.
* [x] Check documentation for issues
   * [x] Example in https://googleforgames.github.io/quilkin/main/book/deployment/configuration.html is broken.

markmandel:
@XAMPPRocky did I miss anything?

markmandel:
Added a section to check on the examples.

markmandel:
Finally able to get this release done.

I'd like to get #749 into the release, but it's possibly not critical, and I know you are swamped @XAMPPRocky -- what do you think, should I get a release running or should we wait for #749? (it only needs docs).

markmandel:
Forgot to note -- I'm going to check on the Grafana dashboards and examples now, so there is some time.

markmandel:
Grafana dashboards look good!

![image](https://github.com/googleforgames/quilkin/assets/298370/976b3167-6757-47fd-aab1-b56c63d6faaf)

Working through the examples - finding a few breaking changes, but nothing major so far.

markmandel:
Got the sidecar examples working, just needed some small config changes.

xDS is next.

markmandel:
xDS works. Just noting to myself I think I didn't test the sidecar compression example... it doesn't look like it would work as I have it now (should be using `--port`).

markmandel:
@XAMPPRocky I think we're pretty close to a release! Shall we wait on https://github.com/googleforgames/quilkin/pull/746 - or just cut the release? (that's more based on your schedule).

Happy to wait, but figured I would check.

XAMPPRocky:
Let's not wait, it can always be in the next release.

Generally I prefer to release consistently and often, even if it means not all the features we've added are available.

markmandel:
SGTM. I'll get moving on the release!

markmandel:
# Release Checklist for Release 0.6.0

## Prerequisites

[...]

PR #730, opened 2023-04-03
Add experimental notification in the docs

XAMPPRocky:
work on #725

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

Closes #

### DIFF
--- a/docs/src/services/relay.md
+++ b/docs/src/services/relay.md
@@ -5,6 +5,10 @@
 | ADS | 7800 | gRPC(IPv4) |
 | CPDS | 7900 | gRPC(IPv4) |
 
+> **Note:** This service is currently in active experimentation and development
+  so there may be bugs which cause it to be unusable  for production, as always
+  all bug reports are welcome and appreciated. 
+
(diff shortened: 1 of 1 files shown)


markmandel (commented):
Looks good. Marking as approved, but you may also want to add a similar section to https://github.com/googleforgames/quilkin/blob/main/src/cli/relay.rs#L25-L29 so it shows up in the command line help for `quilkin relay --help` as well.

I'll leave it to you on if you want to do that in a separate PR or not.

markmandel (approved):

XAMPPRocky:
> so it shows up in the command line help

Personally I think it's fine for now to have it just in the docs, as the whole tool is still `0.x` so there's so some assumption of instability anyways, and the relay stuff does work, we just haven't actually deployed it yet, don't have an integration example, and I think there's a couple of changes to how it works that we'll make after we've first deployed it. It's not that experimental that it warrants multiple disclaimers in my opinion.

PR #736, opened 2023-04-17
Refactor filter metrics into a single vector of metrics with labels

XAMPPRocky:
This replaces the bespoke filter metrics with generic metrics, and we now put the specifics into the labels of the generic filter. This will make it easier to build general visualisations that can work with little to no setup in Grafana. Previous filter metrics that reported dropped packets were removed as we now report drops keyed by filter so that replaces the need for those metrics.

This would also allow us in the future to offer a metric API to guest filters.


src/filters/metrics.rs
```diff
@@ -0,0 +1,73 @@
...
+        metadata.len() <= 1,
+        "shared metadata exceeds current label cardinality"
+    );
+    static HISTOGRAMS: Lazy<HistogramVec> = Lazy::new(|| {
+        prometheus::register_histogram_vec_with_registry! {
+            prometheus::histogram_opts! {
+                "filter_histogram",
+                "generic filter histogram, see help label for more specific info",
+            },
+            &["id", "label", "help", Direction::LABEL, "shared_metadata_1"],
```
markmandel, line 60: What's this "shared_metadata_1" used for?

docs/src/services/proxy/metrics.md
```diff
@@ -97,6 +97,12 @@ The proxy exposes the following metrics around sessions:
   The duration it took for a `filter`'s `write` implementation to execute.
   * The `filter` label is the name of the filter being executed.
 
+* `filter_int_counter{id, label, help, direction}`
```
markmandel, line 100: I think it would be good to have a section on Filter's generic metrics and how it gets generated, with examples of what label, help, direction etc could possibly be as well.

src/filters/compress.rs
```diff
@@ -193,14 +193,6 @@ mod tests {
             expected.len(),
             read_context.contents.len()
         );
-        assert_eq!(
```
markmandel, line 196: Is it possible to keep the metrics tests - just to ensure we are counting the metrics as we expect to? Or is that not posisble with this refactor?

src/filters/capture/affix.rs
```diff
@@ -1,20 +1,8 @@
...
-        if metrics.packets_dropped_total.get() % 1000 == 0 {
-            tracing::warn!(count = ?metrics.packets_dropped_total.get(), "Packets are being dropped due to their length being less than {} bytes", size);
-        }
-        metrics.packets_dropped_total.inc();
-
-        false
-    } else {
-        true
-    }
+/// Returns whether the capture size is bigger than the packet size.
```
markmandel, line 3: It's really nice how clean this now - no more need to manually manage these metrics 👍🏻

markmandel (commented):
Overall this looks really good! Just a couple of questions and a note about documentation.

Will need to update the sample dashboard as well, but that can come later, and we can expand that over time 👍🏻

src/filters/metrics.rs
XAMPPRocky: It's a generic label for any additional metadata that could be used by a filter, currently it's only used by the `timestamp` filter so it has a label to place what metadata key it used to generate the result.

src/filters/compress.rs
XAMPPRocky: It is possible, it's just more effort. We'd need to move the metrics tests into the integration tests (`tests/`) so that the metrics aren't shared between the tests.

If it's okay, I think I would prefer to add it back (along more tests for metrics in general), in a separate PR(s) since it requires making a lot of visibility changes to expose the metrics publicly so they can be used in the metrics tests. We can make a tracking issue that tracks adding these tests.

markmandel:
Apologies for delay - Kubecon is on. But this looks good - only thing I think we need is some more fleshed out documentation on how metrics get created for each Filter, and this is good to merge, and one of us to create that tracking ticket for updating tests for metrics 👍🏻

markmandel (approved):

PR #740, opened 2023-04-25
Update termination code to allow proxies to wait until all sessions expire

XAMPPRocky:
Currently Quilkin proxies will immediately terminate when they receive a termination signal, this PR changes it so that it will first wait for all active sessions to expire before it terminates, this in combination with a long graceful termination period should help avoid situations where a k8s scaling results in a load balancer serving active connections to be terminated.

I think there's an open question as whether this logic should also apply to `SIGINT`, since that's usually from user's `CTRL+C`'s the process while testing, but there might be situations where `SIGINT` is sent that I'm not thinking of.

### DIFF
--- a/src/cli/proxy.rs
+++ b/src/cli/proxy.rs
@@ -115,13 +115,22 @@ impl Proxy {
             None
         };
 
-        self.run_recv_from(&config, sessions, shutdown_rx.clone())?;
+        self.run_recv_from(&config, sessions.clone(), shutdown_rx.clone())?;
         tracing::info!("Quilkin is ready");
 
         shutdown_rx
             .changed()
             .await
-            .map_err(|error| eyre::eyre!(error))
+            .map_err(|error| eyre::eyre!(error))?;
+
+        tracing::info!(sessions=%sessions.len(), "waiting for active sessions to expire");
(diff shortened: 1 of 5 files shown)


src/cli/proxy.rs
```diff
@@ -115,13 +115,22 @@ impl Proxy {
...
 
         shutdown_rx
             .changed()
             .await
-            .map_err(|error| eyre::eyre!(error))
+            .map_err(|error| eyre::eyre!(error))?;
+
+        tracing::info!(sessions=%sessions.len(), "waiting for active sessions to expire");
+        while sessions.is_not_empty() {
+            tokio::time::sleep(SESSION_TIMEOUT_SECONDS).await;
```
markmandel, line 128: Non blocking suggestion:

```suggestion
            tokio::time::sleep(Duration::from_secs(1)).await;
```
?

I just figured if we're checking every 60 seconds we could potentially shoot over by >=59 seconds from when we actually have all sessions expired. I figure no harm in checking one a secord to see if it's finally empty - and then you'll end up shutting down pretty quick after the session map is empty.

markmandel (approved):

PR #741, opened 2023-05-09
Move QCMP to a seperate port

XAMPPRocky:
This PR moves QCMP onto its own port and worker in the proxy, I initially I thought I needed this, but after I wrote I found a better solution in the client, but I decided to open the PR anyway as makes the implementation cleaner and in theory faster as it avoids allocating on the heap (though I haven't actually measured if this is noticeable). This has the added benefit of allowing operators to choose to disable the protocol by not exposing the port, or allowing operators to have QCMP only instances.

Also I removed the `shutdown_rx` in the downstream workers as that would cause the process to stop accepting packets after it shut down.

### DIFF
--- a/benches/throughput.rs
+++ b/benches/throughput.rs
@@ -34,6 +34,9 @@ fn run_quilkin(port: u16, endpoint: SocketAddr) {
 
         let proxy = quilkin::cli::Proxy {
             port,
+            qcmp_port: runtime
+                .block_on(quilkin::test_utils::available_addr())
+                .port(),
             ..<_>::default()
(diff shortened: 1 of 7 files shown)


markmandel:
> @markmandel 503 on the lib.rs again, should we just change it to `crates.io` to avoid this problem?

We ignore crates.io links as well, since they also don't like to be crawled. I'll mark this link as ignored as well in my prep for 0.7.0 PR, and update the User Agent we use as well.

markmandel (approved):
Looks good!

PR #742, opened 2023-05-12
Remove shutdown_rx from downstream loop

XAMPPRocky:
Part of having better `SIGTERM` handling, I noticed the shutdown could potentially still affect traffic because it shut down the main downstream socket recv loop


markmandel (approved):
I totally missed this PR. Yep - approved. Good find.

PR #746, opened 2023-05-26
Add ASN cardinality to packet metrics

XAMPPRocky:
This PR adds ASN cardinality to the packet processing metrics and adds a `packet_jitter` metric that measures the time between received packets.


markmandel (commented):
Only thing missing from this is some doc updates on the updated metric labels and such 😄

markmandel:
404 on https://lib.rs/crates/quilkin. That's not a thing (unless they are blocking bots)

XAMPPRocky:
Yep, I'll add docs so we can merge this soon, currently focused on some infra stuff and I needed an updated image to test with.

> (unless they are blocking bots)

Does the bot have a user agent? A lot of sites and Proxies will reject traffic if it doesn't have any kind of user agent.

markmandel:
> Does the bot have a user agent? A lot of sites and Proxies will reject traffic if it doesn't have any kind of user agent.

Yeah it does -- looks like it went through fine the second time. Just a weird blip.

docs/src/services/proxy/metrics.md
```diff
@@ -47,29 +47,33 @@ The proxy exposes the following general metrics:
...
+* `quilkin_errors_total{event, asn, ip_prefix}`
 
   The total number of errors encountered while reading a packet from the upstream endpoint.
 
 ## Session Metrics
 
 The proxy exposes the following metrics around sessions:
 
-* `quilkin_session_active{asn}{ip_prefix}`
+* `quilkin_session_active{asn, organization, cc, ip_prefix, prefix_entity, prefix_name}`
```
markmandel, line 76: What's "cc" here - it's the only items I couldn't find some docs on above.

markmandel (commented):
Looks good! Just one question on `cc`.

docs/src/services/proxy/metrics.md
XAMPPRocky: Ah, it's "country code" I'll update the field to be more clear.

markmandel (approved):

PR #744, opened 2023-05-23
Implement Agent service

XAMPPRocky:
This PR adds a new service called "Agent", the goal of this service to provide a service more designed for relay setups by only having configuration forwarding rather than also being a management server, and this also has a QCMP service that allows it singable, which will be used in to be able to act as pingable "beacon" as it were in a given cluster.

Following this PR I intend to make a follow-up that removes the `--relay` flag from the management server (as it's not really needed anymore with `agent`), and moves the proxy's QCMP implementation to be its own port using the same task the agent uses to bring those services in line with this one.

PR #752, opened 2023-06-09
[PR/Fixes] Agent Service / Dual Stack Socket

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

This fixes #744 so it passes tests, and does some other small cleanup. 

This includes:

* Dual stack UdpSocket
* Update Agent code to use Dual Stack Socket
* Fix shutdown channel in Agent test
* Upgrade to 50ms delay on ping test (sometimes failed for me after compilation at 10ms)

Also foundational work on #666.

Once this is merged, I can move on to using the DualStackLocalSocket throughout Quilkin, so we can finish #666.

Also, I rebased against main, so it created some mess (but I also fixed up the conflicts in #744) - apologies for that.

### DIFF
--- a/src/protocol.rs
+++ b/src/protocol.rs
@@ -16,10 +16,11 @@
 
 //! Logic for parsing and generating Quilkin Control Message Protocol (QCMP) messages.
 
-use std::net::SocketAddr;
-
 use nom::bytes::complete;
+use std::net::SocketAddr;
 use tracing::Instrument;
+
+use crate::utils::net;
 
 // Magic number to distinguish control packets from regular traffic.
 const MAGIC_NUMBER: &[u8] = b"QLKN";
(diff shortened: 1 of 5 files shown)


src/utils/net.rs
```diff
@@ -40,9 +52,69 @@ fn enable_reuse(sock: &Socket) -> io::Result<()> {
...
+        }
+
+        Ok(Self {
+            v4: socket_with_reuse_and_address((std::net::Ipv4Addr::UNSPECIFIED, port).into())?,
+            v6: socket_with_reuse_and_address((std::net::Ipv6Addr::UNSPECIFIED, port).into())?,
+        })
+    }
+
+    pub async fn recv_from(&self, buf: &mut [u8]) -> io::Result<(usize, SocketAddr)> {
+        let mut v4_buf = vec![0; buf.len()];
```
markmandel, line 81: Is there a better way than creating a whole new buffer each time? I can't reuse the same buffer in both places as Rust won't let you borrow something mutable more than once.

markmandel:
Looks like https://developers.google.com/protocol-buffers now redirects to https://protobuf.dev/ - I'll file a separate PR  to fix.

XAMPPRocky (approved):

src/utils/net.rs
XAMPPRocky: I think a reasonable first implementation would be to have two member buffers, and have the API return the slice (returning a slice should work here, but you might need to copy into a vec). This means slightly more initial memory, but will reduce the number of allocations.

markmandel: I like it. If I need to copy to a vec, I might also look at returning an enum on which buffer was populated, so we don't have to create a whole vec each time.

agones/src/xds.rs
```diff
@@ -120,7 +120,10 @@ filters:
         gs.metadata
             .annotations
             .get_or_insert(Default::default())
-            .insert(token_key.into(), base64::encode(token));
+            .insert(
+                token_key.into(),
+                base64::Engine::encode(&base64::engine::general_purpose::STANDARD, token),
```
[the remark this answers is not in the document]
markmandel: Couldn't find it -- but this is identical to `main` - just was after the update to all the dependencies.

src/protocol.rs
```diff
@@ -32,65 +33,62 @@ const DISCRIMINANT_LEN: usize = 1;
...
+            let mut v6_buf = vec![0; 1 << 16];
+            let mut output_buf = Vec::new();
+
+            loop {
+                tracing::debug!(%v4_addr, %v6_addr, "awaiting qcmp packets");
+
+                match socket.recv_from(&mut v4_buf, &mut v6_buf).await {
+                    Ok((size, source)) => {
+                        let received_at = chrono::Utc::now().timestamp_nanos();
+                        let contents = match source {
```
markmandel, line 54: Since we get a `SocketAddr` from `recv_from`, that actually ends up being a good indicator of which buffer to look at.

agones/src/xds.rs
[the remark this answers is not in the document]
markmandel: Oooh I see why this is the way it is. The function in utils is only `pub(crate)`, so it's not available to this test. I can update it to be `pub` in this PR?

src/protocol.rs
markmandel: This felt like a good approach to me, as then we don't have to create a vec on each packet either.

agones/src/xds.rs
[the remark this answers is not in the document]
XAMPPRocky: Ah right I forgot this is just for the integration tests, this is fine then, we shouldn't expose the utils module

src/protocol.rs
XAMPPRocky: Yeah I like this approach a lot 👍

markmandel: Sweet!

Anything I can do to make this more easily mergable?

I'm happy to update `ep/agent-service` to `main` and resolve conflicts so that this is more easily merged, if you are swamped with other things.

XAMPPRocky: Yeah I'm swamped right now, feel free to just push your commit to the PR

markmandel:
@XAMPPRocky I cleaned up the commits on both PRs, rebased #744 against `main`, and rebased this one as well.

Assuming it passes CI, this should be good to approve and merge into your PR, and then we can approve your PR, and everything should bubble up nicely.

PR #755, opened 2023-06-23
Update examples to work with upcoming 0.6.0

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind feature
> /kind hotfix

Updates all examples to work with new default ports and changes to configuration options.

Work on #725

### DIFF
--- a/examples/agones-xonotic-xds/xds-control-plane.yaml
+++ b/examples/agones-xonotic-xds/xds-control-plane.yaml
@@ -103,7 +103,7 @@ spec:
       annotations:
         prometheus.io/scrape: "true"
         prometheus.io/path: /metrics
-        prometheus.io/port: "9091"
+        prometheus.io/port: "8000"
     spec:
       containers:
         - name: quilkin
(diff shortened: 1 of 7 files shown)


examples/agones-xonotic-xds/xds-control-plane.yaml
```diff
@@ -103,7 +103,7 @@ spec:
       annotations:
         prometheus.io/scrape: "true"
         prometheus.io/path: /metrics
-        prometheus.io/port: "9091"
```
XAMPPRocky, line 106: I have been wondering if we should switch the admin port back to this since it is the default, but then again you have to specify it anyway,

XAMPPRocky (approved):

examples/agones-xonotic-xds/xds-control-plane.yaml
markmandel: Hunting around a bit, I _think_ prometheus will default to `80` (it's not the best documented) -- and looking around at a few different services, it seems like the port changes to all kind of things.

So no strong opinions on what the port ends up being.

PR #761, opened 2023-07-14
Unit and integration test debugging enhancements

### LINKED ISSUE #695: Print pod logs when steps fail in kubernetes integration tests

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

* Finer control over log levels and filters by allowing env filters in a new `enable_log(filter: impl Into<EnvFilter>)` implementation.
* On the `agones` integration side a new `debug_pods(client: &Client, labels: String)` function that will output both the Events and all the logs for any pods that match a given selector.
* Example of use within a currently failing `agones_token_router` agones integration test in PR #712.

Closes #695

The full log of a pod can end up being quite large. We'll see how it ends up looking in cloud build.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,11 +18,12 @@
 members = [".", "./macros", "./agones"]
 
 [workspace.dependencies]
-kube = { version = "0.83.0", features = ["derive", "runtime", "rustls-tls", "client"], default-features = false }
+kube = { version = "0.84.0", features = ["derive", "runtime", "rustls-tls", "client"], default-features = false }
 k8s-openapi = { version = "0.18.0", features = ["v1_25", "schemars"] }
 tokio = { version = "1.28.1", features = ["rt-multi-thread", "fs", "signal", "test-util", "parking_lot", "tracing"] }
 base64 = "0.21.0"
 tracing = "0.1.37"
+futures = "0.3.28"
 
 [package]
 name = "quilkin"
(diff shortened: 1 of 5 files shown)


markmandel:
@XAMPPRocky in case not obvious, this is a PR against your PR, so I could add necessary debugging to work out what's going wrong.

It gave me a bit of an idea now 😄 but I'm curious to see if it's obvious within the cloud build output.

XAMPPRocky (approved):

markmandel:
Hmnn, I was hoping the log output would wrap. Lemme see if I can break it up line by line.

markmandel:
Now outputting with each log line as a line. Lemme know which one you like more.

markmandel:
I think that's way better broken out by line.

🤔 Would it be better if I broke out these utilities into it's own PR off main, and then came back around to https://github.com/googleforgames/quilkin/issues/695 ? (I'm leaning that way).

Looks like the PR is having some impact on the /ready http endpoint (I've not dug in yet).

XAMPPRocky:
Yeah splitting it up makes sense though similarly, I'm still having trouble seeing what the cause is.

markmandel:
Not sure - I think I'll have to build it and see what the readiness probe returns.

I'll split out the PR now though. Closing this one.

PR #764, opened 2023-07-21
Fix: mdbook-variables failing on doc building.

markmandel:
**What type of PR is this?**
>
> /kind breaking

/kind bug

> /kind cleanup
> /kind documentation
> /kind feature
> /kind hotfix

This PR should fix both the development docs being updated, and also the daily rebuild of our build container images that are used in CI.

N/A

Also reported: https://gitlab.com/tglman/mdbook-variables/-/issues/3

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -16,7 +16,7 @@ jobs:
       - run: echo docs_dir=docs-build/${{ github.ref_name }} >> $GITHUB_ENV
       - run: echo documentation directory=${{ env.docs_dir }}
       - run: cargo install mdbook
-      - run: cargo install mdbook-variables --version=0.2.0
+      - run: cargo install mdbook-variables --locked
       - run: mkdir -p ${{ env.docs_dir }}
       - run: |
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Hmm, seems like it's no getting any response from the status checks 🤔

markmandel:
> Hmm, seems like it's no getting any response from the status checks thinking

Ooh, good chance for me to see if I can see if the CI is running.

I am wondering if for single PR's like this, it will take longer to merge, because it will run CI again (although that's not the worst thing).

markmandel:
We have a build running for the merge queue - I'm not 100% sure if the right statuses are required for the new format, but we'll see how it pans out.

markmandel:
Okay - so this is interesting. I've set the CI status for the CI-merge-queue (which is the trigger for merge queues) to be required, but not the CI system which runs on PR submission.

I'm not entirely sure if that's going to generally work, but I think it will for this PR. I can see the CI-merge-queue status running for the merge queue right now.

PR #770, opened 2023-08-08
Add `qcmp ping` command

### LINKED ISSUE #751: Add `qcmp` command
Now that I've been using QCMP for a while in production, I think there's room for quilkin to add a `qcmp` subcommand that allows you to easily debug and query QCMP. Right now I think the smallest version of this would be the following, which pings a endpoint at `address` and prints the round-time-trip, this would just make it very easy to test whether a proxy or beacon is reachable.

```
quilkin qcmp ping <address>
```

XAMPPRocky:
fixes #751 

I needed this for testing reachability, so I made it.

### DIFF
--- a/src/cli/qcmp.rs
+++ b/src/cli/qcmp.rs
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+use std::net::SocketAddr;
+
+use crate::protocol::Protocol;
+
+#[derive(Clone, Debug, clap::Subcommand)]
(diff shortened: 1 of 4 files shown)


src/cli/qcmp.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub endpoint: SocketAddr,
+    /// The number of pings to send to the endpoint (default: 5).
+    #[clap(short, long, default_value_t = 5)]
+    pub amount: usize,
+}
+
+impl Ping {
+    pub async fn run(&self) -> crate::Result<()> {
+        tracing::info!("starting ping task");
+        let socket = tokio::net::UdpSocket::bind((std::net::Ipv4Addr::UNSPECIFIED, 0)).await?;
```
markmandel, line 39: Since you know the `SocketAddr` should check if it's ipv4 or ipv6 and use the appropriate `UNSPECIFIED`.

(not sure if you can go from v6 ->, v6?)

Could possibly also wait for #666, which I'm working on right now, but since it's not listening for a long time, it's probably overkill.

src/cli/qcmp.rs
```diff
@@ -0,0 +1,117 @@
...
+                );
+            }
+            None => tracing::error!("no successful results"),
+        }
+
+        Ok(())
+    }
+}
+
+fn median(numbers: &mut [i64]) -> Option<i64> {
```
markmandel, line 99: Unit tests would probably be a nice idea.

markmandel (commented):
I think https://github.com/googleforgames/quilkin/blob/main/docs/src/services/proxy/qcmp.md would be a good place to drop a small note on this command.

markmandel (approved):

PR #771, opened 2023-08-09
Hack around empty responses to relay

XAMPPRocky:
This shouldn't actually be merged, it's just a hack to help me debug, why the relay occasionally has empty endpoints despite the agent having a endpoint.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -169,6 +169,11 @@ impl Config {
         tracing::trace!(resource=?response, "applying resource");
 
         let apply_cluster = |cluster: Cluster| {
+            if cluster.endpoints().count() == 0 {
+                tracing::warn!(?response, "empty endpoints");
+                return;
+            }
+
             tracing::trace!(endpoints = %serde_json::to_value(&cluster).unwrap(), "applying new endpoints");
             self.clusters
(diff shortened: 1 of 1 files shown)


markmandel:
So is that a "I should merge this" or "not merge this" 😁

XAMPPRocky:
Don't merge, I just wanted a image to deploy and test with.

Though I would appreciate some help with writing an integration test for the relay, ones like what we have for the management server should be able to catch this, essentially like one in every ten requests from the relay is blank, so it leads to periods of stuttering from packets getting dropped from the relay sending the load balancer bad data.

markmandel:
> Though I would appreciate some help with writing an integration test for the relay, ones like what we have for the management server should be able to catch this

Sounds good. Happy to write it or point you in the right directions.

If you want me to write it, I'd suggest writing up a quickstart, and I can use as the guide to make the integration test.

markmandel:
> I mean, I'm not going to say no to that 😄

Works for me, since then I also get to learn how it all works too 😄 

🤞🏻  I should wrap up #666 this week.

XAMPPRocky:
superseded by #773

PR #772, opened 2023-08-10
Add quickstart for the relay setup

XAMPPRocky:
This is the Quickstart for the relay setup, as mentioned this is still mostly the same as the xds setup, except with a few paragraph changes and additions to the `xds-control-plane.yaml` to launch a relay and a agent.


markmandel:
Sweet! Will finish off #666 (almost there!) and then pop over to this 👍🏻

markmandel:
Finally getting around to taking this for a spin and building out the integration tests.

I'm rebasing against `main` so I 'll make the tweaks to remove `localities` as well. Would you prefer updates direct to this branch, or should I submit a PR to your branch instead?

XAMPPRocky:
> Would you prefer updates direct to this branch, or should I submit a PR to your branch instead?

Whichever works best, I don't mind either way, but pushing to here seems like the easiest, don't worry about not rewriting the commits to get through the rebase.

markmandel:
Added a commit with fixes for the quickstart and associated docs. Please take a look and let me know what you think!

Html link testing in CI will pass once #807 has been merged.

docs/src/services/relay.md
```diff
@@ -1,9 +1,9 @@
 # Control Plane Relay
 
-| services | ports | Protocol |
-|----------|-------|-----------|
-| ADS | 7800 | gRPC(IPv4) |
-| CPDS | 7900 | gRPC(IPv4) |
+| services | ports | Protocol           |
+|----------|-------|--------------------|
+| ADS      | 7800  | gRPC(IPv4 OR IPv6) |
+| CPDS     | 7900  | gRPC(IPv4 OR IPv6) |
```
markmandel, line 6: One thing I noted when doing the examples is that it's not just `quilkin relay` it's `quilkin relay (agones|file)` (at least from the command prompt).

And in the example for Agones - we do run `quilkin relay agones`. I am wondering if that's because we still want to pickup the Filter ConfigMap at the Relay level, but the endpoint configuration at the Agent level?

But it does seem like we have the same CLI args shared between `quilkin relay` and `quilkin agent`, which may be confusing (Also, noting that `quilkin relay` with no extra args works will run, but I assume will skip any of the provider/agent specific logic.

For example, `quilkin relay agones --config-namespace="gameservers"` is valid, but I'm guessing that `--config-namespace` doesn't actually do anything? (maybe to get removed later?). Same for `quilkin relay file` ?

Just wanted to make sure I wasn't missing anything here when writing docs. I figure this is just an artifact of work in progress, but wanted to check.

XAMPPRocky:
@markmandel LGTM, I can't approve since it's my PR, so you or @koslib  need to approve for it to be merged.

docs/src/services/relay.md
markmandel: Yeah that's an interesting point. I don't hugely worry about the cli being dot seperated, but also don't hate it 😄 

But I'll also add some notes to the agent/quickstart around relay sub-command functionality so that is also clear.

Just so I make sure to get that right:

`quilkin relay --gameservers-namespace` .... that doesn't do anything does it? Since the relay isn't looking at Agones `GameServer` data locally? (or does it???)

I also assume `quilkin relay file <path>`  wouldn't actually do anything, since the full config would come from seperate `quilkin agents` - or would it grab the local file info and merge it with whatever else is coming in?

koslib (approved):

PR #773, opened 2023-08-10
Fix watch on clusters, by removing inner Arc

XAMPPRocky:
This should actually fix the issue with changes not being sent across three services, and removes the constant re-checking for changes by finally fixing the issue with watching clusters. The bug was devilishly subtle, the `Watch` struct watches for changes between two values and sends if they change, however `ClusterMap` had inner `Arc` and thus both the value being held and the value in the channel were always the same.

I also added a `ReadGuard` and split it into `read` and `write` methods, so that's more clear what the intent is.

### DIFF
--- a/src/xds/server.rs
+++ b/src/xds/server.rs
@@ -101,13 +101,14 @@ impl ControlPlane {
         tokio::spawn({
             let this = this.clone();
             async move {
+                let mut watcher = this.config.clusters.watch();
                 loop {
-                    tokio::select! {
-                        _ = this.config.clusters.has_changed() => {
-                            this.push_update(ResourceType::Endpoint);
-                            this.push_update(ResourceType::Cluster);
-                        }
+                    tracing::debug!(?watcher, "waiting for changes");
+                    if let Err(error) = watcher.changed().await {
+                        tracing::error!(%error, "error watching changes");
                     }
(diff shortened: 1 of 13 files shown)


src/xds/server.rs
```diff
@@ -328,9 +329,9 @@ impl AggregatedControlPlaneDiscoveryService for ControlPlane {
                     let timeout = tokio::time::sleep(std::time::Duration::from_secs(1));
                     tokio::select! {
                         _ = timeout => {
-                            tracing::info!("sending discovery request");
-                            crate::xds::client::MdsStream::discovery_request_without_cache(&identifier, &mut requests, crate::xds::ResourceType::Cluster, &[])
-                                .map_err(|error| tonic::Status::internal(error.to_string()))?;
+                            // tracing::info!("sending discovery request");
```
markmandel, line 332: Was this meant to remain commented out?

markmandel (approved):
Nice find!

There's some code that's commented out (see comment) that I'm not sure if it was meant to be that way, but merge when ready otherwise.

src/xds/server.rs
XAMPPRocky: Yes, they aren't needed anymore, but I went ahead and cleaned it up

PR #775, opened 2023-08-16
Locally listen on IPv4 and IPv6 addresses

### LINKED ISSUE #666: Listening on IPv4 and IPv6 addresses

markmandel:
**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

> /kind hotfix

Implemented the `DualStackLocalSocket` for the locally listening socket, and updated the code base and fixed several issues that existed for handling IPv6 addresses in the code base.

This included updating the `test_utils` framework to provide coverage in tests for both ipv4 and ipv6 by allowing for either an IPv4 or IPv6 address to be created from the framework.

Closes #666

Let me know what you think of the `Random` address type approach. Felt like the easiest way to get general coverage and force testing of both paths, without having to have some kind of dual-run test harness.

### DIFF
--- a/src/endpoint/address.rs
+++ b/src/endpoint/address.rs
@@ -133,7 +133,16 @@ impl FromStr for EndpointAddress {
         let host = url
             .host_str()
             .map(String::from)
+            .map(|value| {
+                // Strip ipv6 brackets
+                let len = value.len();
+                if len > 2 && value.starts_with('[') && value.ends_with(']') {
+                    return value[1..len - 1].to_string();
+                }
+                value
+            })
             .ok_or(ParseError::EmptyHost)?;
+
         // Infallible
         let host = host.parse::<AddressKind>().unwrap();
         let port = url.port().ok_or(ParseError::EmptyPort)?;
@@ -293,7 +302,11 @@ impl TryFrom<crate::xds::config::endpoint::v3::Endpoint> for EndpointAddress {
 
 impl fmt::Display for EndpointAddress {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}:{}", self.host, self.port)
+        if let AddressKind::Ip(IpAddr::V6(ip)) = self.host {
+            write!(f, "[{}]:{}", ip, self.port)
+        } else {
+            write!(f, "{}:{}", self.host, self.port)
+        }
     }
 }
 
--- a/src/test_utils.rs
+++ b/src/test_utils.rs
@@ -14,13 +14,11 @@
  * limitations under the License.
  */
 
+use rand::{thread_rng, Rng};
 /// Common utilities for testing
 use std::{net::SocketAddr, str::from_utf8, sync::Arc, sync::Once};
 
-use tokio::{
-    net::UdpSocket,
-    sync::{mpsc, oneshot, watch},
-};
+use tokio::sync::{mpsc, oneshot, watch};
 use tracing_subscriber::EnvFilter;
 
 use crate::{
@@ -29,6 +27,7 @@ use crate::{
     endpoint::{Endpoint, EndpointAddress, LocalityEndpoints},
     filters::{prelude::*, FilterRegistry},
     metadata::Value,
+    utils::net::DualStackLocalSocket,
 };
 
 static LOG_ONCE: Once = Once::new();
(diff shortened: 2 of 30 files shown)


markmandel:
Huh weird. Testing locally seems to work just fine, but when running inside the container test suite, it falls over.

At least I can replicate locally.

markmandel:
If I run the tests in the container with `--network=host`, everything passes! So I'm _thinking_ it's something to do with binding to 0.0.0.0 / [::] or something like that in the container? 🤔

markmandel:
Now I'm actually wondering if I can use a socket with `set_only_v6(false);` to be a dual stack socket, without all this extra work.... which may be why this is failing in (potentially) a slightly different kernel/network setup.

I'll take it for a spin and let you know how I go.

src/endpoint/address.rs
```diff
@@ -133,7 +133,16 @@ impl FromStr for EndpointAddress {
         let host = url
             .host_str()
             .map(String::from)
+            .map(|value| {
+                // Strip ipv6 brackets
+                let len = value.len();
+                if len > 2 && value.starts_with('[') && value.ends_with(']') {
+                    return value[1..len - 1].to_string();
+                }
```
XAMPPRocky, line 141: Do the brackets not work when passed in? 🤔 Isn't that part of the address?

src/test_utils.rs
```diff
@@ -44,14 +43,41 @@ pub fn enable_log(filter: impl Into<EnvFilter>) {
...
+fn get_address(address_type: &AddressType, socket: &DualStackLocalSocket) -> SocketAddr {
+    let addr = match address_type {
+        AddressType::Random => {
+            let mut rng = thread_rng();
+            // sometimes give ipv6, sometimes ipv4.
+            match rng.gen_range(0..2) {
+                0 => socket.local_ipv6_addr().unwrap(),
+                1 => socket.local_ipv4_addr().unwrap(),
+                _ => unreachable!(),
+            }
```
XAMPPRocky, line 72: You can simplify this by using `rand::random`.
```suggestion
            match rand::random() {
                true => socket.local_ipv6_addr().unwrap(),
                false => socket.local_ipv4_addr().unwrap(),
            }
```

XAMPPRocky (approved):
LGTM! Small nits and questions, but overall the approach is good.

markmandel:
> I think I tried that before, I believe that won't work cross-platform. Maybe we can find a compromise that's good enough.

That was my conclusion as well from research, but now I don't actually know. I'm trying some small experiments.

markmandel:
Just dropping research so I can also find it later:

* https://learn.microsoft.com/en-us/windows/win32/winsock/dual-stack-sockets

Seems to indicate that a dual stack socket _will_ work on windows, but the default for a socket is different on Windows (IPV6_V6ONLY is true by default on windows, but false on Linux).

So I think we can do this with one ipv6 socket -- but the testing changes through here will still be useful at least.

Making updates and will see how we go.

markmandel:
Initial tests with a ipv6 socket with `IPV6_V6ONLY` set to false look very promising! (saves us a `select` on each socket, so that is nice!)

But that helped me narrow down the issue - within a Docker container, it seems that it hates sending traffic to `[::]`.

I may just force it to run on `--network=host` and call it a day.

src/endpoint/address.rs
markmandel: If the brackets are still there in the host when it gets to

```
let host = host.parse::<AddressKind>().unwrap();
```

Then it parses it as as `AddressKind(Name("[::]"))` rather than `AddressKind(Ip(V6("::")))`, which is not what we want.

XAMPPRocky: Right, but I think a better solution would be to update the `FromStr` implementation of `AddressKind` to handle that instead of `EndpointAddress`, so that it's always correct, rather than `EndpointAddress::parse` and `AddressKind::parse` providing different results.

markmandel: Aaah gotcha. I will make that change 👍🏻

markmandel:
I've got work on this going on a separate branch (using purely ipv6 sockets). Will likely close this PR and resubmit, but just keeping it open since there's some feedback I want to make sure I capture in here.

markmandel:
New PR incoming using ipv6 sockets, shortly, closing this one to avoid the noise. Also incorporates the feedback in here.

PR #777, opened 2023-08-22
Temporarily retain endpoints which are in active use

XAMPPRocky:
I don't think should be the actual solution, but I needed to test it, and figure we can also use the PR to discuss how to solve the problem in a better way.

Consider the following infrastructure

proxy -> management server
proxy -> relay <- agent

If the connection to the relay or management server, or the agent is disrupted for some reason, and those services need to be restarted, there's a chance that the proxy or relay will receive a response before the agent or management server has received information from k8s, which would lead to the proxy clearing its all of its endpoints, which will lead to player's traffic being dropped for that period of time.

This PR changes that behaviour by adding a shared atomic counter to endpoints, and when it merges new configuration will retain the old endpoints that are currently in use. It will still replace their metadata so if we receive the same endpoint, but we'll keep any in-use endpoints for their session duration.


markmandel:
Hmnn. That's super interesting.

I actually wonder if we solve this in the same way we solve proxies not being available behind a loadbalancer if they have no endpoints.

https://github.com/googleforgames/quilkin/blob/1fc39541b55eac64f8c3c74c45e424da6a5834b8/src/admin.rs#L98-L106

Which is to say a relay, agent, or management server can't return `OK` from `/ready` unless it has some endpoints. So if the proxy does restart for some reason, it (in theory) won't get added back into the LB until it has some sort of configuration.

The downside is - if there are legitimately no Endpoints, then that can't be communicated to the proxies 🤔 so maybe that's not the best approach.

Unless there is some way on the relay / agent / etc layer to not return true from `/ready` until they can say "hey, I've gone and checked at least once from the config I need to check from" and therefore won't send data back up to the proxy?

Did that make sense? Basically I think we should manage this in the `'/ready` handler since that's what will determine if it's endpoint (and it's configuration) is available to the outside world.

src/endpoint/locality.rs
```diff
@@ -277,34 +278,79 @@ impl LocalitySet {
...
+                // If an endpoint already exists:
+                // - If `sessions` is zero then it is dropped.
+                // If that endpoint exists in the new set:
+                // - Its metadata is replaced with the new set.
+                // Else the endpoint remains.
+                //
+                // This will mean that updated metadata such as new tokens
+                // will be respected, but we will still retain older
+                // endpoints that are currently actively used in a session.
+                Entry::Occupied(entry) => {
```
markmandel, line 306: Is this dangerous? Basically if an endpoint is getting a malicious traffic and the solution the end user wants to implement is to drop the endpoint entirely to protect the internal resources -- doesn't this mean that that won't work while the malicious traffic is continuing?

XAMPPRocky: > Basically if an endpoint is getting a malicious traffic and the solution the end user wants to implement is to drop the endpoint entirely to protect the internal resources -- doesn't this mean that that won't work while the malicious traffic is continuing?

Well yeah, that is why we shouldn't accept this version. I'll write up my response to your other questions/proposed solution in a separate comment, but I think what we need is to be able to have different verbs here. We need to be able to distinct between "services are disrupted or a bug has caused a endpoint to be lost in the chain" where we want the load balancer to retain that endpoint because we don't want player's to be dropped mid-game, and "stop sending traffic to this endpoint because we want to stop traffic to it".

XAMPPRocky:
> Did that make sense? Basically I think we should manage this in the '/ready handler since that's what will determine if it's endpoint (and it's configuration) is available to the outside world.

So I don't like that approach for couple reasons.

- gRPC Connection won't initiate until `/ready` has completed. This does mean that there will be additional latency or spurious unresolvable connection failures between services. When games are allocated we should want to have that information distributed as quickly as possible.
- Works for management servers, not for relay. Even though it won't initiate until it has servers, in a multi-cluster setup, what we're more likely to run into is that one cluster goes down, and the load balancer won't be aware of that because it's getting its information from the relay rather than directly from the management server, it will just see an incomplete config.

markmandel:
> This could be having ["Delta xDS"](https://www.envoyproxy.io/docs/envoy/latest/configuration/overview/xds_api#config-overview-delta) replace our current xDS protocol, so that we're communicating changes, and as added benefit we're reducing the amount of traffic we're communicating as we scale.

Ultimately that's probably the better path, I think you are right. 

If a proxy hasn't receive the request to remove something, then it will just keep it. If the relay server crashes, and a agent behind that also happened to crash at the same time - the relay can't tell any of the proxies to delete something that it wasn't aware of in the first place.

On the management server / agent perspective - I would say it still make sense to have a `/ready` that only returns true one it has checked and parsed the initial state of whatever world it's looking at. That way at least it can't send out an empty config before it actually has information. That at least solves issues at that layer.

> gRPC Connection won't initiate until /ready has completed. This does mean that there will be additional latency or spurious unresolvable connection failures between services. When games are allocated we should want to have that information distributed as quickly as possible.

At the agent / management layer, I don't think this is a concern, because fast connections to systems that have bad data don't help anyone. We could block on the gRPC request until the initial check completes, but at that point we're introducing the same delay anyway? Or maybe a check at both layers is good, just to be extra sure. I'd lead more towards optimising more on correct information than speed of delivery for this type of scenario.

XAMPPRocky:
> We could block on the gRPC request until the initial check completes, but at that point we're introducing the same delay anyway?

Well actually I do like that better, because that delay is only on initialisation right? It wouldn't be whenever there's no endpoints. Even though I think you're right from the agent / management server perspective that it mostly doesn't matter, and what matters is that it's returning data. From the load balancer side, it's good to see stuff like "warning cannot connect to <management-server>" whenever it has zero endpoints, but having the check be until it receives the first update from k8s does prevent that from happening outside of startup.

XAMPPRocky:
Closing as superseded by #850 and #836

PR #784, opened 2023-09-09
add cli option for log formatting (Closes #531)

### LINKED ISSUE #531: Output JSON or pretty logs based on terminal
Currently when running Quilkin it will always print out the logs in JSON, this is great for machines, as it makes them much more machine-readable. However this comes at the cost of not being very human-readable, especially compared to the default pretty output that tracing gives us.

We should add a small check at startup that detects whether we are in a interactive terminal or not, if we are then we should output pretty format, otherwise default to JSON. Also to be sure, we should add a `--log-format` CLI argument that allows to specify `pretty` or `json` in case you as a user absolutely want it in one format or the other.

Baschtie:
**What type of PR is this?**
/kind feature

this pr adds an option to chose between the following formatting option:
* `auto` - default option, detects if running in an interactive terminal and will choose either "json" or "plain"
* `json` - outputs logs in json
* `plain` - uses default tracing_subscriber output
* `pretty` - uses _tracing_subscriber.pretty()_ output

<details>
  <summary>How the options look</summary>
  
### json
Bildschirmfoto 2023-09-09 um 19 10 04

### plain
Bildschirmfoto 2023-09-09 um 19 09 47

### pretty
Bildschirmfoto 2023-09-09 um 20 06 17

</details>

Closes #531

~Not sure about the naming, "detect" could also be "auto" or if "pretty" should be something like "debug". I am also not sure if the "ppretty" option is needed at all.~

*updated with renamed options*

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -19,10 +19,12 @@ use std::{
     sync::Arc,
 };
 
+use clap::builder::TypedValueParser;
 use clap::crate_version;
 use tokio::{signal, sync::watch};
 
 use crate::{admin::Mode, Config};
+use strum_macros::{Display, EnumString};
 
 pub use self::{
     agent::Agent, generate_config_schema::GenerateConfigSchema, manage::Manage, proxy::Proxy,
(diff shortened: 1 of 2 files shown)


src/cli.rs
```diff
@@ -68,6 +70,27 @@ pub struct Cli {
...
+    )]
+    pub log_format: LogFormats,
+}
+
+/// The various log format options
+#[derive(Copy, Clone, PartialEq, Eq, Debug, EnumString, Display, Default)]
+pub enum LogFormats {
+    #[strum(serialize = "detect")]
+    #[default]
+    Detect,
```
XAMPPRocky, line 87:
```suggestion
    #[strum(serialize = "auto")]
    #[default]
    Auto,
```

src/cli.rs
```diff
@@ -68,6 +70,27 @@ pub struct Cli {
...
+    #[strum(serialize = "json")]
+    Json,
+    #[strum(serialize = "pretty")]
+    Pretty,
```
XAMPPRocky, line 91:
```suggestion
    #[strum(serialize = "plain")]
    Plain,
```

src/cli.rs
```diff
@@ -68,6 +70,27 @@ pub struct Cli {
...
+    #[strum(serialize = "ppretty")]
+    PrettyPretty,
```
XAMPPRocky, line 93:
```suggestion
    #[strum(serialize = "pretty")]
    Pretty,
```

XAMPPRocky (commented):
Thank you for your PR! This looks good to me, however I would change `detect` to `auto`, `pretty` to `plain`, and `ppretty` to `pretty`. I think that's a much more clear naming convention for describing what each of those options do.

Baschtie:
agree and updated 👍

XAMPPRocky (approved):

PR #785, opened 2023-09-10
Refactor ClusterMap

XAMPPRocky:
This PR refactors our internal storage of transmission of clusters over xDS to be more focused on endpoints rather than mimicking Envoys layout. We weren't using basically any of it, and having all of those extra fields was creating a lot of extra bytes and made the encoding logic more complicated.

This PR moves to using a simpler map of locality -> endpoints, as that was the only part we were using, and uses our own proto types for xDS discovery. We still use the core discovery request and response wrapper, it's just the types inside are now Quilkin types rather than envoy types.

We don't have a xDS benchmark, but I wrote a small micro benchmark of just the time and size of discovery responses, and there's a significant improvement to using this implementation, being roughly 30%-250% faster depending on the endpoint amounts (being faster for smaller amounts of endpoints where the message size overhead is more noticeable), and being 12% smaller.

### Old
endpoints | time (seconds) | size (bytes)
-|-|-
0 | 0.001 | 53
100 | 0.005 | 5537
1000 | 0.028 | 55013
10000 | 0.117 | 550013
65535 | 0.721 | 3653593

### New
endpoints | time (seconds) | size (bytes)
-|-|-
0 | 0.001 | 53
100 | 0.002 | 4910
1000 | 0.011 | 48984
10000 | 0.085 | 489984
65535 | 0.539 | 3260352

### Diff
endpoints | time (faster) | size (smaller)
-|-|-
100 | 2.5x | 12%
1000 | 2.5x | 12%
10000 | 37% | 12%
65535 | 33% | 12%

```rust
#[test]
fn protobuf_test() {
    let measure = |count: u16| {
        let config = Config::default();
        if count != 0 {
            config.clusters.write().insert_default(crate::endpoint::LocalityEndpoints::from((0..count).map(|i| Endpoint::from((std::net::Ipv4Addr::UNSPECIFIED, i))).collect::<Vec<_>>()));
        }
        let response_creation = std::time::Instant::now();
        let response = config.discovery_request("", ResourceType::Cluster, &[]).unwrap();
        let time = response_creation.elapsed().as_secs_f64();
        let size = crate::prost::encode(&response).unwrap().len();
        println!("count: {count}, time: {time}, size: {size}");
    };

    (measure)(0);
    (measure)(100);
    (measure)(1000);
    (measure)(10000);
    (measure)(u16::MAX);
    panic!();
}
```

### DIFF
--- a/agones/src/xds.rs
+++ b/agones/src/xds.rs
@@ -181,7 +181,6 @@ filters:
             .as_mut()
             .map(|annotations| annotations.remove(token_key).unwrap());
         gameservers.replace(name.as_str(), &pp, &gs).await.unwrap();
-
         // now we should send a packet, and not get a response.
         let mut failed = false;
         for i in 0..30 {
@@ -196,6 +195,10 @@ filters:
                 failed = true;
                 break;
             }
+        }
+        if !failed {
+            debug_pods(&client, "role=proxy".into()).await;
+            debug_pods(&client, "role=xds".into()).await;
         }
         assert!(failed, "Packet should have failed");
     }
--- a/agones/src/sidecar.rs
+++ b/agones/src/sidecar.rs
@@ -77,10 +77,8 @@ filters:
         on_write: DO_NOTHING
         bytes: c2lkZWNhcg== # sidecar
 clusters:
-  default:
-    localities:
-      - endpoints:
-          - address: 127.0.0.1:7654
+  - endpoints:
+    - address: 127.0.0.1:7654
 "#;
 
         let config_map = config_maps
--- a/benches/throughput.rs
+++ b/benches/throughput.rs
@@ -30,7 +30,7 @@ fn run_quilkin(port: u16, endpoint: SocketAddr) {
         let runtime = tokio::runtime::Runtime::new().unwrap();
         let config = Arc::new(quilkin::Config::default());
         config.clusters.modify(|clusters| {
-            clusters.insert_default(vec![quilkin::endpoint::Endpoint::new(endpoint.into())])
+            clusters.insert_default([quilkin::endpoint::Endpoint::new(endpoint.into())].into())
         });
 
         let proxy = quilkin::cli::Proxy {
--- a/build.rs
+++ b/build.rs
@@ -33,6 +33,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         "proto/data-plane-api/envoy/type/metadata/v3/metadata.proto",
         "proto/data-plane-api/envoy/type/tracing/v3/custom_tag.proto",
         "proto/quilkin/relay/v1alpha1/relay.proto",
+        "proto/quilkin/config/v1alpha1/config.proto",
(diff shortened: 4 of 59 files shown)


markmandel:
Lemme know if you want me to take a look, but also reminding you that we have:

https://github.com/googleforgames/quilkin/blob/d8e93b60ab0be42d8ae7b020d21302345d6dfe7b/agones/src/lib.rs#L418-L425

In case you want to try debugging in here.

XAMPPRocky:
> In case you want to try debugging in here.

I added it but it doesn't seem to be working? 🤔

agones/src/xds.rs
```diff
@@ -57,6 +57,7 @@ mod tests {
     /// for this test, we should only have single Agones integration test in this suite, since they
     /// could easily collide with each other.
     async fn agones_token_router() {
+        quilkin::test_utils::enable_log("agones=debug");
```
markmandel, line 60: Depending on what aspect you want to debug, you will nee to also include a call to `debug_pods(client: &Client, labels: String)` to actually get logs and events about those pods.

For example, `debug_pods(client, "role=proxy")` will output debug information about the proxy instances. You can see the labels they have here:

https://github.com/googleforgames/quilkin/blob/c0d5ca5c1052366a097959711e6c75b2634befbb/agones/src/xds.rs#L374

(This would be good to add to the doc on debug_pods) 

https://github.com/googleforgames/quilkin/blob/d8e93b60ab0be42d8ae7b020d21302345d6dfe7b/agones/src/lib.rs#L418-L425 )

markmandel: Also looking at the build log, it's not failing because something failed to come up (as far as I can tell anyway), it's failing here:

https://github.com/googleforgames/quilkin/blob/c0d5ca5c1052366a097959711e6c75b2634befbb/agones/src/xds.rs#L201

So the existing `debug_pod` statements aren't going to help in this moment - they have to be conditionally put in.

agones/src/xds.rs
```diff
@@ -181,7 +182,7 @@ filters:
             .as_mut()
             .map(|annotations| annotations.remove(token_key).unwrap());
         gameservers.replace(name.as_str(), &pp, &gs).await.unwrap();
-
+        debug_pods(client, "role=proxy");
```
markmandel, line 185: My suggestion might be to change line 201 from

```rust
assert!(failed, "Packet should have failed");
```

to something like:

```rust
if failed {
  // proxies
  debug_pods(client, "role=proxy");
  // xds server
  debug_pods(client, "role=xds");

  panic!("Packet should have failed");
}
```

Then the condition can always stay in the test in case of future debugging requirements.

XAMPPRocky: SGTM

markmandel:
Ah, `parse_ipv6_endpoint` has `localities` in the config still

https://github.com/googleforgames/quilkin/blob/de4988956bdf27e6277b449324599b112fb86cb7/src/config.rs#L412-L424

markmandel (approved):

PR #791, opened 2023-09-18
fix timestamp warnings

XAMPPRocky:
This is preventing all other PRs from succeeding.

### DIFF
--- a/src/cli/qcmp.rs
+++ b/src/cli/qcmp.rs
@@ -70,7 +70,7 @@ impl Ping {
                 }
             };
 
-            let recv_time = chrono::Utc::now().timestamp_nanos();
+            let recv_time = chrono::Utc::now().timestamp_nanos_opt().unwrap();
             let reply = Protocol::parse(&buf[..size]).unwrap().unwrap();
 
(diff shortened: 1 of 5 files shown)


src/cli/qcmp.rs
```diff
@@ -70,7 +70,7 @@ impl Ping {
                 }
             };
 
-            let recv_time = chrono::Utc::now().timestamp_nanos();
+            let recv_time = chrono::Utc::now().timestamp_nanos_opt().unwrap();
```
markmandel, line 73: Is there any risk of `timestamp_nanos_opt()` returning `None` ?

XAMPPRocky: Only if people are still running Quilkin in a few hundred years. https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.timestamp_nanos_opt

> An i64 with nanosecond precision can span a range of ~584 years. This function panics on an out of range DateTime.
> The dates that can be represented as nanoseconds are between 1677-09-21T00:12:44.0 and 2262-04-11T23:47:16.854775804.

markmandel (approved):

Issue #673, opened 2023-01-02
Add Helm Charts


XAMPPRocky:
To make Quilkin installation easier on kubernetes, we should add support for helm for managing Quilkin in your clusters. Combined with #662 the proxy and management servers can work out of the box, with no more than `helm install quilkin`.

XAMPPRocky:
> Wherein --arguments lets you define what arguments to pass to the quilkin executable.

I think we should maybe go a bit higher level than having the user set the executable arguments. I think the options should be like `provider` and `mode`, and then we can have it so that when you set `provider=agones` it implies Agones RBAC permissions.

```
helm install quilkin --set mode=proxy --set provider=agones quilkin/quilkin
```

markmandel:
We could go that route - although I was trying to avoid the inevitable expansion of 100 helm options by providing the most flexible option up front (check out Agones helm configuration params -- https://agones.dev/site/docs/installation/install-agones/helm/#configuration 

It can explode quite fast.

So tradeoffs 😃

This sounds like an excellent topic to walk through some customer user journeys in our next monthly meeting, and list out what each one would create and how, and see what future plans we might envisage, and how that would work as well.

Even fun things like - should proxies be deployed with a `Service`  or no? (then what options on the Service?).

There will probably want to be a Service for an xDS provider, but again, is that private, public, something else?

As we work through, we might find something a bit more generic may work in our favour better (or not!)

markmandel:
Comments from community meeting - we'll aim for

```
helm install quilkin --set mode=proxy --set provider=agones quilkin/quilkin
```

And can expand it over time as more config options are requested.

kiennguyen2403:
Hi @markmandel @XAMPPRocky 
"Hi, I’d like to work on this issue. I’m new to the project. Is this issue still available?

PR #802, opened 2023-09-26
Add health checks for each service

XAMPPRocky:
This PR refactors the health checks, so that we have service specific health checks, such that a service only marks itself as ready if its provider has given a valid response. And the `agent` service is only healthy when it has an active connection from the relay, and will retry its connection if it hasn't heard from it in awhile.


src/xds/client.rs
```diff
@@ -297,6 +307,13 @@ impl AdsStream {
 
                         match next_response.await {
                             Ok(Some(Ok(ack))) => {
+                                runtime_config
+                                    .xds_is_healthy
+                                    .read()
+                                    .as_deref()
+                                    .unwrap()
+                                    .store(true, Ordering::SeqCst);
```
markmandel, line 315: With the current implementation, this will drop the a proxy connected to a management server in and out of the load balancer, since liveness is tied to readiness - and that seems less than ideal.

The proxy should still process packets even it's retrying the connection - up until the point where it has to drop entirely to restart.

As as note - with this change, if a backing xDS server goes down, this will take down all clients (proxies, relays, etc) that are connected with it when they eventually restart. Are we happy with that? That's a pretty big area of effect (this reads like a potential regional outage to me).

Or maybe this is more of a documentation issue? i.e. "make sure you set health checks considerably long enough to allow for the system to retry connections" ?

src/cli/admin.rs
```diff
@@ -0,0 +1,219 @@
...
+    fn handle_request(
+        &self,
+        request: Request<Body>,
+        config: Arc<Config>,
+        health: Health,
+    ) -> Response<Body> {
+        match (request.method(), request.uri().path()) {
+            (&Method::GET, "/metrics") => collect_metrics(),
+            (&Method::GET, "/live" | "/livez") => health.check_healthy(),
+            (&Method::GET, "/ready" | "/readyz") => check_readiness(|| self.is_healthy(&config)),
```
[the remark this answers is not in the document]
XAMPPRocky: > rather than making/assuming they are the same thing - since I would recommend we default to readiness being false, but liveness being true.

Maybe the code isn't clear, but they aren't the same thing. The `health.check_healthy()` check only operates on whether the program has panicked. Where as `Admin::is_healthy` is service specific readiness.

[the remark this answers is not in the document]
markmandel: Ooooh! You are right, I totally saw "healthy" and read that as _liveness_, since the `health.check_healthy()` has the same name.

In which case, let's rename things to they align properly, so something like:

`health.check_healthy(),` ➡️  `health.check_liveness()` 
`RuntimeConfig::is_healthy(...)` ➡️  `RuntimeConfig::is_ready(...)`

So then both use the right words for everything. Sound good?

src/xds/client.rs
markmandel: So just to reiterate this properly, now my understanding is more correct.

On initial creation of the Quilkin process we wait to be Ready until we wait for a successful connection to whatever backing process we are working with (xDS, K8s, whatever).

But if that connection drops, eventually all Quilkin instances handling that connection will take themselves out of the load balancer - which means, the whole system stops processing data - and we have a single point of failure that can cause a regional outage.

Is that okay?

For example: if you had a Proxy connected Relay, and the Relay goes down, should the proxies should continue to be processing data packets for as long as it takes for the Relay to come back up? Or should it eventually stop processing (Ready: false) all together because continuing to process data with stale configuration data would be bad?

Assuming we want a Proxy to eventually stop serving traffic if  a relay goes down, we have some good controls at hand, on the Kubernetes side.  Since we would probably want fast readiness on success, I expect we would advocate for a `periodSeconds` of 1 (the minimum), but then have a `successThreshold` of 1, but `failureThreshold` of something like 30 (or some other larger number) - so end users can control this.

I just want to talk this though, since the implications here can be wide, so we would need to document this really well.

XAMPPRocky: > Or should it eventually stop processing (Ready: false) all together because continuing to process data with stale configuration data would be bad?

It should eventually stop processing, because it will not be able to serve new servers. A proxy should have a long `gracefulTerminationSeconds` so that it has plenty of time to continue to serve existing servers. It's not just that processing with stale data would be bad, it will not be able to process any new traffic if it doesn't have a working provider.

> we have some good controls at hand, on the Kubernetes side.

Relatedly, this is an another place where have #673 would be useful. We'd be able to define good defaults for these thresholds, instead of requiring users to figure it out themselves.

koslib (approved):

src/xds/client.rs
markmandel: > It should eventually stop processing, because it will not be able to serve new servers.

Okay, cool - so we are definitely saying that **if a relay goes down, at _some point_ any connected proxies should stop serving traffic altogether**. (similar for other connected scenarios).

> Relatedly, this is an another place where have https://github.com/googleforgames/quilkin/issues/673 would be useful. We'd be able to define good defaults for these thresholds, instead of requiring users to figure it out themselves.

We can handle some of this with our examples, but we're definitely entering a land where Helm charts are starting to get useful (especially when combo'd with rbac rules etc).

I also feel like it's probably game dependent (most likely tied to session length), so documentation on when and where a Ready would end up being false, is very important here.

For example, if I have a game session that lasts <=30m, I would probably want my proxies to end up non-Ready after > 30m (maybe 35m?) to ensure any existing sessions have time to finish, but after that, if the proxy is not getting any new data, therefore keeping the proxy running becomes a risk, and it should be shut down at that point.

For something like a lobby server or something longer running, removal would probably end up being much longer I expect (a few hours? a day?).

> A proxy should have a long gracefulTerminationSeconds so that it has plenty of time to continue to serve existing servers and recover. 

`gracefulTerminationSeconds` has more to do with things like node shutdownds, upgrades, rolling update deployments etc -- so did you mean this in more of a "we should have more documentation around termination" (100% agree btw), or I may be missing a point here on liveness/readiness.

Also a great other reason for some suggested values / Helm charts, here as well.

markmandel (commented):
Only thing blocking for me then (after the recent discussion) is updates to documentation at https://googleforgames.github.io/quilkin/main/book/deployment/admin.html on readiness and liveness, with potentially some good default values.

I'm happy to take those values and port them into examples as well).

src/xds/client.rs
XAMPPRocky: > I also feel like it's probably game dependent (most likely tied to session length), so documentation on when and where a Ready would end up being false, is very important here.

Definitely, if we were building a helm chart, I would expose a `expectedSessionLength` variable, and use that as the basis of the calcuations we set for thresholds.

> For example, if I have a game session that lasts <=30m, I would probably want my proxies to end up non-Ready after > 30m (maybe 35m?) to ensure any existing sessions have time to finish

You probably want it to be at least 1.5x to 2.5x your session length, because you want to be account for variability in session length, and a session starting at the same time that the provider went down.

> so did you mean this in more of a "we should have more documentation around termination" (100% agree btw), or I may be missing a point here on liveness/readiness.


Sorry I meant the similarly named `terminationGracePeriodSeconds`.

docs/src/deployment/admin.md
```diff
@@ -34,15 +34,36 @@ The admin interface provides the following endpoints:
...
 
-Depending on whether Quilkin is run in Proxy mode i.e. `quilkin proxy`, vs an xDS provider mode, such as `quilkin
-manage agones`, will dictate how readiness is calculated:
+Readiness is service and provider specific, so based on what you're running
+there will be different criteria for a service to be considered ready. Here's
+a list of the criteria for each service an provider.
+
+| Service | Readiness                                                           |
+|---------|---------------------------------------------------------------------|
+| Proxy   | Management server is connected (or always true if config is static) |
```
markmandel, line 51:
```suggestion
| Proxy   | Management server is connected (or always true if config is static)  AND if there is more than one endpoint configured|
```

But otherwise, this looks good to go!

PR #815, opened 2023-10-10
Refactor sessions to use socket pool

XAMPPRocky:
This PR refactors how we handle upstream connections to the gameservers. When profiling quilkin I noticed that there was a lot of time (~10–15%) being spent dropping the upstream socket through its Arc implementation that happened whenever a session was dropped.

As I was thinking about how to solve this problem I also realised there was a second issue, which is that there is a limitation on how many connections Quilkin can hold at once, roughly ~16,383. Because after that we're likely to start encountering port exhaustion from the operating system, since each session is a unique socket.

This brought me to the solution in this PR, which is that while we need to give each connection to the gameserver a unique port, we don't need to give a unique port across gameservers. So I refactored how we create sessions to use what I've called a "SessionPool". This pools the sockets for sessions into a map that is keyed by their destination.

With this implementation this means that we now have a limit of ~16,000 connections per gameserver, which is far more than any gameserver could reasonably need.

### Future Work

Add a limitation of connections per gameserver, if for example, you know the most you're going to have is ~50 players, setting it to be something like 75 seems reasonable. This would help prevent a large scale DDoS from causing port exhaustion (though it would still flood the existing sockets, so it wouldn't prevent a DDoS.)

### DIFF
--- a/benches/throughput.rs
+++ b/benches/throughput.rs
@@ -56,6 +56,9 @@ static THROUGHPUT_SERVER_INIT: Lazy<()> = Lazy::new(|| {
 static FEEDBACK_LOOP: Lazy<()> = Lazy::new(|| {
     std::thread::spawn(|| {
         let socket = UdpSocket::bind(FEEDBACK_LOOP_ADDR).unwrap();
+        socket
+            .set_read_timeout(Some(std::time::Duration::from_millis(500)))
+            .unwrap();
 
         loop {
             let mut packet = [0; MESSAGE_SIZE];
@@ -74,6 +77,9 @@ fn throughput_benchmark(c: &mut Criterion) {
     // Sleep to give the servers some time to warm-up.
     std::thread::sleep(std::time::Duration::from_millis(500));
     let socket = UdpSocket::bind(BENCH_LOOP_ADDR).unwrap();
+    socket
+        .set_read_timeout(Some(std::time::Duration::from_millis(500)))
+        .unwrap();
     let mut packet = [0; MESSAGE_SIZE];
 
     let mut group = c.benchmark_group("throughput");
@@ -125,6 +131,9 @@ fn write_feedback(addr: SocketAddr) -> mpsc::Sender<Vec<u8>> {
     let (write_tx, write_rx) = mpsc::channel::<Vec<u8>>();
     std::thread::spawn(move || {
         let socket = UdpSocket::bind(addr).unwrap();
+        socket
+            .set_read_timeout(Some(std::time::Duration::from_millis(500)))
+            .unwrap();
         let mut packet = [0; MESSAGE_SIZE];
         let (_, source) = socket.recv_from(&mut packet).unwrap();
         while let Ok(packet) = write_rx.recv() {
@@ -142,6 +151,9 @@ fn readwrite_benchmark(c: &mut Criterion) {
     let (read_tx, read_rx) = mpsc::channel::<Vec<u8>>();
     std::thread::spawn(move || {
         let socket = UdpSocket::bind(READ_LOOP_ADDR).unwrap();
+        socket
+            .set_read_timeout(Some(std::time::Duration::from_millis(500)))
+            .unwrap();
         let mut packet = [0; MESSAGE_SIZE];
         loop {
             let (length, _) = socket.recv_from(&mut packet).unwrap();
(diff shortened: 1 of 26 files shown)


XAMPPRocky:
@markmandel I'm seeing key not found and permissions denied on the CI run, so I can't see what's failing.

```
<Error>
<Code>NoSuchKey</Code>
<Message>The specified key does not exist.</Message>
<Details>
No such object: quilkin-build-logs/log-efda5b5a-d699-4bd6-a652-b673102a1c24.txt
</Details>
</Error>
```

markmandel:
You can also have a look at the GitHub check: https://github.com/googleforgames/quilkin/pull/815/checks?check_run_id=17577335488, the log is also there 😄 (I turned it on a while ago).

XAMPPRocky:
@markmandel Great thank you, I would have tested locally, but Quilkin has broken on macOS since the dual stack, and I keep forgetting to make an issue about it (doing so now). So CI is my main way to test network integration now.

markmandel:
Oh that's really clever!

So if we still have two clients talking to GameServer1, each client from the proxy will it's own local port, but those local ports may overlap with the local ports two other clients are using to talk to GameServer2.

Nice!

XAMPPRocky:
> So if we still have two clients talking to GameServer1, each client from the proxy will have it's own local port, but those local ports may overlap with the local ports of two other clients are using to talk to GameServer2.

Yes, in fact, they're guaranteed to overlap (or mux) as we only allocate new upstream sockets when all sockets for a server are exhausted.

markmandel:
> > So if we still have two clients talking to GameServer1, each client from the proxy will have it's own local port, but those local ports may overlap with the local ports of two other clients are using to talk to GameServer2.
> 
> Yes, in fact, they're guaranteed to overlap (or mux) as we only allocate new upstream sockets when all sockets for a server are reserved.

Makes a lot of sense. Nice optimisation!

XAMPPRocky:
@markmandel CI seems stuck.

markmandel:
🤔 that's a 2h timeout, with `01:59:10` in `test-quilkin`. Looks like a test is getting stuck somewhere and not finishing. 

Does `cargo test` complete for you locally?

XAMPPRocky:
Yeah, the test it was stuck on was the benchmark and that's one that I didn't have a problem with.

I had to update a lot of tests because previously we were storing `0.0.0.0` in configuration but traffic never comes from that, so I had to map it to localhost

markmandel:
> I had to update a lot of tests because previously we were storing 0.0.0.0 in configuration but traffic never comes from that, so I had to map it to localhost

Oh that's fun - but also makes sense. Previously we didn't care what the source adress was - if the mapped port got the traffic, we knew where it was going.

In this world, the source address _really_ matters, since we need that to know where to onsend packets going from endpoint back to client.

> Yeah, the test it was stuck on was the benchmark and that's one that I didn't have a problem with.

Also makes me realise we should have some timeouts with failures in our benchmarks, rather than just blocking forever.

I am noting that the `throughput_benchmark` uses `FEEDBACK_LOOP` (just looking for difference to `readwrite_benchmark` which is working, but everything is using `127.0.0.1` as it's root address, so it should map correctly, I would have thought.

Nothing immediate otherwise stands out (assuming some weird ipv4/ipv6 mapping issue like you were stating), but can try running it locally on my linux machine and see if I can replicate.

markmandel:
I'm running tests locally, and I'm getting quite inconsistent failures in unit tests on each run.... which makes me feel like there's a race condition somewhere....

Here's a few different runs:

```
...
test cli::proxy::tests::run_client ... FAILED
...
test proxy::sessions::tests::same_address_uses_different_sockets has been running for over 60 seconds
```

```
test load_balancer_filter ... FAILED

failures:

---- load_balancer_filter stdout ----
thread 'load_balancer_filter' panicked at tests/load_balancer.rs:77:14:
called `Result::unwrap()` on an `Err` value: Elapsed(())
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```


```
...
test xds::tests::token_routing ... FAILED
thread 'xds::tests::token_routing' panicked at src/xds.rs:284:14:
```

My guess then on the benchmark is that it's falling into this same race condition where I expect data is being dropped/not routed correctly.

XAMPPRocky:
> which makes me feel like there's a race condition somewhere

I finally figured out, I was scouring the code, and using debuggers, and I couldn't find anything in the Rust code that was locking or blocking or anything, I was at wits end when I decided to just take the module and port it again to the code. That's when I realised that I allocated an extra socket on the same port for upstream sockets to send to, but it had no recv loop. So what was happening was that Linux was assigning data to that socket, and it was never being read.

koslib (approved):
lgtm

PR #836, opened 2023-10-23
Various improvements based on loadtests

XAMPPRocky:
A continuation of #777 this contains various improvements based on loadtests. This is separated because I attempted to add io-uring to #777 and would like to be able to have both images available (if I can ever get io-uring to compile).


markmandel:
I'm at a conference today/tomorrow - but I can have a look at why the Agones test is failing when I get back, if that's useful.

markmandel:
I assume this is more of a WIP PR, than something you want review on?

Just wanted to check real quick.

koslib (approved):

XAMPPRocky:
> I assume this is more of a WIP PR, than something you want review on?

Correct, this is just a collection of the changes I've made it to make it handle load better, it's all in one PR right now to make it easy to pull an image. Once I have more time the goal is to cherry-pick individual commits into their own PRs so they can each be reviewed separately.

Issue #734, opened 2023-04-04
The backoff for k8s watch doesn't reset if successful


andersroos:
**What happened**:

After analyzing some logs I think the retry mechanism for the k8s watch is flawed. The fix in https://github.com/googleforgames/quilkin/issues/731 made an improvement, but there is a deeper issue.

The problem is that the backoff doesn't reset when the watch starts successfully The watch can fail at times for various reasons. For example in our cluster I have seen this error a few times: `watch stream failed: Error reading events stream: error reading a body from connection: error reading a body from connection: timed out` It doesn't happen often, but for a long lived manager stuff like this will be problematic.

This is not a problem the first few times because the backoff is short. The problem is that if this happens enough times over a period of time there can be a 5 minute outage where no new game servers can be contacted because the watch retry is in backoff. I think the backoff should reset when there is a successful start of the watch. Right now it accumulates over the lifetime of the manager process.

**What you expected to happen**:

When the k8s watch fails I expect it to retry immediatly if there has not been a problem with the watch recetnly.

**Environment**:
- Quilkin version: b13c84f

XAMPPRocky:
Thank you for your issue! That definitely seems like a bug, it's definitely not intentional or intuitive behaviour.

PR #839, opened 2023-10-27
Make MAX_DELAY of tryhard configurable

zezhehh:
**What type of PR is this?**
/kind feature

The default `MAX_DELAY` for `Provider::task` (5min), is way too long for our project, so we'd like it configurable. 

Not sure if it's a good solution, but I'll appreciate any input.

### DIFF
--- a/src/cli/relay.rs
+++ b/src/cli/relay.rs
@@ -120,6 +120,7 @@ impl Relay {
                         Ok(())
                     }
                 },
+                crate::config::providers::MAX_DELAY,
             )))
         } else if let Some(Providers::File { path }) = &self.providers {
             let config = config.clone();
@@ -135,6 +136,7 @@ impl Relay {
                         crate::config::watch::fs(config, provider_is_healthy, path, None).await
                     }
                 },
+                crate::config::providers::MAX_DELAY,
             )))
         } else {
             None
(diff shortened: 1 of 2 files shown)


andersroos:
Would be a work around for this issue: https://github.com/googleforgames/quilkin/issues/734

If we can put the max delay at 1-2 seconds or so the issue wouldn't be noticable.

markmandel:
🤔 if it's a workaround for what breaks the experience - should we instead just hardcode the max retry delay value to something like 5 seconds? (or even 2 seconds - it's not a lot of QPS), rather than have to create a new CLI API that potentially needs to be maintained going forward?

This also seems like a bug that should be reported/fixed in https://github.com/EmbarkStudios/tryhard, yeah?

zezhehh:
> 🤔 if it's a workaround for what breaks the experience - should we instead just hardcode the max retry delay value to something like 5 seconds? (or even 2 seconds - it's not a lot of QPS), rather than have to create a new CLI API that potentially needs to be maintained going forward?
> 
> This also seems like a bug that should be reported/fixed in [EmbarkStudios/tryhard](https://github.com/EmbarkStudios/tryhard), yeah?

Either works for us. I'm just not sure if a shorter max_delay will break other people's current systems. We haven't had a chance to explore `tryhard` yet, might do it later. :D

markmandel:
@XAMPPRocky WDYT? I'm tempted to just drop this down to 2 seconds as a workaround, and then go hunt upstream to `try_hard` to confirm if it resets backoff on success.

XAMPPRocky:
@markmandel I agree

PR #841, opened 2023-11-01
Move admin server to separate OS thread

XAMPPRocky:
This is to prevent heavy load on the main tasks from saturating the workload and causing the admin server to become unresponsive, then triggering a termination from Kubernetes.


koslib (approved):
lgtm

PR #846, opened 2023-11-03
Minor cleanup

Jake-Shadle:
- Remove dirs2 and atty
- Cleanup deny config
- Use std::io::IsTerminal instead of atty
- Fix compile errors when `instrument` feature is enabled

**What type of PR is this?**
/kind cleanup

Fixes some extremely low hanging fruit after doing a `cargo deny --all-features check` and `cargo clippy --all-features`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,9 +18,21 @@
 members = [".", "./macros", "./agones"]
 
 [workspace.dependencies]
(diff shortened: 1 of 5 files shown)


XAMPPRocky (approved):

XAMPPRocky:
Hmm seems that test is flaky 🤔

PR #847, opened 2023-11-03
Remove built dependency

Jake-Shadle:
This was pulling in git2 and writing a lot of data that was never used
in the crate. This just changes build.rs to manually read the HEAD
commit information and set an env var.
AFAICT this is actually superior in every way to built since we no
longer have the extra dependencies, and we actually get up to date
information when the commit changes, which AFAICT was not the case with
built since it was not registering file changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -147,7 +147,6 @@ tonic-build = { version = "0.10.2", default_features = false, features = [
     "prost",
 ] }
 prost-build = "0.12.1"
-built = { version = "0.7.0", features = ["git2"] }
 protobuf-src = { version = "1.1.0", optional = true }
 
 [features]
--- a/build.rs
+++ b/build.rs
@@ -14,6 +14,60 @@
  * limitations under the License.
  */
 
+use std::io::{Error, ErrorKind, Result as IoResult};
+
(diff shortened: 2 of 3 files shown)


XAMPPRocky (approved):

XAMPPRocky:
These tests weren't flaky until today 😄

markmandel:
> These tests weren't flaky until today 😄

Yeah! It's being really grumpy! I wonder what's up with that. 

If you see a flake, please chuck a comment in the PR of a c/p paste of the flake in the PR, and I'll do the same - that makes it easier to search for common flakes through GitHub search and eventually resolve (also we delete old build failures, so it's harder to see the history).

XAMPPRocky:
> If you see a flake, please chuck a comment in the PR of a c/p paste of the flake in the PR, and I'll do the same - that makes it easier to search for common flakes through GitHub search and eventually resolve (also we delete old build failures, so it's harder to see the history).

It wasn't even the same test though which is odd, it was different tests.

markmandel:
> It wasn't even the same test though which is odd, it was different tests.

Could be Cloud Build is having a grumpy day.

PR #850, opened 2023-11-07
Move game traffic sockets to io-uring

XAMPPRocky:
There's a lot of sleeps added in the tests, because this requires spawning threads, the tests need to wait for the threads to spawn. We should probably have a better interface for tests to remove them.

### DIFF
--- a/src/cli/proxy.rs
+++ b/src/cli/proxy.rs
@@ -37,6 +37,7 @@ use crate::filters::FilterFactory;
 use crate::{
     filters::{Filter, ReadContext},
     net::{maxmind_db::IpNetEntry, xds::ResourceType, DualStackLocalSocket},
+    pool::PoolBuffer,
     Config, Result, ShutdownRx,
 };
 
@@ -137,9 +138,24 @@ impl Proxy {
 
         let runtime_config = mode.unwrap_proxy();
 
+        // The number of worker tasks to spawn. Each task gets a dedicated queue to
+        // consume packets off.
+        let num_workers = self
+            .workers
+            .map(|nz| nz.get())
+            .unwrap_or_else(num_cpus::get);
+
+        eyre::ensure!(num_workers > 0, "must use at least 1 worker");
+
         let (upstream_sender, upstream_receiver) =
-            async_channel::unbounded::<(Vec<u8>, Option<IpNetEntry>, SocketAddr)>();
-        let sessions = SessionPool::new(config.clone(), upstream_sender, shutdown_rx.clone());
+            async_channel::unbounded::<(PoolBuffer, Option<IpNetEntry>, SocketAddr)>();
+        let buffer_pool = Arc::new(crate::pool::BufferPool::new(num_workers, 64 * 1024));
+        let sessions = SessionPool::new(
+            config.clone(),
+            upstream_sender,
+            buffer_pool.clone(),
+            shutdown_rx.clone(),
+        );
 
         let _xds_stream = if !self.management_server.is_empty() {
             {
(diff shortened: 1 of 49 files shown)


build/ci/build-image/cloudbuild.yaml
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 
 steps:
+  - name: ubuntu
```
markmandel, line 16: Wrong cloudbuild.yaml, you want the one in the root of the repository. This one gets run on a chron once a day.

build/build-image/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM debian:bullseye
+FROM debian:bookworm
```
markmandel, line 15: Just looked at the error properly, it's failing in:

```
   Compiling fsevent-sys v4.1.0
   Compiling filetime v0.2.22
Some errors have detailed explanations: E0308, E0412, E0425.
For more information about an error, try `rustc --explain E0308`.
error: could not compile `io-uring` (lib) due to 30 previous errors
make: *** [Makefile:155: build-macos-binary] Error 101
```

Which is the compilation of the macOS binary. Is there a way to skip using this functionality on macOS? Is it supported on MacOS?

src/cli/proxy.rs
```diff
@@ -638,29 +699,19 @@ mod tests {
...
+        let (tx, rx) = async_channel::unbounded();
+        let (_shutdown_tx, shutdown_rx) = tokio::sync::watch::channel(());
 
-        proxy
-            .run_recv_from(&config, &sessions, shared_socket)
-            .unwrap();
+        let sessions = SessionPool::new(config.clone(), tx, shutdown_rx);
+
+        proxy.run_recv_from(&config, &sessions, rx).unwrap();
+        tokio::time::sleep(Duration::from_millis(500)).await;
```
markmandel, line 714: My suggestion here (if possible) would be to create a looping poll to check a condition (or run the whole test), with a sleep at the end of each loop, and break the loop on successful checking of the condition.

If I was doing this in Golang I would use https://pkg.go.dev/github.com/stretchr/testify/assert#Eventually but I've found a `for` loop in rust with a sleep at the end of it much easier since Rust closures and ownership are trickier to manage - and I've not been able to find a better way.

XAMPPRocky: > My suggestion here (if possible) would be to create a looping poll to check a condition (or run the whole test), with a sleep at the end of each loop, and break the loop on successful checking of the condition.

My preference would be that we return a signalling primitive. The problem isn't solvable with a looping poll because the issue isn't "the packet recv timeout expires before the system is ready", the problem is "the test sends the packet before the the socket is listening, so the packet is lost and it will never pass."

markmandel: I've definitely done tests (in other systems) where it's basically

* set check = false
* try 10 times
  * send packet to endpoint
  * timeout after 500ms on getting a packet
  * If successfully received, set check = true, and break
  * If not successful, go around the loop again.

So you get a polling operation that eventually passes, and generally isn't prone to hitting race conditions with a single sleep operation.

XAMPPRocky: My worry about that loop, is that it could mask some bugs, for the sake of passing race conditions, for example, it would be indistinguishable from a bug in the proxy that was dropping packets.

markmandel: Ah yep, I do see your point there. That's fair.

> My preference would be that we return a signalling primitive that we await, so the moment it's ready the test starts. 

That makes a lot of sense then - and shouldn't be impossible to do.

Are we thinking we keep the sleeps for now, and then adjust at a later date (or as we track potential flakiness in tests, and fix as needed?) - that would work for me.

XAMPPRocky: Yeah I think keeping the sleeps for now is best, because it will require changing a lot more and it's not the worst right now, I've already implemented part of it in this PR, but there's a lot more code for updating our test framework.

markmandel: SGTM!

Did you want to merge this now, or wait until after performance tests are done?

koslib (approved):

PR #851, opened 2023-11-10
Add datacenter discovery

XAMPPRocky:
This commit adds the concept of "datacenter" to our xDS protocol. Unlike other resources, there's slightly different behaviour between proxies to relays and relays to agents. When a relay queries for a datacenter from the agent, the agent returns its own QCMP port and ICAO location. The host then adds the address of remote agent to identify the datacenter. When the proxy then queries the relay for datacenters, it gets all of the relay's agent's addresses, ports, and ICAO codes.

This will be used with a future commit to measure the distance between proxies and agents.


XAMPPRocky:
Superseded by #854

PR #854, opened 2023-11-15
Implement Phoenix Network Coordinates

XAMPPRocky:
This commit uses the previous commit of datacentre discovery with an
implementation of Phoenix Network Coordinates[1], to build a
decentralised network coordinate map of all the datacentres that the
proxies know about, using that information, we expose a new HTTP server
at the same QCMP port that a client can use to query what datacentres
are the closest to the proxy.

For identifying datacentres we use ICAO codes, this allows us to
colnsildate multiple logical datacentres into a single location, for
example if you have multiple zones allocated in the same physical
datacentre, in order to reduce the amount of options a client has to
pick from in that case they're consolidated under their ICAO code.

This allows for clients to get their full latency to datacentres which
allows operators to a allocate datacentres dynamically, and in the
closest location available to players.


markmandel (commented):
Awesome - passing tests!

Would love some docs - this looks very cool 😄

Jake-Shadle:
Well, works as in, all tests pass locally for me, I have no idea why relay_routing fails in CI.

XAMPPRocky:
> Well, works as in, all tests pass locally for me, I have no idea why relay_routing fails in CI.

Seems like that is still spurious. I'll add you as the code owner, and add the docs for this feature, and then we can get it in.

XAMPPRocky:
> because the amount of just "random" timeouts throughout the tests just feels very fragile.

Agreed, though it would require a lot of refactoring that I'm not sure is worth it right now. Since it would require making a lot of initialisation and watch change notification channels and making those available to tests in a way that's also not hacky.

Jake-Shadle (approved):

Issue #837, opened 2023-10-23
Consider Alternatives To DashMap


XAMPPRocky:
(See: https://github.com/xacrimon/dashmap/issues/285)

While doing loadtests and benchmarking, I've noticed a significant amount of time is spent dropping Arcs from read or write references to DashMap. DashMap is great for concurrent access to multiple shards, but our usage is much more "access everything", where we iterate over the whole map, this causes a tonne of extra overhead that I worked around in a commit by just cloning the endpoints into a separate vec as a cache. 

We should consider alternatives to dashmap that are much more designed for our usage patterns.

https://github.com/googleforgames/quilkin/blob/0a2b4e67e146db0bd20e8f69240f61df8ddc9d90/src/net/cluster.rs#L62

xacrimon:
See https://github.com/xacrimon/dashmap/issues/285#issuecomment-2628831110 for some alternative suggestions based on your usage, though I only briefly skimmed over how it's used in quilkin. Should be an easy swap, not a lot of work to do.

xacrimon:
My suggestion notably makes other significant optimizations trivial. For example your endpoint iteration spends a lot of time collecting a vector of endpoints, just to return it as an iterator. This may be useful under dashmap to release locks earlier but with an `ArcSwap<HashMap>` you can simply use load the arcswap into a guard and use the `::map` method to "project" a `Guard<HashMap>` into a `Guard<impl Iterator>` , and since readers no longer block writers from swapping out the value, while the one you have stays the same, there's no need to perform any intermediate collections.

PR #856, opened 2023-11-17
ClusterMap benchmarks

Jake-Shadle:
During investigation of #837 (which fortunately? ended with the original issue not being observed) I added several benchmarks.

While getting familiar with the code, I noticed an easy win of `Locality` using 3 unnecessary heap allocations, now it uses 0, and no longer heap allocates during parsing. This also simplifies how it is serialized to/from json which gives slight speedups on both ends since it's not serializing as many fields. Really, locality is essentially an opaque identifier as far as the codebase is concerned, and it could probably be even more simplified to a single hash value, but that can be done separately, this was mainly just to see numbers change in the benchmarks.


XAMPPRocky (approved):

XAMPPRocky:
We just need to add an exception in the cargo-deny for xxhash's `BSL-1.0` (Boost Software Licence), that is allowed by Google and we already do the work of including licences in distribution because it's a requirement for Apache. https://opensource.google/documentation/reference/thirdparty/licenses#notice

markmandel:
> We just need to add an exception in the cargo-deny for xxhash's `BSL-1.0` (Boost Software Licence), that is allowed by Google and we already do the work of including licences in distribution because it's a requirement for Apache. https://opensource.google/documentation/reference/thirdparty/licenses#notice

Just making a note for myself - if you add the exception, this will add this notice to our licence.html document - but I do need to do the work to include it in our release .zip (which is easy enough to do) -- I'll write a ticket for this now so I remember to do it before the next release. The licence.html file is already in the Docker image.

Jake-Shadle:
Is all of that necessary? I can just take it out, I was only using it for testing purposes, and it's a dev only dependency so not part of any release artifacts.

markmandel:
> Is all of that necessary? I can just take it out, I was only using it for testing purposes, and it's a dev only dependency so not part of any release artifacts.

We already do 95% of the work already to stay in compliance with OSS licencing - so it's not a big step to add this extra piece. We should probably ship the licence.html anyway in the zip to be honest - so it's work I'll do before our next release anyway whether you want to keep the dependency or not. Up to you 👍🏻

XAMPPRocky:
Sorry, I was eager to bench with these improvements, so I pushed a commit to add the exception 😄

Issue #862, opened 2023-11-21
Look at replacing snappy with lz4


Jake-Shadle:
This is a TODO to investigate switching from snappy to lz4 for compression. Generally speaking (obviously, variables like size of data, data patterns etc affect this) lz4 compresses relatively the same speed (but usually slight better ratio) than snappy, and tends to decompress _much_ faster than snappy.

Jake-Shadle:
Or at least have it as an option to choose from.

markmandel:
> Or at least have it as an option to choose from.

👍🏻

I tried to build this filter so it was extensible to allow for a variety of compression options - I figured being able to choose from a variety of compression would be useful, since it depends!

PR #868, opened 2023-11-22
Add lz4 support

Jake-Shadle:
This adds support for LZ4 compression. Initial benchmarks with (seeded) randomized data shows that snappy is easily beating it on both compression and decompression, but thought I might as well PR it.

Resolves: #862

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1442,6 +1442,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "lz4_flex"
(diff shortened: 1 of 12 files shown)


markmandel (commented):
Except for those type issues, implementation looks good.

Some documentation on the new features here would round this off nicely: https://github.com/googleforgames/quilkin/blob/main/docs/src/services/proxy/filters/compress.md

PR #872, opened 2023-11-25
Make the ready check has endpoints OR has xDS connection

XAMPPRocky:


src/cli/proxy.rs
```diff
@@ -321,8 +321,9 @@ impl RuntimeConfig {
         self.xds_is_healthy
             .read()
             .as_ref()
-            .map_or(true, |health| health.load(Ordering::SeqCst))
-            && config.clusters.read().has_endpoints()
+            .map_or(config.clusters.read().has_endpoints(), |health| {
```
markmandel, line 324: Just triple checking this to make sure...

`xds_is_healthy` makes sure that the proxy has some configuration information from an xDS provider, which means that we don't have to rely here on `config.clusters.read().has_endpoints()` to not put a proxy behind an LB before it has received it's configuration information.

Assuming I got that right, this looks good to me. We just need to tweak https://googleforgames.github.io/quilkin/main/book/deployment/admin.html#ready to be representative of this change 👍🏻

markmandel (approved):

PR #873, opened 2023-11-25
Add xDS client version cache

XAMPPRocky:


src/cli.rs
```diff
@@ -259,6 +259,9 @@ mod tests {
     };
 
     #[tokio::test]
+    // Currently pretty flaky, we still have an agones integration test for this
+    // use-case that will catch it.
+    #[ignore]
```
markmandel, line 264: Making a note that I should look at this flake as well with the work I'm doing on #877 - is likely the same I expect.

XAMPPRocky:
Closing, since we now have delta xDS.

PR #875, opened 2023-11-25
Add pprof endpoint

XAMPPRocky:


markmandel:
How's this pprof endpoint doing? Performing better than the last version we had?

XAMPPRocky:
> How's this pprof endpoint doing? Performing better than the last version we had?

Yeah, this one works, and ready to merge, it does require that you have a writable tmpfs, so had to add some documentation around that.

markmandel (approved):

PR #881, opened 2023-11-30
Use bounded channels to prevent backpressure

XAMPPRocky:


src/cli/proxy.rs
```diff
@@ -148,7 +148,7 @@ impl Proxy {
         eyre::ensure!(num_workers > 0, "must use at least 1 worker");
 
         let (upstream_sender, upstream_receiver) =
-            async_channel::unbounded::<(PoolBuffer, Option<IpNetEntry>, SocketAddr)>();
+            async_channel::bounded::<(PoolBuffer, Option<IpNetEntry>, SocketAddr)>(15);
```
markmandel, line 151: Only question I have is where this number (15) came from? Should it be configurable? (Same as below with (4).

markmandel (approved):

src/cli/proxy.rs
XAMPPRocky: Right now it's a bit arbitrary as I just wanted a PR I could test with, the idea was that when under load it takes roughly a millisecond to process a downstream packet and ¼ of millisecond to process a upstream packet, so these buffer amounts mean it should take less 5ms to send new packets.

markmandel: Gotcha -- also curious, what are you seeing with the unbounded buffers? Memory increases I assume?

🤔 wondering if it makes sense to have bounded/unboanded channels be configurable. I wonder if there are scenarios where people will want all packets processed always, and scenarios where people say "if you get more packets than this this often, drop it" ?

XAMPPRocky: > Gotcha -- also curious, what are you seeing with the unbounded buffers? Memory increases I assume?

No I'm trying to improve the situation for high amounts of clients connected, if for example 100,000 clients sending packets every 50us, you're going to have a lot of packets build up in the buffer, meaning that it will take longer for the latest packets to reach the server.

>  I wonder if there are scenarios where people will want all packets processed always, and scenarios where people say "if you get more packets than this this often, drop it" ?

It might make sense, I've just written as the simplest way right now, because I didn't want to do all the plumbing without testing it first.

markmandel: > It might make sense, I've just written as the simplest way right now, because I didn't want to do all the plumbing without testing it first.

Makes perfect sense. Curious to see the report on how if it improves things 👍🏻

XAMPPRocky:
Don't think this actually helped after testing, so going to close for now.

PR #897, opened 2024-03-01
Fix health_server test

Jake-Shadle:
When running all tests in parallel with cargo nextest this would fail since the port is bound in multiple tests

### DIFF
--- a/tests/health.rs
+++ b/tests/health.rs
@@ -32,7 +32,10 @@ async fn health_server() {
     });
     t.run_server(
         server_config,
-        None,
+        Some(quilkin::Proxy {
+            qcmp_port: 9094,
+            ..Default::default()
+        }),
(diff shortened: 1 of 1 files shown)


tests/health.rs
```diff
@@ -32,7 +32,10 @@ async fn health_server() {
     });
     t.run_server(
         server_config,
-        None,
+        Some(quilkin::Proxy {
+            qcmp_port: 9094,
```
XAMPPRocky, line 36: If it's unused shouldn't it be `0`? that way it wouldn't ever conflict bound ports.
```suggestion
            qcmp_port: 0,
```

XAMPPRocky (approved):

Jake-Shadle:
Yah, so the reason the test was taking so long was because the metrics test also failed to bind to the port but that one didn't have a timeout and so the CI was just stuck permanently, I've gone through and (I think) made all of them ephemeral.

PR #899, opened 2024-03-01
Move non-linux builds to github

Jake-Shadle:
This allows CI to still check windows/macos compile, but actually run in parallel rather than serializing on each platform and wasting valuable CI time

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@
 # with auto-merge enabled and update it to the latest `main`, forming a best-effort queue
 # of approved PRs.
 
-name: Validate protobufs
+name: CI
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
Have to defer @markmandel here, I think there might be some google policy here that prevents this, otherwise I'm all for it.

markmandel:
I had a refresh of our current policies - and GitHub Actions is now definitely on the table 🙌🏻!

I would limit it to anything that doesn't maintain an artifact of some kind / interact with the Google Cloud infrastructure though. Using Cloud Build where there is an artifact, means there's a direct path between the Google Cloud infra - where the artifact gets built - and where it ends up being hosted (even if it ends up being a download on GitHub, it's going through a Google laptop), so we can lean on Google as a whole for security pipeline -- but for just testing, this is totally fine.

The only thought I have is that the current `build` step via Cloud Build also tests that the cross platform build works at release time. It's probably going to be fine / can be fixed at release time if something does break, but just something to consider.   Also since we release every couple of months, and we run CI all the time, the benefit definitely outweighs any extra time we spend at release.

So I'm happy to merge unless that is a bigger concern for anyone else?

XAMPPRocky (approved):

PR #909, opened 2024-03-15
Test refactor proposal

Jake-Shadle:
This is a first pass at showing a different way to write integration tests, ie, tests that use multiple components together.

1. An incremental step at changing APIs and tests either binding sockets ahead of time that are then passed to the APIs, or for the proxy case, not binding and using the port returned from run_server in most cases. While the available_addr function is still used, this is first step at completely removing it. While it works in most cases, it's actually not guaranteed to be 100% reliable as there is nothing stopping the OS of assigning the port to another call causing a bind failure for the second test to use it, so eventually removing this will entirely remove bind-related flakiness.
1. Moves the logic of spinning up the various components into separate pieces from the CLI interface to them. The goal of this change is to start to tease apart the user interface (CLI) portions from the actual code that is run, to make writing tests easier. This also completely removes the notion of unwrapping the `RuntimeConfig` from the `Admin` inside various places in favor of passing in the actual config for the component in a type safe manner. For instance previously running the manage command with a relay server set would panic as .unwrap_agent() was used inside MdsClient
1. Move tests out to a separate crate. This is done so that common test code can be placed in the library and new dependencies can be used etc so that (eventually) test helper code in test.rs can be removed from quilkin itself
1. Make "unit" tests integration tests. While iterating on quilkin I've noticed that the there is a lot of test code in the library that shouldn't be there, as it means changing unit tests needs to compile quilkin (including all the test code) instead of just being able to recompile and relink the single test binary. I'm not sure why it was done this way but I suspect it might be due to...
1. Add helper for test tracing. Currently the tracing-test dev dependency can be used to add a proc macro to a test function to get trace output to be outputted on test failure. Unfortunately this macro has no way to filter messages other than allowing only message from teh current crate, which makes integration tests useless as then no trace messages from quilkin itself will be captured, or else _no filtering_ either by crate nor level (the level used in the tracing-test is trace!) which results in _massive_ spam from eg hyper/h2 that makes test tracing from integration tests utterly useless. So I just made a simple trace_test! macro that makes it trivial to capture only the output we care about in tests to make debugging slightly more tolerable
1. The qt (quilkin-test) crate also introduces a proposed API to make writing integration tests easier, removing a _lot_ of boilerplate by creating various components with minimal necessary configuration that can then be hooked together with dependencies. An example of this is shown in the refactored relay_routing test in mesh.rs


markmandel:
Having a quick look before getting on a plane for GDC - but this is an interesting approach! All of this seems like an improvement.

> Make "unit" tests integration tests. While iterating on quilkin I've noticed that the there is a lot of test code in the library that shouldn't be there, as it means changing unit tests needs to compile quilkin (including all the test code) instead of just being able to recompile and relink the single test binary. I'm not sure why it was done this way but I suspect it might be due to...

Blame me, and my relative lack of experience with Rust vs y'all's at Embarks 😃

> While it works in most cases, it's actually not guaranteed to be 100% reliable as there is nothing stopping the OS of assigning the port to another call causing a bind failure for the second test to use it, so eventually removing this will entirely remove bind-related flakiness.

Binding to 0 will result in a conflict? In theory, that shouldn't be possible - or am I missing something here?

> Move tests out to a separate crate. This is done so that common test code can be placed in the library and new dependencies can be used etc so that (eventually) test helper code in test.rs can be removed from quilkin itself

Sounds good to me - I always struggled with the best way to do this.

> So I just made a simple trace_test! macro that makes it trivial to capture only the output we care about in tests to make debugging slightly more tolerable

Sounds excellent. Right now having to step into a test and rerun it locally after CI failure to get at deeper logs is not pleasant.

Only other note in case this gets reviewed and potentially merged while I'm at GDC - Apache headers on everything please 👍🏻 (just in case).

src/net/xds/client.rs
```diff
@@ -243,34 +244,32 @@ impl MdsClient {
...
                                 _ => break,
                             }
                         }
                     }
 
-                    agent.relay_is_healthy.store(false, Ordering::SeqCst);
+                    rt_config.relay_is_healthy.store(false, Ordering::SeqCst);
 
-                    tracing::warn!("lost connection to relay server, retrying");
+                    //tracing::warn!("lost connection to relay server, retrying");
```
XAMPPRocky, line 272: Why is this commented out?

test/src/lib.rs
```diff
@@ -0,0 +1,596 @@
...
+
+pub enum PailConfig {
+    /// Creates a UDP socket with an ephemeral port that will send received
+    /// packets to a channel
+    Server(ServerPailConfig),
+    // Creates an admin server that is applied to any following relay, proxy, or management pails
+    //Admin(AdminPailConfig),
+    Relay(RelayPailConfig),
+    Proxy(ProxyPailConfig),
+    //Management(ManagementPailConfig),
```
XAMPPRocky, line 117: Are these commented out because they are unused?

XAMPPRocky (approved):

test/src/lib.rs
Jake-Shadle: Yes, for now, I only did a few tests in this PR because it is already so big, and want to convert more in a separate PR.

src/net/xds/client.rs
Jake-Shadle: I was/am debugging some tests and there were a few extremely spammy trace lines, will add them back.

XAMPPRocky:
For @markmandel talked with Jake and we're going to merge this now, and do a followup PR for the one agones test, just to reduce merge conflicts and divergences.

PR #915, opened 2024-03-22
Make idle request logs debug

XAMPPRocky:


src/net/xds/client.rs
```diff
@@ -682,7 +682,7 @@ impl AdsStream {
                                 break;
                             }
                             Err(_) => {
-                                tracing::info!(
+                                tracing::warn!(
```
markmandel, line 685: Triple checking - is it a bad thing if this log shows up? 

Moving to warn will this will still show up in info level logging, and look to end users like it's a bigger deal than before -- should this also be debug, or is it actually something they should be aware of?

XAMPPRocky: Typo

XAMPPRocky, line 685:
```suggestion
                                tracing::debug!(
```

Issue #927, opened 2024-04-15
Datacenter map doesn't seem to be kept up to date in proxies


XAMPPRocky:
I noticed on the latest version, that a proxy had only one datacenter while the relay it was connected to had dozens, it seems like it's not updating correctly. In fact it has only a single datacenter, and it's not one that's found in the relay's configuration.

Scenarios to test

- Regular updates across existing connection
- Relay restarting and the proxies making a new connection.

PR #931, opened 2024-04-23
Don't run `add_host_to_datacenter`

Jake-Shadle:
This was copied from the non-delta version, but results in a bug.

Resolves: #927


XAMPPRocky:
Can you make a followup PR that adds a regression test, that makes sure in a relay setup with two agents, the proxy's datacenter map contains those two agents?

XAMPPRocky (approved):

PR #940, opened 2024-05-06
Add PortPolicy::None

XAMPPRocky:


markmandel:
Agones x-ref: https://github.com/googleforgames/agones/pull/3807

But that will merge soon, the implementation is good, just needs a feature flag to sit behind. Happy to merge this as is.

PR #942, opened 2024-05-07
Allow null ports and address fields in GameServer CRD

XAMPPRocky:


Jake-Shadle (approved):

markmandel (approved):
Thank you! I was in the process of upgrading the Agones test infra,  and ran into the same thing 👍🏻

XAMPPRocky:
It seems like the test infra is failing though 🧐 does it need to be upgraded at the same time?

markmandel:
> It seems like the test infra is failing though 🧐 does it need to be upgraded at the same time?

Oh! Maybe!

I wonder if that older version of Agones even had the "addresses" field. It's quite possible it didn't - it's quite old.

So two options:

1. We make `deserialize_null_default` code handle fields it can't find any data for.
2. I can pull this code into the Agones upgrade I'm working on right now.

#1 wouldn't be bad to have, but also happy to do #2, since the test infra is running Agones 1.33, which is almost a year old.

XAMPPRocky:
I'd prefer No. 2, I can use the image in the mean time.

markmandel:
> I'd prefer No. 2, I can use the image in the mean time.

I also can't find a good way to even check that the field data is missing 🤔  I'm not sure how `#[serde(default)]` does this.

The error that comes back from `let opt = Option::deserialize(deserializer)?;` is from https://docs.rs/serde/1.0.200/serde/de/trait.Error.html#method.missing_field and all errors are `Trait serde::de::Error` 🤔  so you can't differentiate without looking at the text of the error string.

markmandel:
Confirmed - combining your change with the Agones update passes locally! Will submit a combined PR, and update the cluster.

PR #950, opened 2024-05-14
Add test for hashed token router

XAMPPRocky:


Jake-Shadle (approved):

tests/hashed_token_router.rs
```diff
@@ -0,0 +1,102 @@
...
+    server_config.clusters.modify(|clusters| {
+        clusters.insert_default(
+            [Endpoint::with_metadata(
+                echo.clone(),
+                serde_yaml::from_str::<MetadataView<_>>(endpoint_metadata).unwrap(),
+            )]
+            .into(),
+        );
+
+        clusters.build_token_maps();
```
XAMPPRocky, line 57: Should this maybe happen through a watcher task in quilkin instead, to prevent a code path where the token maps aren't built from ever happening.

Jake-Shadle: Well it does create them automatically when endpoint sets are updated in a proxy, this function call is just exposed since tests/benches aren't doing the normal operation but I suppose I could change it so that they're either built like they are now, or else when accessed, the issue would be that if they're built on access in actual running code that would be expensive.

XAMPPRocky: Alright, I guess it's fine for now.

PR #953, opened 2024-05-14
Allow `Option::deserialize` to fail.

XAMPPRocky:
This allows the changes to work with older versions of agones.

### DIFF
--- a/src/config/providers/k8s/agones.rs
+++ b/src/config/providers/k8s/agones.rs
@@ -284,10 +284,10 @@ where
     T: Default + Deserialize<'de>,
     D: serde::de::Deserializer<'de>,
 {
-    match Option::deserialize(deserializer) {
+    Ok(match Option::deserialize(deserializer) {
         Ok(opt) => opt.unwrap_or_default(),
-        Err(_) => Ok(T::default()),
-    }
+        Err(_) => T::default(),
+    })
 }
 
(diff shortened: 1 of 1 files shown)


Jake-Shadle (approved):

src/config/providers/k8s/agones.rs
```diff
@@ -284,8 +284,10 @@ where
     T: Default + Deserialize<'de>,
     D: serde::de::Deserializer<'de>,
 {
-    let opt = Option::deserialize(deserializer)?;
-    Ok(opt.unwrap_or_default())
+    Ok(match Option::deserialize(deserializer) {
+        Ok(opt) => opt.unwrap_or_default(),
+        Err(_) => T::default(),
```
markmandel, line 289: We probably only want to fail if the field is missing, but may want to capture other errors.

Although serde doesn't give us a nice way to to do that other than check the error text for "missing field" (afaik).

https://docs.rs/serde/latest/serde/de/trait.Error.html#method.missing_field

https://github.com/serde-rs/serde/blob/3202a6858a2802b5aba2fa5cf3ec8f203408db74/serde/src/de/mod.rs#L287

Is it worth doing the extra for checking for this specific error type, and letting others through - or should we always fall back to default on any error silently?

XAMPPRocky:
This doesn't seem to actually fix it, so closing.

PR #954, opened 2024-05-15
Support for old agones

XAMPPRocky:
This is just a PR to get a version for older agones, shouldn't be merged.


XAMPPRocky:
Hmm why is agones still failing 🤔

Issue #955, opened 2024-05-15
Agent's Kubernetes provider should not depend on ConfigMap


XAMPPRocky:
Currently the agent process depends on a configmap to exist, this should not be the case as it's not used anywhere, we should update the provider to only look for the configmap on the k8s provider for the relay and manage components.

PR #957, opened 2024-05-16
Ignore config maps for agents

Jake-Shadle:
Resolves: #955

### DIFF
--- a/src/components/agent.rs
+++ b/src/components/agent.rs
@@ -55,12 +55,13 @@ impl Agent {
         }
 
         let _mds_task = if !self.relay_servers.is_empty() {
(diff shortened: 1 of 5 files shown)


XAMPPRocky (approved):
LGTM, I think in the future we should consider splitting providers by config data, and have components expose which config data they support. I think that would be both simpler from a user perspective, removes edge cases like this, and enables people to mix and match their desired configuration sources  E.g.

```
--gameservers.k8s=gameservers
--filters.k8s=quilkin
// OR
--gameservers.file=gameservers.yaml
--filters.file=filters.yaml
```

Issue #960, opened 2024-05-17
Changes To The ConfigMap in the Relay aren't being picked up by proxies


XAMPPRocky:
We've noticed that changing the configmap of filters in the relay doesn't seem to properly update the proxies configuration, like it should.

markmandel:
This may well be due to: https://ahmet.im/blog/kubernetes-inotify/

XAMPPRocky:
We don't mount the configmap as a file though, we discover and watch the map through the k8s API.

markmandel:
oh yeah! Good point. Okay, yeah that is weird.

PR #966, opened 2024-05-28
Fix relay listener updates

Jake-Shadle:
This was caused by `config.filters` being a `Slot<>`, which only allows one closure at a time to be notified when the value changes. The relay creates both a relay and management service, which both subscribed to changes, meaning the management closure was dropped and replaced by the relay one, which meant the delta stream would never send updates for filters unless the client did a refresh if the interval was surpassed.

Resolves: #960

### DIFF
--- a/crates/xds/src/server.rs
+++ b/crates/xds/src/server.rs
@@ -36,98 +36,115 @@ use crate::{
     ResourceType,
 };
 
-#[tracing::instrument(skip_all)]
-pub fn spawn<C: crate::config::Configuration>(
-    listener: TcpListener,
-    config: std::sync::Arc<C>,
+#[derive(Clone)]
+pub struct ControlPlane {
+    config: Arc<Config>,
     idle_request_interval: Duration,
-) -> io::Result<impl std::future::Future<Output = crate::Result<()>>> {
-    let server = AggregatedDiscoveryServiceServer::new(ControlPlane::from_arc(
-        config,
-        idle_request_interval,
-    ))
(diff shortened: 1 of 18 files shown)


crates/xds/src/server.rs
```diff
@@ -180,43 +172,42 @@ impl<C: crate::config::Configuration> ControlPlane<C> {
...
+                    subbed: false,
                 }
             }
         };
 
-        let mut cluster_rx = self.watchers[ResourceType::Cluster].receiver.clone();
-        let mut listener_rx = self.watchers[ResourceType::Listener].receiver.clone();
-        let mut fc_rx = self.watchers[ResourceType::FilterChain].receiver.clone();
-        let mut dc_rx = self.watchers[ResourceType::Datacenter].receiver.clone();
-
```
Jake-Shadle, line 211: Changes these from separate watchers to a single broadcast channel, much more easy to deal with.

src/config/slot.rs
```diff
@@ -46,6 +46,10 @@ impl<T> Slot<T> {
     /// Adds a watcher to to the slot. The watcher will fire whenever slot's
     /// value changes.
     pub fn watch(&self, watcher: impl Fn(&T) + Send + Sync + 'static) {
+        if self.watcher.load().is_some() {
+            panic!("watcher is already set for this slot");
+        }
+
```
Jake-Shadle, line 52: This ensures this particular bug doesn't accidentally happen again.

Issue #964, opened 2024-05-23
0.9.0 Release


markmandel:
I was holding off on setting up the 0.9.0 release, since there were several critical bug fixes in the pipeline.

Looks like that's all sorted now, so just checking if there's anything I should block on before kicking off this release?

XAMPPRocky:
No, kick it off :)

markmandel:
Only thought I have - do we care that we don't have docs for `HashedTokenRouter`  in [our guides](https://googleforgames.github.io/quilkin/main/book/services/proxy/filters/token_router.html)? Or is that more of an experiment that may eventually end up being the TokenRouter? In which case, better to leave it undocumented until testing is complete. I wasn't quite sure.

markmandel:
@Jake-Shadle with #965 merged, https://googleforgames.github.io/quilkin/main/book/services/xds.html#supported-apis is now out of date, yes?

Tracking documentation changes that will need to be made before a release can be cut.

markmandel:
Now https://github.com/googleforgames/quilkin/pull/980 is merged, just waiting on #970 doc updates.

markmandel:
Need these two implemented/merged before release. Everthing above is done/obsolete.

* [ ] https://github.com/googleforgames/quilkin/pull/982
* [ ] https://github.com/googleforgames/quilkin/issues/983

markmandel:
Okay, we have everything that was blocking release merged now! Any final objections before I kick this off @XAMPPRocky @Jake-Shadle ?

XAMPPRocky:
Nope

Jake-Shadle:
No

markmandel:
# Release 0.9.0

## Prerequisites

- [ ] Have at least `Editor` level access to `quilkin` Google Cloud project.
- [ ] Local gcloud configuration is pointing at the `quilkin` Google Cloud project. 
- [ ] Local git remote `upstream` points at `git@github.com:googleforgames/quilkin.git`.

## Steps

- [ ] Review that closed issues have appropriate tags for the changelog.
- [ ] Review that merged PRs have appropriate tags for the changelog.
- [ ] Run `git remote update && git checkout main && git reset --hard upstream/main` to ensure your code is in line
  with upstream.
- [ ] Update Cargo version for release
    - [ ] Edit the `version` field in `./Cargo.toml` and remove the `-dev` suffix.
[...]

PR #970, opened 2024-05-29
Remove state of world reference

Jake-Shadle:
See https://github.com/googleforgames/quilkin/issues/964#issuecomment-2136194423, this removes reference to sotw ADS which is no longer supported. The envoy naming for things is super confusing, I don't know if there is a better way to state this. Also removed 2 dead reference links that weren't used.

### DIFF
--- a/docs/src/services/xds.md
+++ b/docs/src/services/xds.md
@@ -24,35 +24,34 @@ In addition to static configuration provided upon startup, a Quiklin proxy's con
 Communication between the proxy and management server uses the [xDS gRPC protocol][xDS], similar to an [envoy proxy]. xDS is one of the standard configuration mechanisms for software proxies and as a result, Quilkin can be setup to discover configuration resources from any API compatible server. Also, given that the protocol is [well specified][xDS-protocol], it is similarly straight-forward to implement a custom server to suit any deployment's needs.
 
 As described within the [xDS-api] documentation, the xDS API comprises a set of resource discovery APIs, each serving a specific set of configuration resource types, while the protocol itself comes in several [variants][xds-variants].
-Quilkin implements the **Aggregated Discovery Service (ADS)** _State of the World (SotW)_ variant with gRPC.
+Quilkin implements the **Aggregated (Delta) Discovery Service (ADS)** variant with gRPC.
 
 ## Supported APIs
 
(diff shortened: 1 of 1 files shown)


docs/src/services/xds.md
```diff
@@ -24,36 +24,35 @@ In addition to static configuration provided upon startup, a Quiklin proxy's con
...
-  * Only [socket addresses] are supported on an endpoint's address configuration - i.e an IP address and port number combination. Configuration including any other type of addressing e.g named pipes will be rejected.
-  * Any [load balancing information][clapolicy] included in this resource is ignored. For load balancing, use [Quilkin filters][filters-doc] instead.
+  - The proxy uses these resources to discover information about endpoints like their IP addresses.
+  - Endpoints may provide [Endpoint Metadata][endpoint-metadata] via the [metadata][xds-endpoint-metadata] field. These metadata will be visible to filters as part of the corresponding endpoints information when processing packets.
+  - Only [socket addresses] are supported on an endpoint's address configuration - i.e an IP address and port number combination. Configuration including any other type of addressing e.g named pipes will be rejected.
+  - Any [load balancing information][clapolicy] included in this resource is ignored. For load balancing, use [Quilkin filters][filters-doc] instead.
```
XAMPPRocky, line 43: I don't think we use these types from envoy anymore, I think it would be good to include some info and a link to the types protobuf instead.

markmandel: The only other thing I found that might be worth playing with was https://github.com/matze/protoc-gen-mdbook - to generate docs from the proto files themselves.

Just an idea.

markmandel:
I've been playing with [`protoc-gen-doc`](https://github.com/pseudomuto/protoc-gen-doc) and it's really struggling with all our dependencies. I might have to try something else.

markmandel:
> I've been playing with [`protoc-gen-doc`](https://github.com/pseudomuto/protoc-gen-doc) and it's really struggling with all our dependencies. I might have to try something else.

Actually, I've got it working on the command line with straight protoc - lemme see if I can integrate it with `crates/proto-gen` so that when protos get generated, the documentation is generated along with it.

markmandel:
I'm digging through the xDS code right now - happy to take a stab at these changes in combo with #982, especially if you have other things on your plate, to round this work off.

XAMPPRocky:
@markmandel that would be great 🙂

markmandel:
👍🏻  on it.

markmandel:
Now we have #982 I think we can close this?

Issue #983, opened 2024-06-26
Memory leak in proxies


Jake-Shadle:
This was reported by an external user during a soak test showing node state in a cluster.

It's pretty clear that memory usage only increases over time as proxies proxy traffic, before getting OOM killed by the kernel/k8s/docker. The _likely_ culprit are the buffers used to hold the packet data, and either the buffers are not being returned to the pool because there is a reference being held for forever (tracing? profiling?), or else the reference/recycling code is broken altogether? Another option to consider would be a bug in dashmap where the map is mutated while it is also being read? Not sure since I haven't looked at the dashmap internals.

![image](https://github.com/googleforgames/quilkin/assets/2316028/3f50c78b-d43c-43b0-8874-4955af192a38)

Note that _all_ of the packets in the above screenshot are being dropped by the proxy as they have invalid routing tokens. This is most likely why we (Embark) haven't noticed this problem.

The issue also doesn't seem to exist in #962, but was introduced later, so the issue can hopefully be easily binary searched (eg with #984).

PR #986, opened 2024-07-03
Remove dynamic error properties

Jake-Shadle:
I'm opening this as a draft as I want to confirm if this resolves the issue for the original reporter, as I had difficulty solidly reproing the bug, but was able to in a certain scenario, that when repeated with the changes I made resulted in (basically) flat heap usage. This isn't a total fix, as I believe the issue is that in the case of invalid traffic, the case in question being _every_ packet being received generating an error due to it not matching an endpoint results in a pipeline error being sent here https://github.com/googleforgames/quilkin/blob/50d91e4167f9160c5be36178e47789fb7844dfc1/src/components/proxy/packet_router.rs#L189 The problem is that, under load, the task that receives those errors may be starved of CPU time, resulting in the both the channel queue growing unbounded (it is an unboundedchannel after all), as well as causing all of the memory allocated by the error itself (in the case of TokenRouter, the base64 encoded token string) to be kept until the error can be popped off the queue.

Eg. what it looks like before this change
image

Even if the errors were popped quickly enough, every _unique_ error would result in a hash map entry, which means in the token router case, random traffic would cause the heap to grow up to the size of the of the token length given enough traffic.

While investigating I ran into a bug with xDS filterchain where the server will send empty delta updates, so now the server doesn't send updates if the resource(s) haven't changed.

Also added 2 ways to generate unique tokens, one where _n_ tokens of length _l_ are generated and written to ./quilkin-test-config.yaml that can be used with `-c`, and the other is with a new `--to-tokens` arg on proxy that can do the same, assigning unique tokens of a specific length to each address in `--to`.

Resolves: #983

### DIFF
--- a/crates/gen.rs
+++ b/crates/gen.rs
@@ -1,0 +1,49 @@
+#!/bin/env/cargo run-cargo-script
+
+//! ```cargo
+//! [dependencies]
+//! base64-simd = "0.8.0"
+//! ```
+
+extern crate base64_simd;
+
+fn main() {
+    use std::io::Write as _;
+
+    let mut args = std::env::args().skip(1);
+
+    let count: u64 = args.next().unwrap().parse().unwrap();
+    let length: u8 = args.next().unwrap().parse().unwrap();
+    let length = length as usize;
+
+    let file = std::fs::File::create("./quilkin-test-config.yaml").unwrap();
+    let mut writer = std::io::BufWriter::new(file);
+
+    writeln!(
+        &mut writer,
+        r#"version: v1alpha1
+filters:
+  - name: quilkin.filters.capture.v1alpha1.Capture
+    config:
+        suffix:
+          size: {length}
+          remove: true
+  - name: quilkin.filters.token_router.v1alpha1.TokenRouter
+clusters:
+  - endpoints:
+    - address: 127.0.0.1:8078
+      metadata:
+        quilkin.dev:
+          tokens:"#
+    );
+
+    let mut s = String::new();
+    for i in 0..count {
+        s.clear();
+        base64_simd::STANDARD.encode_append(&i.to_le_bytes()[..length], &mut s);
+
+        writeln!(&mut writer, "          - '{s}'");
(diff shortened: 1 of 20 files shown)


Jake-Shadle:
I think really the best fix here is to actually change it so that there is no error channel, and instead either just do the error bookkeeping on the same thread and get rid of the channel altogether, _or_ get rid of the error bookkeeping altogether, as, AFAICT, this is just for convenience of getting log warnings that errors have occurred, but in a production scenario this is already covered by prometheus.

XAMPPRocky:
On holidays, so I've just read what you wrote and not dug into the code.

> get rid of the error bookkeeping altogether, as, AFAICT, this is just for convenience of getting log warnings that errors have occurred, but in a production scenario this is already covered by prometheus.

Not fully, the reason the logging exists, is that we can't include the full error context in prometheus, as that causes cardinality explosion. And it is useful to get specific warnings from filters such as the exact tokens, because having access to that lets you trace it to a specific gameserver, where as without that we can't trace where the error is happening if the error is on your side rather than a deliberate attacker, and it's probably the only context you have for where that packet is trying to be sent.

I do think however we should disable the bookkeeping when we are under load at runtime. That provides the best of both tradeoffs, when under normal operation, we get helpful error messages, while not wasting time on it when someone is trying to overload the system.

Jake-Shadle:
But with the current code we would have cardinality explosion in prometheus as well, one of the labels for the error metric  is the stringized form of the error.

But fair enough, I'll change it so that the bookkeeping is done in the thread the error occurs in to get rid of the channel aspect, and cap the storage of errors and condense the information if the cap is reached.

XAMPPRocky:
Yes, but correct me if I'm wrong but it's a string of the error variant, and doesn't include the full context. So all filter errors go under `FilterError` so there's a fixed set of cardinality there because it's just the variant not the full error.

Jake-Shadle:
Oops nope, you aren't wrong.

PR #987, opened 2024-07-04
Cap error heap usage

Jake-Shadle:
This is a rework of #986

This resolves an unbounded memory growth issue that can occur in a particularly hostile context, ie, every packet is invalid in some way, generating an error, and CPU usage is high enough that tasks, notably

https://github.com/googleforgames/quilkin/blob/50d91e4167f9160c5be36178e47789fb7844dfc1/src/components/proxy/packet_router.rs#L272-L295

do not consistently receive CPU time. In this context, where every received packet generates an error, which is then sent to an unbounded channel before being dequeued, the dequeue task might not run quickly enough to dequeue each error, causing both the memory allocated by the error, as well as by the queue, grow quickly enough that the process can be OOM killed.

This PR changes it so that instead of each error being sent to an unbounded channel, the errors are instead accumulated by the receive tasks themselves, and those accumulations being sent to the print task once they reach a cap, or a certain amount of time since the first error was pushed has elapsed to a **bounded** channel. If the bounded channel is full and too many errors are accumulated on a particular receive task, we drop the errors and note that it happened, rather than risk consuming too much heap memory.

Previously:

image

This PR:

image

While investigating I ran into a bug with xDS filterchain where the server will send empty delta updates, so now the server doesn't send updates if the resource(s) haven't changed.

Also added 2 ways to generate unique tokens, one where _n_ tokens of length _l_ are generated and written to ./quilkin-test-config.yaml that can be used with `-c`, and the other is with a new `--to-tokens` arg on proxy that can do the same, assigning unique tokens of a specific length to each address in `--to`.

Resolves: #983

### DIFF
--- a/crates/gen.rs
+++ b/crates/gen.rs
@@ -1,0 +1,49 @@
+#!/bin/env/cargo run-cargo-script
+
+//! ```cargo
+//! [dependencies]
+//! base64-simd = "0.8.0"
+//! ```
+
+extern crate base64_simd;
+
+fn main() {
+    use std::io::Write as _;
+
+    let mut args = std::env::args().skip(1);
+
+    let count: u64 = args.next().unwrap().parse().unwrap();
+    let length: u8 = args.next().unwrap().parse().unwrap();
+    let length = length as usize;
+
+    let file = std::fs::File::create("./quilkin-test-config.yaml").unwrap();
+    let mut writer = std::io::BufWriter::new(file);
+
+    writeln!(
+        &mut writer,
+        r#"version: v1alpha1
+filters:
+  - name: quilkin.filters.capture.v1alpha1.Capture
+    config:
+        suffix:
+          size: {length}
+          remove: true
+  - name: quilkin.filters.token_router.v1alpha1.TokenRouter
+clusters:
+  - endpoints:
+    - address: 127.0.0.1:8078
+      metadata:
+        quilkin.dev:
+          tokens:"#
+    );
+
+    let mut s = String::new();
+    for i in 0..count {
+        s.clear();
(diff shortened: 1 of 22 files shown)


crates/gen.rs
```diff
@@ -0,0 +1,49 @@
+#!/bin/env/cargo run-cargo-script
```
markmandel, line 1: Apache header please.

src/components/proxy/error.rs
```diff
@@ -0,0 +1,177 @@
+use std::{fmt, hash::Hash};
```
markmandel, line 1: Apache header please.

src/components/proxy/packet_router.rs
```diff
@@ -250,7 +258,7 @@ pub async fn spawn_receivers(
     upstream_receiver: DownstreamReceiver,
     buffer_pool: Arc<crate::pool::BufferPool>,
 ) -> crate::Result<Vec<Arc<tokio::sync::Notify>>> {
-    let (error_sender, mut error_receiver) = mpsc::unbounded_channel();
+    let (error_sender, mut error_receiver) = mpsc::channel(128);
```
markmandel, line 261: 👍🏻

src/cli/proxy.rs
```diff
@@ -46,6 +46,9 @@ pub struct Proxy {
     /// One or more socket addresses to forward packets to.
     #[clap(short, long, env = "QUILKIN_DEST")]
     pub to: Vec<SocketAddr>,
+    /// Assigns dynamic tokens to each address in the `--to` argument
```
markmandel, line 49: what's the delimiter if you want to append multiple tokens?

XAMPPRocky (approved):

PR #997, opened 2024-08-15
Update project to beta status

### LINKED ISSUE #804: Move project to Beta.
Let's move the project to beta!

It's used in production, but we will probably still break a couple of APIs before we go to stable.

XAMPPRocky:
fixes #804

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,9 +20,8 @@ having to custom build and integrate this functionality into their game clients 
 
 ## Project State
 
-Project is currently in *alpha* status, and is being actively developed. Expect things to break.
-
-Not to be used in production systems.
+Project is currently in *beta* status, meaning that it is being actively developed, and being used in production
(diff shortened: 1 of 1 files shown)


markmandel:
Do we also want to bring all our configuration files out of beta?

Or the project is in beta, but each config object (filter or otherwise) stays in alpha, since we might change it?

XAMPPRocky:
Yeah I think it makes more sense to upgrade configs individually when we think those modules are stable, and I don't see much point in changing the name which breaks current configuration just for having beta in there instead of alpha. We shouldn't change it unless we've changed the API or actually stabilising it.

markmandel:
SGTM!

Let's also include this in our release - I need to rerun the process anyway! I should have time tomorrow (I'm in Finland currently pre-devcom/gamescom, so we'll be around the same timezone for the next couple of weeks)

markmandel (approved):

PR #1012, opened 2024-09-03
Just use epoll for qcmp

Jake-Shadle:
This just uses the simple epoll QCMP loop for linux, io_uring gives no benefits as we are only ever doing a single recv or send request at a time. For now it just spawns a task on the main tokio runtime, so in high CPU usage scenarios this will affect the timing, but...that might be a good thing? Unsure.

### DIFF
--- a/src/codec/qcmp.rs
+++ b/src/codec/qcmp.rs
@@ -19,7 +19,7 @@
 use crate::{
     net::{
         phoenix::{DistanceMeasure, Measurement},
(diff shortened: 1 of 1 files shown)


Jake-Shadle:
FWIW this is not necessarily an attempt to fix the issue we saw with QCMP as I never once found a test failure either automatic or manually with the io-uring code, but this at least removes all the complexity to make any issue easier to track down.

markmandel (approved):
LGTM!

PR #1016, opened 2024-09-12
Support interop between v8 and v9+

Jake-Shadle:
This enables v8 proxies and relays communicate with v9+ proxies and relays. I'm not sure why the version of quilkin gets bumped to a -dev version each release, as it means we can't just do a point release of this for 0.9 easily, but I'll leave that to others to figure out.

### DIFF
--- a/crates/xds/src/client.rs
+++ b/crates/xds/src/client.rs
@@ -414,13 +414,11 @@ impl AdsClient {
         config: Arc<C>,
         is_healthy: Arc<AtomicBool>,
         notifier: Option<tokio::sync::mpsc::UnboundedSender<String>>,
-        resources: impl IntoIterator<Item = (&'static str, Vec<String>)>,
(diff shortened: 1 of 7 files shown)


XAMPPRocky (approved):

Jake-Shadle:
I'll have to add additional debugging for this, I thought it was flakiness but it's consistently happening on the same step of the test but the test output doesn't actually show logs for any of the components.

markmandel:
> I'll have to add additional debugging for this, I thought it was flakiness but it's consistently happening on the same step of the test but the test output doesn't actually show logs for any of the components.

If you aren't able to run the tests locally (or it can be useful)

https://github.com/googleforgames/quilkin/blob/339294270abab5d08c273f611913594d8b3cd965/crates/agones/src/lib.rs#L692

Is available to get logs and event data for groups of pods if you want to see what's happening behind the scenes.

PR #1023, opened 2024-10-07
perf: make filters synchronous

XAMPPRocky:
This is a test to check whether making the filters synchronous makes a difference to the performance under the new raw io-uring implementation.

### DIFF
--- a/src/components/proxy/io_uring_shared.rs
+++ b/src/components/proxy/io_uring_shared.rs
@@ -252,16 +252,18 @@ pub enum PacketProcessorCtx {
     Router {
         config: Arc<crate::config::Config>,
         sessions: Arc<crate::components::proxy::SessionPool>,
-        error_sender: super::error::ErrorSender,
-        /// Receiver for upstream packets being sent to this downstream
-        upstream_receiver: crate::components::proxy::sessions::DownstreamReceiver,
+        error_acc: super::error::ErrorAccumulator,
         worker_id: usize,
     },
     SessionPool {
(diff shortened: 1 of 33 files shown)


Jake-Shadle:
https://github.com/googleforgames/quilkin/blob/7a81ba0016f8c9d4ce9cf4687019b8fb57eb6dd0/src/components/proxy/io_uring_shared.rs#L272-L375 Needs to be changed as well, otherwise it is still spawning tasks and sending messages rather than processing in the loop.

Jake-Shadle:
Maybe I wasn't clear enough, the tasks should not be spawned at all since those are on the tokio runtime, I think just processing the packet inside the loop itself would _maybe_ make things better since it won't need to context switch out for something that should take a micro/nanoseconds.

src/components/proxy/io_uring_shared.rs
```diff
@@ -660,9 +648,12 @@ impl IoUringLoop {
...
-                                }
+                                process_packet(
+                                    &mut ctx,
+                                    &process_event_writer,
+                                    packet,
+                                    &mut last_received_at,
+                                );
 
                                 // Queue the wait for the processing of the packet to finish
                                 loop_ctx.push_with_token(
```
Jake-Shadle, line 659: This can be replaced with the `loop_ctx.enqueue_recv(buffer_pool.clone().alloc());` from below now since the packet is finished so there is no need for an extra loop iteration.

Jake-Shadle (changes requested):

Jake-Shadle (approved):

PR #1044, opened 2024-12-02
perf: Use local histogram for processing time

XAMPPRocky:
This should hopefully improve performance in the hot loop by only performing atomic ops related to metrics every 15 seconds.


src/components/proxy/io_uring_shared.rs
```diff
@@ -478,6 +482,8 @@ impl IoUringLoop {
                 // onto the submission queue for the loop to actually function (ie, similar to await on futures)
                 loop_ctx.sync();
 
+                const FLUSH_DURATION: std::time::Duration = std::time::Duration::from_secs(15);
```
Jake-Shadle, line 485: Naming is a bit confusing, `FLUSH_INTERVAL` makes more sense IMO.

src/metrics.rs
```diff
@@ -282,6 +282,23 @@ pub trait CollectorExt: Collector + Clone + Sized + 'static {
...
+}
+
+impl ProcessingMetrics {
+    pub fn new() -> Self {
+        Self {
+            read_processing_time: processing_time(READ).local(),
+        }
+    }
+
+    pub fn flush(&self) {
```
Jake-Shadle, line 297:
```suggestion
    #[inline]
    pub fn flush(&self) {
```

Jake-Shadle (approved):

PR #1052, opened 2024-12-17
Add initial XDP skeleton

Jake-Shadle:
This adds the initial XDP integration skeleton to make future PRs easier to review (hopefully). This PR doesn't "do" anything, but is meant to be small and easier to digest, future PRs will be more substantial.

There are two new crates:

### `crates/ebpf`

This is the actual eBPF program that is loaded into the kernel and attached to a NIC so that it can determine if packets received need to be routed to quilkin or not. This is in a separate workspace because its build requirements are different from the rest of this repo, namely that it requires a different target, `bpfel-unknown-none`, and nightly due to needing build the `core` crate. For now, there is a bash script that can be used to build it, but this first pass just puts the final binary artifact in the repo since it is quite small and means CI has to do less work, but can be built from source later if we really want that.

### `crates/xdp`

This is a shim crate that basically wraps the two main crates involved with XDP, `aya`, which is the external rust crate used to load and manage eBPF programs, and `xdp` the crate I've been working on that is used to create and handle AF_XDP sockets in userspace. The idea with having this separate crate is that quilkin itself will have a smaller surface area for interacting with XDP itself so that it's easier to make changes/remove it in the future and just generally reduce churn in quilkin itself.


crates/xdp/src/lib.rs
```diff
@@ -0,0 +1,174 @@
+use xdp::nic::NicIndex;
```
markmandel, line 1: Apache header please.

Cargo.toml
```diff
@@ -82,7 +82,8 @@ test = false
 [dependencies]
 # Local
 quilkin-macros = { version = "0.10.0-dev", path = "./crates/macros" }
-quilkin-xds = {version = "0.10.0-dev",path = "crates/xds" }
+#quilkin-xdp = { version = "0.1.0", path = "crates/xdp" }
```
markmandel, line 85: Nit:

```suggestion
```

XAMPPRocky (approved):

PR #1055, opened 2024-12-23
Add build support for arm64-linux

### LINKED ISSUE #611: Publish linux/arm64 binary and image

When using arm64 (Linux or mac M1) it would be nice to have a native binary and image that didn't run under emulation.

Also, ARM nodes on GKE are a thing, so it would be handy to be able to run on them too.

* Publish an linux/arm64 binary.
* Make the quilkin release image multi-arch and have images for both linux/amd64 and linux/arm64 under the same name and tag.

Don't do it. But it would be handy, so let's do it. Also shouldn't be too hard, the docker image is pretty straight forward.

**Additional context**

* https://docs.docker.com/build/building/multi-platform/
* Should wait for #608 to get it's upcoming PR, since there is a lot of cross compilation work in there.

kaboing:
Adds support for arm64-linux.

Uses buildx to build multi-platform image.

Spawned by: #611

**What type of PR is this?**
>
> /kind breaking
> /kind bug
> /kind cleanup
> /kind documentation

/kind feature

Adds support for arm64/linux. Builds multi-platform using buildx.  

Closes #611 


markmandel:
Was `make build` working for you locally?

kaboing:
Yeah, it works on my machine ™️ . I build on an apple m2 pro with docker-desktop. 

I noticed that the error messages complains about building `ring`, but I don't know if that is relevant.

markmandel:
> Yeah, it works on my machine ™️ . I build on an apple m2 pro with docker-desktop.

Ah fun, so you would be building on ARM, and Cloud Build is building on Intel.

My suggestion would be to add the package `g++-aarch64-linux-gnu` to 

https://github.com/googleforgames/quilkin/blob/00345bec721076fa683ee01efce1db93341f1972/build/build-image/Dockerfile#L28-L31

Which should give the system what it needs.

kaboing:
Gotcha. Let's test this version then!

XAMPPRocky:
> Step #5 - "build": ERROR: multiple platforms feature is currently not supported for docker driver. Please switch to a different driver (eg. "docker buildx create --use")

kaboing:
Updated PR to use docker-container as build driver. However, according to docs, 
> Unlike docker driver, built images will not automatically appear in docker images and [build --load](https://docs.docker.com/reference/cli/docker/buildx/build/#load) needs to be used to achieve that.

Docs for `build --load`
> Shorthand for [--output=type=docker](https://docs.docker.com/reference/cli/docker/buildx/build/#docker). Will automatically load the single-platform build result to docker images.

=> does not seem to work for multi-arch images. So we perhaps need to restructure a bit to push directly, or configure some other reasonable output. Any thoughts on this?

markmandel:
> does not seem to work for multi-arch images

In my experience the multi arch image is stored temporarily in the build cache, so a subsequent `build` with `--push` will be almost immediate, because it won't rebuild.

markmandel:
Well it built! But the image could not be found after the fact.

kaboing:
Well, it's progress at least! And I see in the logs that it build for both architectures, so that's nice.
When trying `--load` locally (latest commit), the entire thing hanged on sending the resulting tarball, but let's at least try it in the gcb environment and see what happens.

XAMPPRocky:
So it successfully built, but the image uploaded to CI doesn't look multi arch to me.

```
docker buildx imagetools inspect us-docker.pkg.dev/quilkin/ci/quilkin:0.10.0-dev-0aa6b83

Name:      us-docker.pkg.dev/quilkin/ci/quilkin:0.10.0-dev-0aa6b83
MediaType: application/vnd.docker.distribution.manifest.v2+json
Digest:    sha256:5a56ba3b94663f46b4e0aa673735c662ad474c21dcb6a1ad908000f75b474985
```

kaboing:
No, I concur. One of the other manifests uploaded to the registry at the same time do have different entries for arm64 and amd64: 
https://console.cloud.google.com/artifacts/docker/quilkin/us/ci/quilkin/sha256:d9d655c2d817e95d85b56286cee74a633cb1efb6d32650fdcb145702454485aa;tab=manifest?inv=1&invt=AbmXyg

```
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.oci.image.index.v1+json",
   "manifests": [
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "digest": "sha256:3a22f7b558eb554c10c351db637d438803472ac59a6af5e5fd6c4ba357c3df3c",
         "size": 3895,
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "digest": "sha256:4f2a123d6da1396fd987f71ae63f7597c180c76bde2c78199aec5d017f27cff1",
         "size": 3895,
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      ...
}
```

I'm trying to figure out what more needs to be done. 

Just to make sure, I've tried to run the image created on arm64 and it does not work.

XAMPPRocky:
Seeing the same result unfortunately

```
docker buildx imagetools inspect us-docker.pkg.dev/quilkin/ci/quilkin:0.10.0-dev-09bbe46
Name:      us-docker.pkg.dev/quilkin/ci/quilkin:0.10.0-dev-09bbe46
MediaType: application/vnd.docker.distribution.manifest.v2+json
Digest:    sha256:cf0e208da1ecd98500284826b1cfc20a8b04e3d8f857dc36d3d7a13073c550aa
```

kaboing:
Thanks for running the gcbruns for me on this PR, I'll investigate further where the problem is.

kaboing:
This digest should have the tag, it is the manifest list that points to the other manifests. Weird.
CleanShot 2025-01-17 at 16 14 59@2x

CleanShot 2025-01-17 at 16 16 43@2x

maggie44:
Is this how a working build and deploy ready for merging? Happy to jump in and take a look if not.

kaboing:
Yes, it's been tried and it ended up with the manifest list being wrongly tag. The explicit creation of manifest list was created in the hopes of fixing that problem, but we've ended up in the same situation. 

(also, usual caveats, I might have stumbled earlier and the proposed solution might work, but as far as I know, it's been tested)

XAMPPRocky:
Thank you for your PR, I've now implemented an updated version for our current infra setup on github actions.

PR #1060, opened 2025-01-14
refactor: Remove pipeline logging

XAMPPRocky:
This isn't as needed anymore, and creates a lot of noise in the logs. I think we could maybe bring the info back under a HTTP endpoint, but for now we can pull this part out.

### DIFF
--- a/src/components/proxy/packet_router.rs
+++ b/src/components/proxy/packet_router.rs
@@ -87,7 +87,16 @@ impl<P: PacketMut> DownstreamPacket<P> {
         let timer = metrics::processing_time(metrics::READ).start_timer();
         if let Err(error) = self.process_inner(config, sessions, destinations) {
             let discriminant = error.discriminant();
-            metrics::errors_total(metrics::READ, discriminant, &metrics::EMPTY).inc();
+
+            // We only want to mark potential I/O errors as errors, as they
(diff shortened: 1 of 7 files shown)


src/components/proxy/packet_router.rs
```diff
@@ -87,17 +85,10 @@ impl<P: PacketMut> DownstreamPacket<P> {
...
-                error_acc.maybe_send();
-            }
-            Err(error) => {
-                let discriminant = error.discriminant();
-                metrics::errors_total(metrics::READ, discriminant, &metrics::EMPTY).inc();
-                metrics::packets_dropped_total(metrics::READ, discriminant, &metrics::EMPTY).inc();
-
-                error_acc.push_error(error);
-            }
+        if let Err(error) = self.process_inner(config, sessions, destinations) {
```
Jake-Shadle, line 88: I'm not sure I understand, the PR title doesn't match, this seems like errors are still being logged?

XAMPPRocky: Sorry to clarify, they're being "logged" through prometheus metrics yes, they're not being sent to stderr though. This PR is about cleaning up our tracing logging.

Jake-Shadle (approved):

PR #1065, opened 2025-01-17
refactor: Move network servers under publish flags

XAMPPRocky:
This is the first big step in my CLI refactoring, the goal is to move away from the distinct modes (agent, proxy, relay, etc) towards allowing users to specify individual components they want to enable (for example, to have the same functionality as the old `proxy` you would write `--publish.udp --publish.qcmp --publish.phoenix`). The intention here is to simplify both our code and the concepts around them and to make deployments more composable. This will reduce a lot of test surface because it will make writing tests for specific publishers much easier, and allow us to remove redundant code paths where we had to handle components that were shared between two modes but were unused by one mode.

Currently the modes are still included, because I wanted to divide the PR up, the intention is make a followup that applies the same logic to configuration providers and that will remove the modes entirely.


Jake-Shadle (approved):
Approving, though the `publish` term is not fully making sense to me.

XAMPPRocky:
> though the publish term is not fully making sense to me.

I thought about it in the sense of it being services that quilkin "exposes" or "publishes" to the user. Also in the sense of "config provider" being an input to quilkin, the "publisher" is output of quilkin. I was thinking about making config providers under a `subscribe` flag so it might feel more intuitive with those added. I am open to better suggestions though this was just me looking at a bunch of words and trying to find the one that matches the intent.

Jake-Shadle:
No, that makes sense, I guess I always just first associate publish with like, releases/artifacts etc.

PR #1066, opened 2025-01-20
First pass implementation of XDP I/O loop

Jake-Shadle:
This PR is the real meat of the XDP implementation, the I/O loop.

The setup of the memory, sockets, and rings is done in the root xdp module, along with the spawning of one (named) thread per NIC queue, the actual packet processing code is in the xdp/process module. This split of code allows much easier testing, since we can just have a Umem (memory map) and don't need to do any actual I/O.

For the sake of simplicity in this initial pass I decided not to share any code at all with the current sessionpool, instead opting for a more minimal approach to pairing client <-> server sessions by having a single TTL map for all servers, where each unique client gets its own port in the range of valid ephemeral ports that Linux itself doesn't assign, which gives 4535 unique clients per server, so the TTL applies to the server instead of each individual client. Other than that difference, the only thing different about XDP versus the other implementations is that packets need to parsed manually when received to pull out the source information, and, unless being routed to multiple servers, the packet is then modified inline and queued for sending with the new source and destination information as determined by the filters.


crates/xdp/src/lib.rs
```diff
@@ -40,8 +43,18 @@ pub enum LoadError {
     DefaultPortRangeModified(u16, u16),
 }
 
+pub struct XdpWorker {
```
XAMPPRocky, line 46: Can we get some documentation on this struct and its fields? Focused on their role rather than what they are.

src/components/proxy/xdp/process.rs
```diff
@@ -0,0 +1,606 @@
...
+};
+use std::{
+    net::{IpAddr, SocketAddr},
+    sync::{
+        atomic::{AtomicU16, Ordering},
+        Arc,
+    },
+};
+
+struct QPacket {
```
XAMPPRocky, line 24: Can this be renamed to be more clear? Like could it be called what layer of packet it is, e.g. IPPacket, EthernetPacket, etc. Can there be docs on what it's purpose are here?

XAMPPRocky (commented):
LGTM! I most have some docs questions, and overall questions, the code itself when laid out like this is relatively straightforward.

The other more general question I have is, is this possible to test in a deployment scenario yet? I wasn't 100%, because it doesn't seem like it runs in the proxy entrypoint yet, but I want to doublecheck.

> which gives 4535 unique clients per server

Is this per gameserver or per proxy? Also in the future (e.g. once the feature is stable) do you think it's possible to move to a similar (or same impl) as we have for user space where we share ports between clients talking to different servers?

> so the TTL applies to the server instead of each individual client.

Is there a reason for this method instead of the other. I think it's fine for now, but we'll probably want to think about the pros and cons of each approach and unify on a solution for both I/O options.

src/components/proxy.rs
```diff
@@ -18,6 +18,9 @@ pub(crate) mod error;
 pub mod packet_router;
 mod sessions;
 
+#[cfg(target_os = "linux")]
+pub mod xdp;
```
XAMPPRocky, line 22: Also minor nit: Can we move this under `net` as opposed to in the proxy. I think it belongs there.

XAMPPRocky (approved):

Jake-Shadle:
> The other more general question I have is, is this possible to test in a deployment scenario yet? I wasn't 100%, because it doesn't seem like it runs in the proxy entrypoint yet, but I want to doublecheck.

No, the next PR is the one that will actually tie everything together be able to use XDP or io-uring at startup.

> Is this per gameserver or per proxy?

Per gameserver.

> Also in the future (e.g. once the feature is stable) do you think it's possible to move to a similar (or same impl) as we have for user space where we share ports between clients talking to different servers?

It might be possible, it would just be a larger refactor that doesn't make sense for a first pass implementation IMO.

> Is there a reason for this method instead of the other. I think it's fine for now, but we'll probably want to think about the pros and cons of each approach and unify on a solution for both I/O options.

This was just the simplest implementation for now, I though about the limitations of this approach versus the current one, it was just easier to do this one now, I think I've done it in a way that it's trivial to change to something different later without changing too much.

PR #1069, opened 2025-01-20
Add XDP CLI options and initialization code

Jake-Shadle:
This adds the support for several XDP specific options, and the initialization code to bring up the XDP I/O loop, falling back to io-uring if setup fails for some reason. This PR _always_ falls back to io-uring at the moment, this is just to get the code in so that it can be iterated on in further PRs, as I only tested that we can spinup quilkin proxy using XDP and shut it down cleanly, not the actual I/O which I'm leaving for tomorrow.

Fixed several bugs:

- Fix bug in ephemeral port range parsing
- Remove warning about eBPF logging, the current eBPF program has no log messages, which makes log initialization fail, but we would want to enable it in the future if we did add logging to the program
- `pthread_cancel` is apparently super not good and was causing a crash at shutdown due to signal shenanigans, so I just replaced it with an atomic bool for now
- We include the eBPF program inlined in the quilkin binary, but it could be aligned however rustc/LLVM decided, and object parsing requires it be aligned correctly, which for 64-bit ELF files is 8 bytes

### DIFF
--- a/src/components/proxy.rs
+++ b/src/components/proxy.rs
@@ -72,6 +72,7 @@ pub struct Proxy {
     pub qcmp: socket2::Socket,
     pub phoenix: crate::net::TcpListener,
     pub notifier: Option<tokio::sync::mpsc::UnboundedSender<String>>,
+    pub xdp: crate::cli::proxy::XdpOptions,
 }
 
 impl Default for Proxy {
@@ -89,6 +90,7 @@ impl Default for Proxy {
             qcmp,
             phoenix,
             notifier: None,
+            xdp: Default::default(),
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/components/proxy.rs
```diff
@@ -295,7 +297,66 @@ impl Proxy {
     pub async fn spawn_packet_router(
         &mut self,
         config: Arc<crate::config::Config>,
-    ) -> eyre::Result<impl FnOnce(crate::ShutdownRx)> {
+    ) -> eyre::Result<Box<dyn FnOnce(crate::ShutdownRx) + Send>> {
+        #[cfg(target_os = "linux")]
+        'xdp: {
```
XAMPPRocky, line 302: Could this instead be its own function?

XAMPPRocky (approved):

PR #1078, opened 2025-01-24
Add XDP integration test

Jake-Shadle:
This adds an XDP integration test that spins up a proxy using XDP in a separate network namespace, and a client + echo server that runs in another namespace. I attempted to get a k8s cluster working but fail miserably because k8s is too complicated, the network namespace solution is much easier.


.ci/xdp/integration-test.sh
```diff
@@ -0,0 +1,82 @@
...
+metadata:
+  name: proxy
+spec:
+  containers:
+  - name: proxy
+    image: "$proxy_image"
+    ports:
+    - containerPort: 7777
+      hostPort: 7777
+    args: ["proxy", "--to", "${server_ip}:8078", "--publish.udp.xdp"]
```
XAMPPRocky, line 37:
```suggestion
    args: ["proxy", "--to", "${server_ip}:8078", "--service.udp.xdp"]
```

XAMPPRocky (approved):

Issue #1076, opened 2025-01-22
Move QCMP to eBPF or XDP


Jake-Shadle:
The XDP code is really new and will take time to fully test, but I wanted to bookmark this now so I don't forget.

Basically, the current XDP implementation is binding to all network device queues where each socket + queue gets their own separate thread. This is good in the sense that the load will be spread across all threads fairly evenly due to receive side scaling, but it has a downside, in that, though we poll, giving every thread a wait point where it can schedule other CPU work, in a regular case with hundreds or thousands of connections, these threads will all be active constantly, potentially starving other threads for long (seconds) periods of time. This is probably ok for eg. the HTTP admin service, but would be bad for the QCMP task as it would skew measurements unpredictably. The QCMP code is fairly simple, so it could just be moved to run entirely in the eBPF program we use to route packets, but that would make the fairly simple program more complicated, so we could just move it to be processed in the same I/O loop as the regular client/server packets since it's just handling packets sent to a specific port.

XAMPPRocky:
Yeah I think this makes sense. I think it might make sense that the QCMP module is written as an entirely separate eBPF module. It would be good if our code infra supported multiple eBPF modules so we can have different types of filters.

PR #1084, opened 2025-01-29
Add QCMP support in XDP I/O loop

Jake-Shadle:
Resolves: #1076


XAMPPRocky:
Did you see my comment about having this be a separate ebpf module?

https://github.com/googleforgames/quilkin/issues/1076#issuecomment-2610848553

Jake-Shadle:
We can only have 1 eBPF module loaded on a device at a time, libbpf gets around that by doing tail calls, but we'd have to do that ourselves, I guess I'm not sure why we would want that added complexity? Keeping the code in userspace means the eBPF code gets to stay extremely simple, the more complex it gets the harder it is to maintain since then you need to ensure it keeps passing the validator at all times and makes it harder to share code that already exists in quilkin, like for the QCMP stuff.

XAMPPRocky:
> I guess I'm not sure why we would want that added complexity? Keeping the code in userspace means the eBPF code gets to stay extremely simple, the more complex it gets the harder it is to maintain since then you need to ensure it keeps passing the validator at all times and makes it harder to share code that already exists in quilkin, like for the QCMP stuff.

My thought with QCMP, is that since it is such a simple protocol, it doesn't need to exist in the user space at all. Also this code when combined into a single program needs to handle the other UDP not being set by the user space program (e.g. the agent mode exposes QCMP but not UDP). And it might be simpler if that logic was in the user space program where it only loads the modules that are requested, and then the module just assumes it is needed.

Jake-Shadle:
So you're saying you want the agent to run QCMP in eBPF and share that code with the proxy via an eBPF module?

XAMPPRocky:
> So you're saying you want the agent to run QCMP in eBPF and share that code with the proxy via an eBPF module?

No I'm just clarifying that QCMP and UDP should be considered separate services, and thus that you can either one or both services active at a time. So either we need to put that in one module where we pass that as a setting to the module, or we have them considered to be separate modules and we decide at load time, whether to have UDP, QCMP, or UDP+QCMP.

Jake-Shadle:
Sure, I understand they are separate logically, but in practice with XDP it makes more sense to have them in the same I/O loop because of the thread starvation issue and timing sensitivity of the QCMP packets. Putting that code in eBPF instead of userspace means code duplication, and just makes it harder since the only source of time available to eBPF is nanoseconds since system boot, not unix epoch. I've made it so that if QCMP is disabled it just passes 0 as the port, and it would be trivial to create a different eBPF program for the agent that could only handle QCMP in the future if we felt that made sense.

XAMPPRocky:
Okay, the lack of epoch time is a non-starter. So if we have just one XDP module, would you be able to add it so it also doesn't run the UDP service if its port is zero?

Then we can have XDP loaded if either UDP or QCMP are enabled in `Service::spawn_services`.

How does that sound?

XAMPPRocky (approved):

Issue #1094, opened 2025-02-04
Timeout from QCMP ping when using XDP.


XAMPPRocky:
Hey, I'm trying to ping quilkin with the XDP QCMP functionality. Both in the cluster and locally, I see the following log, but I always get timeouts. When disabling XDP QCMP pings work as expected.

```
2025-02-04T07:24:28.492207Z  INFO ThreadId(31) quilkin::net::xdp: src/net/xdp.rs: selected NIC nic=2 "eth0"
```

Commands:

```
cargo r -- --service.udp --service.udp.xdp --service.qcmp
```

```
cargo r -- qcmp ping 127.0.0.1:7600
```

PR #1095, opened 2025-02-04
Fixup QCMP ping

Jake-Shadle:
This fixes a few CLI issues with the UDP + QCMP setup, and adds an integration test so IMO this resolves the issue.

**HOWEVER**, it took me a while to figure this out because right now, all service setup errors are being completely swallowed with no feedback or error messages.

https://github.com/googleforgames/quilkin/blob/089fe1deb222ac36843319210bc0275bae8e1f06/src/cli/service.rs#L233-L239

In this code, if `--service.udp.xdp` was specified, and eg. it failed to get a default network device, the error would bubble up, causing the io-uring UDP and QCMP task to not spawn, but there would be no error messages since every call to `spawn_services` is throwing away the actual task handle that is being returned.

Resolves: #1094


XAMPPRocky:
> HOWEVER, it took me a while to figure this out because right now, all service setup errors are being completely swallowed with no feedback or error messages.

We should change that.

PR #1105, opened 2025-02-06
Add session metrics to XDP

Jake-Shadle:
These metrics will be different than the ones in the session pool due to tracking sessions more coarsely, but should be good enough for testing purposes.


XAMPPRocky:
Went through some other metrics and found some others missing.

Can we also add `quilkin_packets_total`, `quilkin_bytes_total`, and `quilkin_packets_processing_duration`? Either as a followup or in this?

XAMPPRocky (approved):

Issue #1107, opened 2025-02-10
Block io-uring shutdown on full session termination


XAMPPRocky:
Previously with epoll where the i/o lifetime was tied to the sockets in the sessions, now currently with io-uring, it shutdowns upon receiving the termination immediately. This means that we can't continue those sessions and they immediately shutdown anyway. This should be changed so that io-uring shutdown is triggered after all sessions have dropped.

PR #1108, opened 2025-02-10
Defer shutdown until actives sessions are 0

Jake-Shadle:
This defers shutdown by still processing I/O requests in the existing downstream and session handlers, but blocks downstream loops from creating new sessions, so sessions will stay alive as long as traffic continues, but once they don't have traffic any longer they will eventually be pruned from the pool, and shutdown will continue, also shutting down the phoenix thread if it exists before informing the proxy to exit.

Resolves: #1107

### DIFF
--- a/src/components/proxy/packet_router.rs
+++ b/src/components/proxy/packet_router.rs
@@ -73,7 +73,6 @@ impl<P: PacketMut> DownstreamPacket<P> {
         config: &Arc<Config>,
         sessions: &S,
         destinations: &mut Vec<crate::net::EndpointAddress>,
-        deny_new_sessions: bool,
     ) {
         tracing::trace!(
             id = worker_id,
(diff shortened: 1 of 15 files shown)


src/components/proxy/packet_router.rs
```diff
@@ -73,6 +73,7 @@ impl<P: PacketMut> DownstreamPacket<P> {
         config: &Arc<Config>,
         sessions: &S,
         destinations: &mut Vec<crate::net::EndpointAddress>,
+        deny_new_sessions: bool,
```
XAMPPRocky, line 76: Hmm, I'm not sure if we should disallow new sessions as that could cause disruption for players who have an IP:port change in the middle of a session while a proxy is being shut down. I think a better alternative would be to have a "termination timeout" that can specified in the CLI, where if enabled it would shutdown now matter what after a specified number of seconds. If not specified it should have no timeout, because for example in k8s we already specify a graceful termination time of 40 minutes to give time for games to finish before it has to shutdown.

Jake-Shadle: Ok.

XAMPPRocky (approved):

PR #1119, opened 2025-02-17
Update xdp and print out device caps after query

Jake-Shadle:


src/net/xdp.rs
```diff
@@ -164,6 +164,8 @@ pub fn setup_xdp_io(config: XdpConfig<'_>) -> Result<XdpWorkers, XdpSetupError>
         .query_capabilities()
         .map_err(|err| XdpSetupError::NicQuery(name, err))?;
 
+    tracing::debug!(?device_caps, nic = ?nic_index, "XDP features for device");
```
XAMPPRocky, line 167: Could we print this out as info level on the `ZeroCopyUnavailable` and `TxChecksumUnavailable` error paths?

Issue #1117, opened 2025-02-13
Refactor `Config` to be dynamic


XAMPPRocky:
As the scope of different services has grown, we've continually added to the config struct, this has added a lot of little branches for handling the same configuration in different modes. We should instead move from a hardcoded set of fields, to a more dynamic type map or similar structure where each config provider is able to provide the same or different data, and services can subscribe to updates to data that is relevant for them. This would reduce the amount of code needed to add new integrations as adding new types, providers, and services can all be encapsulated.

Jake-Shadle:
Maybe I am missing something, but I don't really understand why this change is needed. AFAIK, the Config struct actually hasn't changed that much within the past year, and there is really only one wrinkle with it as it is now, the problem that agent is a special case different from every other config. To me it would be much easier to split out agent from the other components and then special case the paths where agent is used, since otherwise most of the code is shared between the components. It seems that moving to a dynamic config would just make all of the code that deals with Config even worse as everything would be checked at runtime, losing basic compile time checking.

PR #1123, opened 2025-02-19
Begin config migration

Jake-Shadle:
- Adds `DynamicConfig` as a field of `Config`
- Moves `id` and `version` to `DynamicConfig`
- Change phoenix and k8s to only take the fields they actually need instead of an entire config
- Change admin to only take in a `<C: Serialize>` as it only cares about dumping the config as JSON

Part of #1117

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -79,8 +79,6 @@ pub struct Config {
     pub filters: Slot<crate::filters::FilterChain>,
     #[serde(default = "default_proxy_id")]
     pub id: Slot<String>,
-    #[serde(default)]
-    pub version: Slot<Version>,
     #[serde(flatten)]
     pub datacenter: DatacenterConfig,
 }
@@ -562,7 +560,6 @@ impl Config {
             clusters: Default::default(),
             filters: Default::default(),
             id: default_proxy_id(),
-            version: Slot::with_default(),
             datacenter: DatacenterConfig::Agent {
                 icao_code: Default::default(),
                 qcmp_port: Default::default(),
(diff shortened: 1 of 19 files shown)


src/config.rs
```diff
@@ -552,17 +728,21 @@ impl Config {
...
         Self {
             clusters: Default::default(),
             filters: Default::default(),
-            id: default_proxy_id(),
-            version: Slot::with_default(),
+            dyn_cfg: DynamicConfig {
+                id: default_id(),
+                version: Version::default(),
+                typemap: default_typemap(),
+            },
```
XAMPPRocky, line 745: Is there a reason DynamicConfig doesn't impl default?

src/config.rs
```diff
@@ -67,31 +71,248 @@ pub enum DatacenterConfig {
...
-    pub id: Slot<String>,
-    #[serde(default)]
-    pub version: Slot<Version>,
-    #[serde(flatten)]
     pub datacenter: DatacenterConfig,
+    pub dyn_cfg: DynamicConfig,
+}
+
+#[cfg(test)]
+impl<'de> Deserialize<'de> for Config {
```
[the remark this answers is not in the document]
Jake-Shadle: Because the dynamic config will require manual deserialization so just getting it out of the way now.

src/config.rs
Jake-Shadle: Not really, will probably add it eventually.

XAMPPRocky (approved):

PR #1125, opened 2025-02-20
Make Config::clusters dynamic

Jake-Shadle:


src/config.rs
```diff
@@ -392,33 +413,59 @@ impl quilkin_xds::config::Configuration for Config {
...
+            let dc = match &control_plane.config.datacenter {
+                crate::config::DatacenterConfig::Agent { .. } => None,
                 crate::config::DatacenterConfig::NonAgent { datacenters } => {
-                    let mut dc_watcher = datacenters.watch();
+                    Some(datacenters.clone())
+                }
+            };
+
+            match (clusters, dc) {
+                (Some(clusters), Some(dc)) => {
```
XAMPPRocky, line 426: As a followup, I think it would be good if this implemented a little differently, where we iterate over the typemap, and get the watcher for each type, and when any of those watchers changes it pushes an update with it's matching xds type.

The typemap library we're using doesn't have a `.iter()`, but the type is just a wrapper around `HashMap`, so it should be straightforward to add that.

XAMPPRocky (approved):

PR #1152, opened 2025-03-07
Cleanup filter tests

Jake-Shadle:
This changes the filter integration tests to use the XDP interface so that we can get rid of the flaky I/O with real sockets in favor of just using memory. I mostly didn't change the tests themselves outside of that, but using this method means we can be more creative with test cases since we can use any IP address we want for clients/servers etc since we aren't using real sockets, eg. the firewall tests only use local addresses.

This also adds a super simple `filter_chain!` macro to easily create filter chains in tests, there are already tests for the actual deserialization of filter configurations, and IMO it's cleaner to have the tests that are testing the functionality not have to deal with the particulars of deserialization, eg. many tests were using tokens, but those have to be base64 encoded in the yaml configs, so there would be a comment about that, but it's a lot simpler to just have `b"token"` in the test code.


XAMPPRocky (approved):
Nice work!

Issue #1166, opened 2025-03-13
h2 protocol error: error reading a body from connection


XAMPPRocky:
```
{"timestamp":"2025-03-12T11:09:34.635440Z","level":"ERROR","fields":{"message":"error receiving delta response","error":"status: Unknown, message: \"h2 protocol error: error reading a body from connection\", details: [], metadata: MetadataMap { headers: {} }"},"target":"quilkin_xds::server","filename":"crates/xds/src/server.rs","span":{"client":"agent","id":"agent","name":"xds_server_stream"},"spans":[{"client":"agent","id":"agent","name":"xds_server_stream"}],"threadId":"ThreadId(2)"}
```

XAMPPRocky:
This seems to be related to unexpected EOF on TLS, so we might just need to handle those errors properly.

https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof

PR #1174, opened 2025-03-19
Move set_registry call

Jake-Shadle:


XAMPPRocky (approved):

XAMPPRocky:
Are we sure that the flakiness in this test is not #1166 ?

PR #1175, opened 2025-03-20
Reject cluster changes from different ips

Jake-Shadle:


XAMPPRocky:
I think there's some logic missing. Mainly, when a client attempts to apply to cluster information and fails, a NACK should be returned by the client to the server. That way we can block rollout of a version based on whether nacks are present.

XAMPPRocky (approved):

Issue #1180, opened 2025-03-25
Guide no longer building versions from `main`


XAMPPRocky:
Looking at the `gh-pages` branch, the last push was a month ago, there haven't been any docs pushes since.

https://github.com/googleforgames/quilkin/tree/gh-pages

PR #1181, opened 2025-03-25
Fix docs deploy

Jake-Shadle:
Resolves: #1180

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -9,6 +9,8 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    env:
+      docs_dir: docs-build
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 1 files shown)


Jake-Shadle:
Actually this isn't fixable without the ability to add a deploy key, we only have a github token, which doesn't have permissions to push, so we would need to have the repo transferred first to be able to fix this.

Jake-Shadle:
> remote: Permission to googleforgames/quilkin.git denied to github-actions[bot].
> fatal: unable to access 'https://github.com/googleforgames/quilkin.git/': The requested URL returned error: 403

In addition, I looked at one of my repos that does something similar, and it uses a deploy key to do this rather than the default token because (at least at the time) it wouldn't actually trigger a page deploy unless using a deploy token, though that might have been fixed over the years.

XAMPPRocky:
Yeah that used to be the case, but now you can configure it. But thanks for confirming it's not working here 🙏

PR #1192, opened 2025-03-31
refactor(cli): remove old cli components

XAMPPRocky:
Another attempt at #1137


Jake-Shadle (approved):

crates/agones/src/relay.rs
```diff
@@ -62,9 +62,9 @@ mod tests {
         quilkin::test::enable_log("agones=debug");
 
         run_test(true, true, true, 0).await;
-        run_test(true, true, false, 1).await;
-        run_test(true, false, true, 2).await;
-        run_test(false, true, true, 3).await;
+        //run_test(true, true, false, 1).await;
+        //run_test(true, false, true, 2).await;
+        //run_test(false, true, true, 3).await;
```
Jake-Shadle, line 67: This needs to be disabled until we've changed the previous image to a version that supports the new arguments.

Issue #320, opened 2021-07-07
Quilkin should drop packets that are sent from itself.


XAMPPRocky:
Quilkin currently doesn't do any checking of where the source is from, and the source ip and port can be easily forged, and common version of that forgery is to have the source of the packets also be the destination as the response can be larger than the request (see UDP amplification attacks), so the solution is to ban all traffic that is sourced from itself.

PR #1197, opened 2025-04-14
Drop packets from disallowed source IPs

glindstedt:
**What type of PR is this?**

/kind feature

Partly addresses https://github.com/googleforgames/quilkin/issues/320, but does not cover dropping packets with the source IP set to quilkin's external IP address.

Related to #320

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -17,6 +17,7 @@
 //! Quilkin configuration.
 
 use std::{
+    collections::BTreeSet,
     net::{IpAddr, SocketAddr},
     sync::{
         Arc,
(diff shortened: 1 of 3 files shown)


src/config.rs
```diff
@@ -881,6 +882,22 @@ impl Config {
         Ok(())
     }
 
+    pub fn cluster(
```
glindstedt, line 885: A bit unsure about method name (e.g. could be `with_cluster` or `add_cluster`), and if we should "pollute" `Config::` with builder methods or if it's worth having a `ConfigBuilder` with a `build() -> Config`?

XAMPPRocky: I think the method name is fine and having the builder methods direct is fine. Having `ConfigBuilder` would make sense if we need to do some initialisation logic (e.g. if the `build` method can fail because it sets stuff up), but I don't think we need that for `Config` since it is just a data store.

XAMPPRocky (approved):

Jake-Shadle (approved):

PR #1203, opened 2025-04-25
Fix ClusterMap deadlock

Jake-Shadle:
The `ClusterMap::remove_endpoint` and `ClusterMap::remove_endpoint_if` would 100% cause a deadlock if the last gameserver was removed from the locality as a reference was held via the clustermap iterator, while map.remove() was called. This just moves the locality removal outside of the iteration to avoid it.

### DIFF
--- a/crates/test/tests/xdp.rs
+++ b/crates/test/tests/xdp.rs
@@ -434,7 +434,7 @@ async fn multiple_servers() {
         endpoints(
             servers
                 .iter()
-                .map(|a| (a.clone(), &tok[..]))
+                .map(|a| (*a, &tok[..]))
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
> would 100% cause a deadlock if the last gameserver was removed from the locality as a reference was held via the clustermap iterator.

The code makes sense to me, but would you be able to add a test that catches this behaviour? Because it seems like something that would be easy run into again if we refactor the code.

Jake-Shadle:
Test added, confirmed the test deadlocks on main.

XAMPPRocky (approved):

Issue #1187, opened 2025-03-28
Consolidate xDS and mDS into single `gRPC` service that serves both


XAMPPRocky:
What it says on the tin, instead of having them separate, we should merge the two services under one server to make it easier to understand, configure, and to better match our protocol unit of services.

PR #1205, opened 2025-05-08
Combine xDS and mDS services

Jake-Shadle:
Closes: #1187

### DIFF
--- a/crates/xds/src/client.rs
+++ b/crates/xds/src/client.rs
@@ -338,9 +338,39 @@ pub(crate) struct DeltaClientStream {
 impl DeltaClientStream {
     #[inline]
     async fn connect(
-        mut client: AdsGrpcClient,
+        endpoints: &[Endpoint],
         identifier: String,
-    ) -> Result<(Self, tonic::Streaming<DeltaDiscoveryResponse>)> {
+    ) -> Result<(Self, tonic::Streaming<DeltaDiscoveryResponse>, Endpoint)> {
+        if let Ok((mut client, ep)) = MdsClient::connect_with_backoff(endpoints).await {
+            let (req_tx, requests_rx) =
+                tokio::sync::mpsc::channel(100 /*ResourceType::VARIANTS.len()*/);
+
+            // Since we are doing exploratory requests to see if the remote endpoint supports delta streams, we unfortunately
+            // need to actually send something before the full roundtrip occurs. This can be removed once delta discovery
+            // is fully rolled out
+            req_tx
+                .send(DeltaDiscoveryRequest {
+                    node: Some(Node {
+                        id: identifier.clone(),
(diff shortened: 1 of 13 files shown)


crates/xds/src/client.rs
```diff
@@ -338,9 +338,39 @@ pub(crate) struct DeltaClientStream {
...
         identifier: String,
-    ) -> Result<(Self, tonic::Streaming<DeltaDiscoveryResponse>)> {
+    ) -> Result<(Self, tonic::Streaming<DeltaDiscoveryResponse>, Endpoint)> {
+        if let Ok((mut client, ep)) = MdsClient::connect_with_backoff(endpoints).await {
+            let (req_tx, requests_rx) =
+                tokio::sync::mpsc::channel(100 /*ResourceType::VARIANTS.len()*/);
+
+            // Since we are doing exploratory requests to see if the remote endpoint supports delta streams, we unfortunately
+            // need to actually send something before the full roundtrip occurs. This can be removed once delta discovery
+            // is fully rolled out
```
XAMPPRocky, line 350: I think we've rolled delta xDS long enough that is no longer needed. The client can assume delta.

src/cli/service.rs
```diff
@@ -247,7 +235,6 @@ impl Service {
         shutdown_rx: &crate::signal::ShutdownRx,
     ) -> crate::Result<tokio::task::JoinHandle<crate::Result<()>>> {
         let mut shutdown_rx = shutdown_rx.clone();
-        let mds_task = self.publish_mds(config)?;
```
XAMPPRocky, line 250: We still need to keep the old path in until this change has been rolled out.

I think it would make sense to spawn a separate task that does both when you provide `--service.grpc`, then once we've rolled out and switched the infra to use that task, we can remove all code in the `--service.xDS` and `--service.mds` paths.

XAMPPRocky (changes requested):
Overall the approach looks good to me, the one feedback I have is around preserving the API for at least one version/commit so that this can be seamlessly deployed.

XAMPPRocky:
I have a question that I'm not sure I can tell from the code in PR. If I switch `--service.xds --service.mds` to `--service.xds --service.grpc`, does that still run two http servers the ports for xds and mds? The answer should be yes in order to the deployment to work without disruption, as an operator needs to:

- Deploy new version, no argument changes.
- Update arguments for the relay to  `--service.xds --service.grpc` so that now both new and existing traffic can be served
- Update arguments for all connected deployments to switchover to the new grpc endpoint.
- Remove  `--service.xds` to close the service.

Jake-Shadle:
Pausing this change the filter_update test consistently fails and I haven't found a way to fix it since I don't know what is wrong.

Jake-Shadle:
The bug was fixed in #1215

Issue #1131, opened 2025-02-21
Build config at startup


XAMPPRocky:
That's how all of the different fields work right now, the goal would be to remove the Config::default* stuff (or gate it for tests only) and build the config at startup based on the requested features, but I didn't want to do that in this change.

_Originally posted by @Jake-Shadle in https://github.com/googleforgames/quilkin/pull/1128#discussion_r1965236458_

PR #1215, opened 2025-05-15
Remove Slot

Jake-Shadle:
This change removes `Slot<>` in favor of using concrete types for the individual pieces that used to be inside `Slot`.

- `Slot<FilterChain>` is now `config::filter::FilterChainConfig`, which is still an ArcSwap, but now takes advantage of the new type to use https://docs.rs/arc-swap/latest/arc_swap/cache/struct.Cache.html inside the UDP packet processing (io_uring and xdp) for better performance, as before we were using load_full even when the FilterChain is going to be the same 99.99% of the time.
- `Slot<Agent>` is now `QcmpPort` which is just an atomic u16, as the ICAO code is part of the base config that we can combine on the agent to publish `Datacenter` changes.
- `id: Slot<String>` is now just a Mutex<String>, it was never monitored for changes

Since I was modifying Config so much and changing how changes are monitored I went ahead and fixed 2 issues in this change that were related.

- Added `Service::build_config` which builds a config based on the enabled components, rather than using all components.

Resolves: #1131

The QCMP service now properly monitors the QCMP port for changes, attempting to bind a socket to the new port and listening to QCMP pings on that, if an error occurs it will emit the error and continue listening on the old port.


XAMPPRocky (commented):
So I've looked through the code and it in general looks good, but I have a question about the approach and where this goes.

Is the intent to eventually also replace the datatypes introduced here with whatever replaces `Watch`? Right now looking at this I don't know what I would do if I wanted to add a new field or struct to config. The current approach seems to be hardcoding slot rather than removing slot.

Jake-Shadle:
Eventually I would want to move towards a path where changes are serialized into a stream to be sent to remotes as notifications now are almost exclusively used for xds, to then retrieve the state to serialize it, so eventually it will hopefully be simpler again.

XAMPPRocky (approved):

PR #1230, opened 2025-05-27
Add CRDT collections

Jake-Shadle:
This is just the CRDT collections and types from #1225


src/config/crdt/types/datacenter_map.rs
```diff
@@ -0,0 +1,437 @@
...
+
+impl crate::config::DynamicConfig {
+    #[inline]
+    pub fn xds_datacenters(&self) -> Option<&Arc<XdsDatacenterMap>> {
+        self.typemap.get::<XdsDatacenterMap>()
+    }
+}
+
+#[derive(Debug)]
+pub struct XdsDatacenterMap {
```
XAMPPRocky, line 27: nit: Can we remove `Xds` from the name? the information isn't xds specific, it could come from other providers (e.g. config file from fs).

XAMPPRocky (approved):

PR #1231, opened 2025-05-28
feat(unreal): update unreal plugin.

XAMPPRocky:
This commit includes several updates to the unreal plugin.

- Proxy failover support
- Datacenter latency measurement.
- GetLowestLatencyEndpoint can now return shortest path to datacenter.
- Minimum version bumped to 5.3.


Jake-Shadle:
I've never looked at this code, never used unreal and haven't done c++ on many years, should someone else do the review?

PR #1259, opened 2025-09-02
Update crates

Jake-Shadle:
Updates crates, notably tonic from 0.12 -> 0.14 as tonic-build is no longer used and is now tonic-prost-build.

I've removed tonic/prost as a dependency of pprof, as that crate generally lags behind, and it's easier for us to just implement the prost encoding in quilkin as it doesn't change and means we don't get duplicates of tonic/prost when we update quilkin's dependencies.


glindstedt (approved):
LGTM

PR #1267, opened 2025-09-22
feat: Clean up Datacenters which fail to respond to qcmp pings

glindstedt:
A proxy that is connected to a relay may miss an update where a Datacenter is removed if there is a temporary disruption. This adds a channel where the phoenix subsystem can report "bad nodes" that it has failed to ping for a period, so that they can be cleaned up from the Config.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -493,7 +493,7 @@ impl Config {
         Ok(())
     }
 
-    pub fn get_bad_node_informer(&self) -> Option<BadNodeInformer> {
+    pub fn bad_node_informer(&self) -> Option<BadNodeInformer> {
         self.bad_node_informer.clone()
     }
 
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -409,6 +493,10 @@ impl Config {
         Ok(())
     }
 
+    pub fn get_bad_node_informer(&self) -> Option<BadNodeInformer> {
```
XAMPPRocky, line 496: FYI You don't need to call this `get_`, Rust has UFCS, so there's no way for `self.bad_node_informer` to be confused as a function, as `self.bad_node_informer()` is sugar for `Config::bad_node_informer(self)`.
```suggestion
    pub fn bad_node_informer(&self) -> Option<BadNodeInformer> {
```

XAMPPRocky (approved):

PR #1268, opened 2025-09-23
Build container image with github actions

glindstedt:
Just the bare minimum to build the same container image that we got from cloud build via a github action and push it to `ghcr.io/embarkstudios/quilkin`. I've bypassed `build/Makefile` entirely and only picked out the relevant steps, but leaving all the old stuff as is for now. I created a new `Dockerfile.ghaction` that runs some of the old make targets in a dockerfile multi stage build instead as that works better with `docker/build-push-action`.

The `build-env` image is not saved in a remote registry but should be cached between runs in the gha cache.

The action should now run on
* any push to main
* any new tag (without `/` in the name)
* triggering the workflow manually (can't test this until the workflow is on main)

One image pushed so far: https://github.com/EmbarkStudios/quilkin/pkgs/container/quilkin, same size as the last cloud  build image and still has the commit sha embedded in the binary.

### DIFF
--- a/image/Dockerfile.ghaction
+++ b/image/Dockerfile.ghaction
@@ -1,0 +1,17 @@
+FROM build-env AS builder
+ENV CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER=/usr/bin/x86_64-linux-gnu-gcc
+COPY . /workspace
+WORKDIR /workspace
+RUN cargo about generate license.html.hbs > license.html
+RUN cargo run -p proto-gen -- generate
+RUN cargo build --profile=lto --target x86_64-unknown-linux-gnu
+RUN ./image/archive_dependencies.sh
+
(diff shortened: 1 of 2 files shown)


image/Dockerfile.ghaction
```diff
@@ -0,0 +1,17 @@
+FROM build-env AS builder
+ENV CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER=/usr/bin/x86_64-linux-gnu-gcc
+COPY . /workspace
+WORKDIR /workspace
+RUN cargo about generate license.html.hbs > license.html
+RUN cargo run -p proto-gen -- generate
+RUN cargo build --profile=lto --target x86_64-unknown-linux-gnu
+RUN ./image/archive_dependencies.sh
```
glindstedt, line 8: These steps replicate the `build-license-report`, `gen-protobuf` and `build-linux-binary` make targets

PR #1271, opened 2025-09-29
Add persistent UDP stream

Jake-Shadle:
Follow up on #1266 which never got any reviews, so this PR is a bit bulky.

In addition to the things in #1266, this adds a quinn-based UDP server and client implementation that is meant for agents (clients) to use to connect to relays (servers) so that the relay can track the connection status of each agent. This essentially replaces the usage of corrosion's client implementation for mutations, as that would have needed a parallel channel so that the relays could determine the up status of each agent, so this just combines the connection with the mutations themselves.

The DB schema has changes, as the `servers` table has 2 new columns, `contributors` and `cont_time`. `contributors` is a JSONB object where each key is the IP address of an agent that has contributed to it. When the connection to an agent is lost, it is removed as a contributor from all the servers. We can then run a garbage collection task that can remove servers with no contributors, if they have not been updated in whatever amount of time we want to.

This adds some basic tests for the various DB operations and the quinn stream.


crates/corrosion/src/server.rs
glindstedt, line 1: TODO or remove?

crates/corrosion/src/persistent.rs
```diff
@@ -0,0 +1,301 @@
...
+            }
+            theirs => {
+                return Err(HandshakeError::UnsupportedVersion {
+                    ours: server_version,
+                    theirs,
+                });
+            }
+        };
+
+        Ok((version, this))
```
XAMPPRocky, line 105: ditto

crates/corrosion/src/persistent.rs
```diff
@@ -0,0 +1,301 @@
...
+        req[8..12].copy_from_slice(self.icao.as_bytes());
+        req
+    }
+
+    #[inline]
+    pub fn read(buf: [u8; 6]) -> Result<Self, HandshakeError> {
+        let qcmp_port = buf[0] as u16 | (buf[1] as u16) << 8;
+        let icao = buf[2..].try_into()?;
+        Ok(Self { qcmp_port, icao })
+    }
```
XAMPPRocky, line 76: ditto

crates/corrosion/src/server.rs
Jake-Shadle: TODO

Issue #1292, opened 2025-10-22
Subscription I/O


Jake-Shadle:
Once #1291 is completed, the next step is actually transporting the serialized subscription events. Corrosion only supports streaming HTTP/2 bodies, and the HTTP server implementation is tightly coupled with their agent, so we could either reimplement our own HTTP server implementation since we already have one, or possibly add support to corrosion to make it easier to work with as a library.

IMO a better option is to avoid HTTP/2 altogether, as #1271 already adds support for quinn streams, so making the quinn server implementation also support subscriptions via individual quinn streams means the I/O for transactions and subscriptions can be in one place and IMO easier to read and understand, but this is totally up for discussion.

PR #1308, opened 2025-11-17
Subscription I/O

Jake-Shadle:
This PR expands on #1296 and implements the actual I/O over a quic stream of subscribing and receiving events.

Ideally #1296 would be merged first so it's easier to review the changes unique to this PR over that one.

Resolves: #1292


Jake-Shadle:
I haven't yet tracked down the non-determinism bug in the test that was failing yet, so this still shouldn't be merged (or if it is, need to disable that test for now).

PR #1313, opened 2025-12-09
Update crates and rust version

Jake-Shadle:
Recently updated rust-analyzer and it now gives a warning about requiring rust 1.90.0+, so this just updates the rust toolchain, fixes all of the lints/formatting, and updated a bunch of crates while I was doing this.


rust-toolchain.toml
```diff
@@ -13,5 +13,5 @@
 # limitations under the License.
 
 [toolchain]
-channel = "1.85.0"
+channel = "1.90.0"
```
XAMPPRocky: Any reason why 1.90.0 and not 1.91.1?

PR #1379, opened 2026-04-14
feat: add http provider + provider benchmarks

XAMPPRocky:
This PR adds a HTTP API for adding endpoints to Quilkin, this allows other services to more easily integrate with Quilkin if they don't run on kubernetes, and makes it easier for us to add benchmarks for adding and removing endpoints.


Jake-Shadle (approved):

targrub:
To keep attack surface lower if desired, maybe this should be behind a crate feature?

XAMPPRocky:
@targrub that wouldn't make sense because our main distribution is a container image.

The feature is already behind a CLI flag that you have to explicitly enable.

PR #1393, opened 2026-05-07
Remove sub DBs when streams are terminated

Jake-Shadle:
We previously weren't explicitly cleaning up subscriptions when the subscriber connection was closed, and we now completely remove all subscription databases both on start and shutdown.


XAMPPRocky:
Can we add some tests that check what state the database is in after the operation?

Jake-Shadle:
I've added a couple of assertions that the sub path is cleaned up as expected.

XAMPPRocky (approved):

PR #1412, opened 2026-06-08
Replace forwarded-header-value with inline code

Jake-Shadle:
This replaces the forwarded-header-value crate with simpler inline code that only does what we need. This actually removed 2 dependencies since it transitively depended on another crate unused elsewhere.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1654,16 +1654,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "forwarded-header-value"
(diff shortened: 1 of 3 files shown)


XAMPPRocky:
What is the advantage to this approach? It seems like it makes the code more brittle for little benefit. The dependency is less 500 lines I don't think it's having an affect on build times.

Jake-Shadle:
Fair enough.

PR #1423, opened 2026-06-25
feat: QuicDcid strategy for capture filter (#1419)

### LINKED ISSUE #1419: Proxying QUIC

I would like to use quilkin to proxy [QUIC](https://quicwg.org) traffic.
It is designed to have Destination Connection ID unencrypted specifically for loadbalancing/proxying!
Combining it with assigning part of DCID as a server identifier and using TokenRouter should work, right?

But it looks like currently quilkin filters are a tad short to be able to parse it out:
- There are 2 types of packet headers in QUIC protocol that are discriminated by the first bit.
- Long headers are used during handshake and have DCID length at byte 5 and DCID itself starting from byte 6.
- Short headers have DCID at byte 1. Length is known after handshake completion and controlled by server, so pretty much a constant for the proxy config.

I think some small buffs to `Capture` filter (configurable offset for preffix strategy) and `Match` filter (configurable bitmask for metadata value) would already be enough to extract DCID of known length from connection. It could also help with similar protocols that have session identifiers already built into packets, but at some dynamic offset?

Alternatively a separate strategy for a `Capture` filter (or a similar but separate filter) that targets QUIC protocol specifically could be more performant solution.

unlightable:
**What type of PR is this?**
/kind feature

Allows capturing DCID form QUIC packets to use it in further routing.

Closes #1419


XAMPPRocky:
Thank you for your PR! This looks good to me, though I want to mention we are thinking of redesigning the filter feature to make it more approachable and require less setup. But I think it would be good to keep QUIC support in mind for any replacement.

XAMPPRocky (approved):

unlightable:
> This looks good to me, though I want to mention we are thinking of redesigning the filter feature to make it more approachable and require less setup

It doesn't seem too daunting in its current state, but looking forward for updates!
Meanwhile I've fixed tests and linter, though my run of `cargo clippy` didn't trip. Maybe different rust toolchain version ...

XAMPPRocky:
ci is failing :)

unlightable:
It seems like CI fails due to some vulnerability in dependency (https://github.com/EmbarkStudios/quilkin/actions/runs/28241230180/job/85566137959#step:4:82).
But it has nothing to do with this PR? Would it be wise to fix it in main branch, then rebase this PR and rerun CI again ? (:

XAMPPRocky:
> Would it be wise to fix it in main branch, then rebase this PR and rerun CI again ?

It is already fixed on main, it just needed a rebase.

XAMPPRocky:
Thank you for your PR, and congrats on your first contribution! 🎉

PR #1446, opened 2026-07-17
Requeue io-uring receives even if there is an error

### LINKED ISSUE #1445: io_uring: session socket permanently stops receiving after any recv error (e.g. ENOBUFS) - multishot recv never rearmed
**What happened**:

After a receive error on a session (upstream) socket, that socket permanently stops receiving for the remainder of the process lifetime. The send path keeps working, so the proxy forwards client -> server traffic normally while silently losing 100% of server -> client traffic for every session assigned to that pooled socket, with no drop/error metric on 0.10.1.

In `pop_recv` (`src/net/io/completion/io_uring.rs`), a CQE with a negative result returns early before the `IORING_CQE_F_MORE` requeue check:

```rust
let ret = cqe.result();

if ret < 0 {
    let error = std::io::Error::from_raw_os_error(-ret);
    tracing::error!(%error, "error receiving packet");
    return None;                                  // <- early return
}

let flags = cqe.flags();

// Requeue the recv if needed
if flags & flags::IORING_CQE_F_MORE == 0 {
    self.enqueue_recvmsg();                       // <- never reached on error
}
```

An errored multishot `RecvMsgMulti` completion does not carry `IORING_CQE_F_MORE` - the multishot is terminated and must be rearmed. The early return skips the rearm, so the socket never reads again. In our case the trigger was `ENOBUFS` (os error 105), i.e. momentary provided buffer ring exhaustion during a small burst of first replies from a game server.

The interaction with `SessionPool` makes the symptom deterministic and very confusing: the *first* flow to a destination deterministically picks the same pooled socket (`ports_to_sockets.iter().next()` on a stable map), and the failed session then *reserves* that socket for the destination for the 60s session TTL. So players see: **first join to a fresh game server always times out; an immediate retry always works (diverted to a healthy socket); a retry after >2 minutes fails again (dead socket repicked)**.

**What you expected to happen**:

A transient recv error (particularly ENOBUFS, which just means "no provided buffer available right now") should not permanently deafen the socket,  the multishot recv should be rearmed and the socket should continue receiving.

**How to reproduce it (as minimally and precisely as possible)**:

Deterministic from code inspection: deliver any negative result CQE for the multishot `RecvMsgMulti` on a session socket (easiest: exhaust the provided buffer ring with a reply burst so the kernel completes with `-ENOBUFS`), then send further packets from the upstream endpoint to that socket. They are never received; `pop_recv` logged the error once and the recv was never reenqueued.

In our environment it reproduces on the first UDP flow through a freshly started proxy toward an Agones GameServer (Capture + TokenRouter): client handshake packets forward upstream fine, the game server's replies arrive at the proxy pod but are never read, the client times out after ~5s. An immediate reconnect works; a reconnect after the 60s session TTL fails again on the same (dead) socket.

**Anything else we need to know?**:

- **This is very likely the actual root cause of [#1405](https://github.com/EmbarkStudios/quilkin/issues/1405)** ("Quilkin Degradation" - same `ENOBUFS os error 105` log line, followed by silent permanent packet loss). The mechanism explains that report's shape exactly: each ENOBUFS permanently deafens one socket, so a proxy degrades gradually and unreproducibly over hours as sockets die one by one, while `netstat` shows no kernel level receive errors (the loss is a userspace recv that's simply never rearmed).
- **#1444 reduces the trigger frequency but doesn't fix the failure mode.** It bumps the *listener* ring 512→2048, but (a) session sockets are created with a **64-buffer ring** (`SessionPool::spawn_session` → `IoUringLoop::new(64, ...)`) and are untouched - our deaf sockets were all session sockets, where a single server reply burst >64 packets between refills can trip ENOBUFS, and (b) whenever any recv error does occur, at any ring size, the socket is still permanently dead. The `refreshes_recv_ring` test exercises buffer recycling under sequential load; it never delivers an actual errored CQE, which is the path that kills the socket.
- Still present on `main` as of `0fef3d2` ("fix(io): count more errors in metrics") - that commit adds an `errors_total` increment on this path but keeps the early return, so the socket still goes deaf (now with a metric).
- Suggested fix: read `cqe.flags()` and perform the `F_MORE`/requeue check before (or regardless of) the `ret < 0` early return. We are running exactly that as a small reorder patch on top of 0.10.1. One consideration: on a *persistent* socket error, unconditional rearm turns a dead socket into an error log loop so distinguishing transient errors (ENOBUFS) from fatal ones may be worth it, but silently deafening the socket is strictly worse.
- Because the drop happens in the dead recv path, `quilkin_packets_dropped_total` and all session metrics stay flat; the only signal is a single ERROR log line at the moment the loop dies.

**Environment**:
- Quilkin version: 0.10.1 (`ghcr.io/embarkstudios/quilkin:0.10.1-719d705`, commit `719d705`)
- Execution environment (binary, container, etc): container on Kubernetes (DigitalOcean DOKS); proxy DaemonSet running `--service.udp` with `--provider.xds.endpoints` against a separate `--service.xds` deployment (`--provider.k8s` + `--provider.k8s.agones`)
- Operating system: Linux (DOKS managed nodes; debian:bookworm-slim container)
- Custom filters? (Yes/No - if so, what do they do?): No - stock `Capture` (8-byte prefix, remove) + `TokenRouter`
- Log(s):
  ```
  {"timestamp":"2026-07-16T07:52:42.880713Z","level":"DEBUG","fields":{"message":"Session created","source":"[::ffff:<client>]:59066","dest":"<node>:7168"},"target":"quilkin::net::sessions"}
  {"timestamp":"2026-07-16T07:52:48.780836Z","level":"ERROR","fields":{"message":"error receiving packet","error":"No buffer space available (os error 105)"},"target":"quilkin::net::io::completion::io_uring","filename":"src/net/io/completion/io_uring.rs"}
  ```
  (after which zero packets are ever read from that socket; a later session reusing it at 07:56:08 lost all replies the same way)
- Others: backend endpoints are Agones GameServers addressed via node address + hostPort; upstream forwarding was verified working throughout (the game server receives and answers - replies arrive at the proxy pod and are never read)

**Verification of the suggested fix**:

We deployed a build of 0.10.1 with only the reorder patch applied (session ring left at 64 buffers, i.e. the ENOBUFS trigger conditions unchanged) to the environment that reproduced this on each first join. Both previously always failing cases - the first flow to a fresh game server, and a reconnect after the 60s session TTL (which repicks the previously dead pooled socket) now succeeds on the first attempt, with the proxy's read/write counters advancing together from the first handshake packet. Can open a PR with the patch if helpful.

Jake-Shadle:
This also exposes the ring buffer lengths of both the core receive workers and each session pool.

Resolves: #1445

### DIFF
--- a/src/net/io/completion/io_uring.rs
+++ b/src/net/io/completion/io_uring.rs
@@ -51,6 +51,7 @@ pub fn spawn_listener(
     listener: crate::net::io::Listener,
     pending_sends: crate::net::PacketQueue,
     filter_chain: CachedFilterChain,
+    recv_ring_len: u16,
 ) -> eyre::Result<()> {
     let crate::net::io::Listener {
         worker_id,
@@ -70,7 +71,7 @@ pub fn spawn_listener(
 
     let socket = crate::net::DualStackLocalSocket::new(port).context("failed to bind socket")?;
 
-    let io_loop = IoUringLoop::new(2048, socket)?;
+    let io_loop = IoUringLoop::new(recv_ring_len, socket)?;
     io_loop
         .spawn_io_loop(
             format!("packet-router-{worker_id}"),
@@ -324,17 +325,28 @@ impl<'uring> LoopCtx<'uring> {
         &mut self,
         cqe: io_uring::cqueue::Entry,
         br: &'rb super::ring::BufferRing,
-    ) -> Option<RecvPacket<'rb>> {
+    ) -> std::io::Result<Option<RecvPacket<'rb>>> {
         let ret = cqe.result();
+        let flags = cqe.flags();
 
         if ret < 0 {
             let error = std::io::Error::from_raw_os_error(-ret);
             metrics::errors_total(metrics::READ, &error.to_string(), &metrics::EMPTY).inc();
-            tracing::error!(%error, "error receiving packet");
-            return None;
-        }
 
-        let flags = cqe.flags();
+            // For now, only requeue recv if it's ENOBUFS, most of the other errors that could theoretically occur for
+            // recvmsg either can't happen in our io-uring context, or would indicate a terminal error
+            if -ret != libc::ENOBUFS {
+                return Err(error);
+            }
+
+            tracing::debug!(%error, "no buffers available in recv ring buffer");
+
+            if flags & flags::IORING_CQE_F_MORE == 0 {
+                self.enqueue_recvmsg();
+            }
+
+            return Ok(None);
+        }
 
         // Requeue the recv if needed
         if flags & flags::IORING_CQE_F_MORE == 0 {
@@ -344,8 +356,9 @@ impl<'uring> LoopCtx<'uring> {
         // This _should_ theoretically never happen
         if flags & flags::IORING_CQE_F_BUFFER == 0 {
             metrics::errors_total(metrics::READ, "no buffer selected", &metrics::EMPTY).inc();
-            tracing::error!("failed to receive packet, a buffer was not selected");
-            return None;
+
+            // Differentiate this from ENOBUFS
+            return Err(std::io::Error::other("no buffer selected"));
         }
 
         let buffer_id = (flags >> 16) as u16;
@@ -356,14 +369,16 @@ impl<'uring> LoopCtx<'uring> {
             Err(error) => {
                 metrics::errors_total(metrics::READ, &error.to_string(), &metrics::EMPTY).inc();
                 tracing::error!(%error, "failed to extract source address");
-                return None;
+                // TODO: should this be terminal? not having this would indicate that the kernel is not upholding its
+                // API promise of prepending the source address before the packet contents
+                return Ok(None);
             }
         };
 
-        Some(RecvPacket {
+        Ok(Some(RecvPacket {
             buffer: rb,
             source: addr,
-        })
+        }))
     }
 
     /// Enqueues a `send_to` on the socket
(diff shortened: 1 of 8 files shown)


src/net/io/completion/io_uring.rs
```diff
@@ -482,18 +497,18 @@ impl IoUringLoop {
...
-                .build(self.concurrent_sends >> 1)?;
+                .setup_cqsize(recv_buffer_len)
+                .build(recv_buffer_len >> 1)?;
 
         let mut pending_sends_event = pending_sends.1;
         let pending_sends = pending_sends.0;
 
         let rb = super::ring::BufferRing::new(
-            concurrent_sends as u16,
+            self.recv_buffer_len as u16,
```
XAMPPRocky:
```suggestion
            recv_buffer_len as u16,
```

src/net/io/completion/io_uring.rs
```diff
@@ -589,7 +604,17 @@ impl IoUringLoop {
                             let op = (ud >> 56) as u8;
                             match op {
                                 IORING_OP_RECVMSG => {
-                                    let Some(packet) = loop_ctx.pop_recv(cqe, &rb) else { continue; };
+                                    let packet = match loop_ctx.pop_recv(cqe, &rb) {
+                                        Ok(Some(packet)) => packet,
+                                        Ok(None) => {
+                                            // A (hopefully) transient error occurred
```
XAMPPRocky, line 610: Should we not log it even if it's transient? Otherwise how would this case be visible to an operator?

XAMPPRocky (approved):

src/net/io/completion/io_uring.rs
Jake-Shadle: There is a message, it's just in a different location since it's specifically for this one error.

https://github.com/EmbarkStudios/quilkin/pull/1446/changes#diff-d7e8606084ba52d3c99e03da8e73ba5b87dd888bcc4ec7c00287f05e03c3f7efR336-R343

PR #1480, opened 2026-09-23
fix(mmdb): accept any width of integer for ASN

XAMPPRocky:


src/net/maxmind_db.rs
```diff
@@ -277,4 +304,116 @@ mod test {
         check((u32::MAX - 1) as _, &(u32::MAX - 1).to_string());
         check(u32::MAX as _, &u32::MAX.to_string());
     }
+
+    /// Just enough of the mmdb encoding to build a one record database.
```
Jake-Shadle:
```suggestion
    /// Just enough of the mmdb encoding to build one record database.
```

Jake-Shadle (approved):

glindstedt (approved):

